PR #8, opened 2025-02-11
Only trigger publishing on docs changes

janbrasna:
## One-line summary

Skips running when nothing new to build & deploy.

Includes docs with config, requirements, and the workflow itself (not sure if that one's somehow included implicitly, but added it for extra visibility anyway.)

Also removes unused RTD build & deploy config for sphinx no longer in use.


stevejalim (approved):
r+ Thanks, Jan!

PR #9, opened 2025-02-12
Misc fixes

alexgibson:
## One-line summary

- Reinstates `page_title_prefix` and `page_title_prefix` (which turns out are used on a small number of pages still)
- Fixes hard coded `/features/` page link in Firefox sub navigation.

## Issue / Bugzilla link

N/A

## Testing

Check page titles match what's in prod for mozorg:

- http://localhost:8000/en-US/download/
- http://localhost:8000/en-US/download/?xv=basic
- http://localhost:8000/en-US/ios/testflight/


stevejalim (approved):
r+

PR #11, opened 2025-02-12
Update /set-as-default/ to /default/

alexgibson:
## One-line summary

Migrates `/set-as-default/` to `/default/`

Note I didn't move or rename the Fluent files, as we should probably try and limit doing that unless we think it's worth the extra effort?

## Issue / Bugzilla link

N/A

## Testing

http://localhost:8000/en-US/default/
http://localhost:8000/en-US/default/thanks/


stevejalim (approved):
R+

Agreed re not moving the ftl files - it's not the same scale or change as going from new/ to download/

PR #13, opened 2025-02-12
Update db filename

stevejalim:
This changeset switches the sqlite DB from bedrock.db to springfield.db. There has also been infra work to bootstrap the S3 buckets that hold these files to download. 

In a short time, we'll switch to GCS for the file storage, but I wanted to get this working as-is for now.

## Testing

* `rm data/springfield.db`
* `make preflight`
* Confirm there's a new DB file at `data/springfield.db`
* Run the site - see if releasenotes works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 __pycache__/
 !.vscode/extensions.json
-!root_files/bedrock_db_info.json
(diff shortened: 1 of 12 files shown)


alexgibson (approved):
r+

PR #30, opened 2025-02-13
Remove X logos from product images

alexgibson:
## One-line summary

Ported from https://github.com/mozilla/bedrock/pull/15991

## Testing

http://localhost:8000/en-US/
http://localhost:8000/en-US/download/


stevejalim (approved):
r+

PR #31, opened 2025-02-13
Migrate updated features/adblocker template

alexgibson:
## One-line summary

Copies over new ad-blocker page template and fluent strings

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/pull/15929
https://github.com/mozilla/bedrock/pull/15959
https://github.com/mozilla/bedrock/pull/15974

## Testing

http://localhost:8000/en-US/features/adblocker/


springfield/firefox/views.py
```diff
@@ -599,3 +599,24 @@ def get_template_names(self):
             template_name = "firefox/features/pdf-editor.html"
 
         return [template_name]
+
+
+class firefox_features_adblocker(L10nTemplateView):
```
stevejalim, line 604: style nit: For views that are classes, the convention is to use PascalCase rather than the snake_case we use for function names.

stevejalim (approved):
r+wc

PR #37, opened 2025-02-16
Clean up wmo root files

janbrasna:
## One-line summary

Removes wmo-specific content from root_files and configs.

1. The internal CA certs are no longer used and mostly expired, only [linked](https://wiki.mozilla.org/MozillaRootCertificate) from wikimo to wmo, not needed here.
2. The banner (only content of /images in root) is a wmo legacy to preserve links, not needed here.
3. The matrix well-knowns are not needed here (they verify EMS instance at chat.m.o with asset URLs etc. and provide some domain-based discovery for the Element/Matrix at mozorg…)
4. Removed nonlocale entries incl. for redirects long gone…
5. (Also updates the domain/URL in info.txt)


stevejalim (approved):
Thanks for the laser-eyes, Jan!

PR #54, opened 2025-03-19
Fix functional test failures

alexgibson:


stevejalim (approved):
All makes sense - thank you!

PR #57, opened 2025-03-20
Update ALLOWED_HOSTS to support pod IP

robhudson:
## One-line summary

`$POD_IP,` is prepended to the env var for `ALLOWED_HOSTS` allowing us serve readiness and health checks at the pod level.

### DIFF
--- a/springfield/settings/base.py
+++ b/springfield/settings/base.py
@@ -476,15 +476,23 @@ EXTRA_INDEX_URLS = {
 #   },
 ALT_CANONICAL_PATHS = {}
 
-ALLOWED_HOSTS = config(
-    "ALLOWED_HOSTS",
-    parser=ListOf(str, allow_empty=False),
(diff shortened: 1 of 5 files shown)


springfield/settings/base.py
```diff
@@ -476,15 +476,23 @@ def language_url_map_with_fallbacks():
...
+ALLOWED_HOSTS = list(
+    filter(
+        None,
+        config(
+            "ALLOWED_HOSTS",
+            parser=ListOf(
+                str,
+                allow_empty=True,
+            ),
+            default="",
```
stevejalim, line 493: Will the lack of a default mean we have to ship webservices-infra changes first?

stevejalim (approved):
R+wc

springfield/settings/base.py
robhudson: you've got values set up already in all 3 envs from what I can see.

PR #62, opened 2025-03-31
Upgrade to Django 5.1

robhudson:


springfield/base/middleware.py
```diff
@@ -263,7 +263,7 @@ def __init__(self, get_response):
         if not settings.ENABLE_METRICS_VIEW_TIMING_MIDDLEWARE:
             raise MiddlewareNotUsed
 
-        self.get_response = get_response
+        super().__init__(get_response)
```
robhudson, line 266: Minor change here. Probably should have been written this way to begin with. The mix-in is now doing some async checks that started failing. This let's the parent class set up the object variables so there's no issue now.

PR #67, opened 2025-04-01
Update Fluent L10N repo names to be Springfield/firefox.com specific

stevejalim:
This changeset makes the default config for the l10n repos point to springfield-specific ones, which I have now also created.


.github/workflows/routine_l10n_obsolete.yml
```diff
@@ -51,8 +51,8 @@ jobs:
             Compare the files in /l10n/en-US/ to the /en/ folder in the www-l10n repo and remove
             files from www-l10n that no longer exist in springfield.
 
-            If you have both repos installed and up to date you can run this command:
-            `diff -r www-l10n/en springfield/l10n/en | grep ".ftl"`
+            If you have both mozmeao/repos installed and up to date you can run this command:
```
stevejalim, line 54:
```suggestion
            If you have both mozmeao and mozilla-l10n repos installed and up to date you can run this command:
```

stevejalim (commented):
Self-r: r+

PR #68, opened 2025-04-02
Fix CDN and download tests

stevejalim:
This is a first swing at making the CDN and Download tests pass, which are currently failing in part because they are intended for prod and we don't have prod up yet, and certainly not yet on www.firefox.com.

In the meantime, I've disabled the schedule and we should at least be able to run the tests against Dev to check that the paths they are seeking are correct, etc


stevejalim:
self-r: r+

going to merge this for now so we can iterate on `main` - can't run the GHAs without a build container in the registry

PR #74, opened 2025-04-08
(chore) L10N: reinstate updated Pontoon config for newsletters.ftl

### LINKED ISSUE #73: Update pontoon config to ensure newsletters.ftl gets translated
## Description

Spotted by @flodolo - we need to ensure that the `newsletter/newsletters.ftl` file, which moved from `mozorg/` is configured for pontoon similar to https://github.com/mozilla/bedrock/blob/main/l10n/configs/pontoon.toml#L309-L326

This change will also be a nice test of the automation between the repos, so we should ensure the relevant GHAs are enabled

stevejalim:
...which got lost when pruning out mozorg-related Pontoon config

## Issue / Bugzilla link

Resolves #73 

## Testing

We should be able to see this change flow all the way from `mozmeao/springfield` to `mozilla-l10n/www-firefox-l10` and then back to `mozmeao/www-firefox-l10n`

### DIFF
--- a/.github/workflows/send_firefox_fluent_strings_to_l10n_org.yml
+++ b/.github/workflows/send_firefox_fluent_strings_to_l10n_org.yml
@@ -10,8 +10,8 @@ on:
     branches:
       - main
   workflow_dispatch:
-  #schedule:
-    #- cron: "0 0/3 * * *" # Every 3 hours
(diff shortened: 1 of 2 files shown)


alexgibson (approved):
r+

PR #83, opened 2025-04-15
Fix saucelabs errors after upgrades

robhudson:
The error:
WebDriverException: Message: Sauce-specific capabilities are no longer accepted as W3C WebDriver capabilities: avoidProxy, tags. Ensure to enclose these in sauce:options capabilities.


pmac (approved):

robhudson:
yep, just the firefox-remote one now. I think I saw something for that one too. I'll look again.

PR #87, opened 2025-04-16
Limit publish_docs to mozmeao/springfield

janbrasna:
Similar to [mozilla/bedrock#16182](https://github.com/mozilla/bedrock/pull/16182/files)

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -21,6 +21,7 @@ concurrency:
 
 jobs:
   build:
+    if: github.repository == 'mozmeao/springfield'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


stevejalim (approved):

janbrasna:
BTW:

> "This action currently fails with an (expected) error every time someone commits to their fork of bedrock. This added check makes it so we only try building and deploying the docs when commits are made here."

The failure is expected if you don't have GHA deployments (=environment) enabled for the fork (which is the default after forking ofc) but there's nothing else that would stop building and publishing the site under the fork's ghio domain — I personally did allow the publishing back in the day to fiddle with the first mkdocs attempts and effectively used it as a build preview… so I see some utility to be able to workflow-dispatch e.g. the PR branch and have own preview deployment to link up to the PR etc. …

Unfortunately GHA doesn't have a simple "is this enabled" check; and the action has lost the ability to force-create the environment last year to actually be able to deploy (as it has some of the pages:write token permissions already, just not the protection rule admin to tweak deployments now:/…) but I'll be looking into checking some GH API queries to try to infer a pages configuration or environment enablement or whatever's used in their OIDC publishing config, and wire it up to some if:always() or continue-on-error way of not failing over a publishing not set up correctly on forks — or more precisely, find a way to fail it very loudly for the origin repos, but silently skip nonexistent environments on forks… Will report back once I come up with something reasonable.

PR #90, opened 2025-04-21
89 csp connect src expansion

### LINKED ISSUE #89: Support env-configurable CSP_CONNECT_SRC values
At the moment the only way to expand CSP_CONNECT_SRC is by adding the host to CSP_DEFAULT_SRC, which is more lax that we want to be

stevejalim:
* Adds support for expanding `CSP_DEFAULT_SRC` via an env var
* Switches to walrus operator to make things slightly less verbose

## Issue / Bugzilla link

Resolves #89 

## Testing

```
$ ./manage.py shell
>>> from django.conf import settings
>>> settings.CONTENT_SECURITY_POLICY  # should contain only defaults
>>> exit
$ export CSP_DEFAULT_SRC=test.example.com
>>> from django.conf import settings
>>> settings.CONTENT_SECURITY_POLICY  # should contain `test.example.com` as a `connect-src` value in addition to all the original ones


robhudson (approved):
`:=`

PR #98, opened 2025-04-29
Refactor CSP settings

robhudson:
## One-line summary

This refactors the CSP settings a bit to consolidate patterns, layer on better security practices, and bring some of the report-only policies into the enforced policy.

I tried to break this down into logical commits as much as possible to help show the rationale, but happy to connect for a walk-through if needed. 

## Testing

I did some local driving and things looked good. Anything I found locally I fixed.

E.g. the newsletter sign-up at `/en-US/developer/` was blocked due to `connect-src` not having a basket url (whereas on bedrock the `*.mozilla.org` wildcard catches that), and that is now fixed here.

### DIFF
--- a/springfield/settings/__init__.py
+++ b/springfield/settings/__init__.py
@@ -85,6 +85,10 @@ _csp_form_action = {
     BASKET_URL,
     FXA_ENDPOINT,
 }
+# On hosts with wagtail admin enabled, we need to allow the admin to frame itself for previews.
+_csp_frame_ancestors = {
+    csp.constants.SELF if WAGTAIL_ENABLE_ADMIN else csp.constants.NONE,
+}
 
 # 2. TEST-SPECIFIC SETTINGS
 # TODO: make this selectable by an env var, like the other modes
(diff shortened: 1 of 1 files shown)


springfield/settings/__init__.py
```diff
@@ -22,17 +22,64 @@
...
+    "o1069899.ingest.sentry.io",
+    FXA_ENDPOINT,  # noqa: F405
+}
+_csp_font_src = {
+    csp.constants.SELF,
+}
+_csp_form_action = {
+    csp.constants.SELF,
+    # NOTE: Check if these need to be in the `_csp_connect_src` list as well since we often
+    # progressively enhance forms by using Javascript.
```
stevejalim, line 55: Can we check this before we merge?

springfield/settings/__init__.py
```diff
@@ -22,17 +22,64 @@
...
+# - if no path is provided, all resources are allowed from the origin.
+# - if path is provided with no trailing slash: an exact-match is required.
+#   - e.g. `https://example.com/api` will only match `https://example.com/api`
+# - if path is provided with trailing slash: the path is a prefix-match.
+#   - e.g. `https://example.com/api/` will match anything that starts with `https://example.com/api/`
+
 _csp_default_src = {
+    # Keep `default-src` minimal. Best to set resources in the specific directives.
     csp.constants.SELF,
-    "*.firefox.com",
```
stevejalim, line 27: Just to double-check: we're able to drop this because it's effectively going to be the same as `SELF` yeah?

stevejalim (approved):
r+wc

springfield/settings/__init__.py
robhudson: That comment is more for future devs. I've checked it for what's there now.

springfield/settings/__init__.py
robhudson: yes, but this also removes the wildcard which is more safe. If we did need something like `api.firefox.com` it's better to add that explicitly. Also, with these changes most of the main fetch directives have values so won't use `default-src` as a fallback. Safest is to set it to `'none'` and adjust the specific directives but I didn't want to jump all the way there in this round of clean-up.

PR #100, opened 2025-04-30
Review warnings during test run

robhudson:


springfield/firefox/migrations/0003_initial.py
```diff
@@ -20,7 +20,6 @@ class Migration(migrations.Migration):
     dependencies = [
         ("cms", "0004_rename_bedrockrendition_springfieldrendition"),
         ("wagtailcore", "0093_uploadedfile"),
-        ("firefox", "0002_delete_firefoxosfeedlink"),
```
robhudson, line 23: I think we can do this safely. We may have a couple records in the db but the migrations tools work fine still.

stevejalim: I wonder why this is called `*_initial.py` - did Django assume that because there was no frozen state (due to 0001 adding a model and 0002 dropping it)?

robhudson: Not sure, but I think this is pretty safe if you'd like to deploy it. Mostly it's for the tests to avoid creating and dropping a db table which has deprecated code that outputs a warning.

stevejalim (approved):

PR #104, opened 2025-05-01
(fix) Springfield: try to deal with missing C support for psycopg

stevejalim:
For local dev, you may need to install libpq to get deps to install, if you don't already have postgres installed


stevejalim:
Currently fails to install prod deps locally with

```
21.65       couldn't run 'pg_config' --includedir: [Errno 2] No such file or directory: 'pg_config'
21.65       error: [Errno 2] No such file or directory: 'pg_config'
21.65       [end of output]
```

and the same in the Docker container

robhudson (approved):

PR #107, opened 2025-05-02
(fix) Allow customisation of img-src, script-src and media-src CSP directives

stevejalim:
Because we're no longer combining the the default-src with, say, script-src, we're getting some blocking of things that didn't get blocked before.  This changeset allows us to expand img, script and media directives. We may also need style-src


robhudson:
Closing in favor of #108

PR #108, opened 2025-05-02
Add `CSP_ORIGIN_HOST_FALLBACK` to appropriate CSP settings

robhudson:
## One-line summary

Springfield is accessible under different URLs (origin and CDN) but `STATIC_URL` points to specific URL. We encountered CSP issues when these don't match. This patch allows the origin host to be configured and added to the CSP where appropriate to allow the assets to load.


pmac:
Looks like for bedrock we just have `*.allizom.org` in the CSP. Could we just do `*.moz.works` here?

robhudson:
For security I'd say it's better to avoid wildcards as much as possible. But this should fix the currently broken CSP issues for the integration tests which are hitting the origin.

PR #122, opened 2025-05-13
Fx–themed pages briefly show blue links in nav on hover

alexgibson:
## One-line summary

See: https://github.com/mozilla/bedrock/pull/16147

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/15873

## Testing

http://localhost:8000/en-US/

### DIFF
--- a/media/css/m24/components/navigation-refresh.scss
+++ b/media/css/m24/components/navigation-refresh.scss
@@ -364,6 +364,11 @@ $margin-top: 54px; // top margin offset for mobile navigation menu
         text-decoration: none;
     }
 
+    // band aid fix for https://github.com/mozilla/bedrock/issues/15873
+    &:hover:is(a) {
+        color: $color-black;
+    }
+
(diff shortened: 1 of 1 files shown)


media/css/m24/components/navigation-refresh.scss
```diff
@@ -364,6 +364,11 @@ $margin-top: 54px; // top margin offset for mobile navigation menu
         text-decoration: none;
     }
 
+    // band aid fix for https://github.com/mozilla/bedrock/issues/15873
```
janbrasna, line 367: BTW I just checked if the type stack +vars in MZPv20 and simplified base fixed anything about this quirk but unfortunately it'd still exhibit, so for now this has to stay:/

(There seems to be more unwanted interaction between firefox-base specifics and global components having to resist them, not unlike https://github.com/mozilla/bedrock/issues/15803#issuecomment-2845902837 …)

PR #130, opened 2025-05-15
Bootup optimisations

### LINKED ISSUE #49: [Infra] stop the need for the pods to download the sqlite DB in order to start up
In k8s we have init pods that ensure a fresh sqlite DB is downloaded before the deployment can roll out. We don't need this any more, but we DO still need fresh l10n files, so we should amend the init pods.

This applies to both Bedrock and to Springfield. Let's solve it on Bedrock and then we can replicate for Springfield

stevejalim:
* Remove need for the DB to be downloaded before startup if DATABASE_URL shows it's postgres
* Optimise speed of l10n updating by adding a new scheduled task that only does l10n files not also a DB; this avoids breaking demos and springfield-test

## Issue / Bugzilla link

Resolves #49 

## Testing

We're gonna need to try this on springfield-dev. I've got an infra PR here which disables the DB-download initContainer sidecar, and also uses the l10n-only scheduled job: https://github.com/mozilla-it/webservices-infra/pull/5518

We should also push main to run-integration-tests if Dev comes up fine after everything is merged

### DIFF
--- a/bin/cron.py
+++ b/bin/cron.py
@@ -172,6 +172,14 @@ def schedule_file_jobs():
                 raise
 
 
+def schedule_l10n_jobs():
+    call_command("l10n_update --clean")
+
+    @scheduled_job("interval", minutes=DB_UPDATE_MINUTES)
+    def update_locales():
+        call_command("l10n_update")
+
+
 def main(args):
     has_jobs = False
     if "db" in args:
@@ -179,6 +187,9 @@ def main(args):
         has_jobs = True
     if "file" in args:
         schedule_file_jobs()
+        has_jobs = True
+    if "l10n" in args:
+        schedule_l10n_jobs()
         has_jobs = True
 
     if has_jobs:
--- a/bin/run-prod.sh
+++ b/bin/run-prod.sh
@@ -7,13 +7,15 @@
 # look for the required files and fail quickly if it's not there
 STARTUP_FILES=(
     "data/last-run-update_locales"
-    "data/last-run-download_database"
 )
 # If DATABASE_URL is defined, that means we're using Postgres not sqlite.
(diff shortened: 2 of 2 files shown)


bin/run-prod.sh
```diff
@@ -7,13 +7,15 @@
...
-    "data/last-run-download_database"
 )
 # If DATABASE_URL is defined, that means we're using Postgres not sqlite.
 # However, if DATABASE_URL is NOT defined, we need to be sure the sqlite DB file
-# is already present at startup
+# is already present at startup and that we've downloaded the latest one
 if [[ -z "$DATABASE_URL" ]]; then
-    STARTUP_FILES+=("data/springfield.db")
+    STARTUP_FILES+=(
+        "data/springfield.db"
```
pmac, line 16: do we still need to check for this?

pmac: If `DATABASE_URL` is not defined at startup we should probably just not startup and let k8s keep the old pods around.

stevejalim: > do we still need to check for [data/springfield.db]?

Good point - as long as we're checking for the db-download sentinel we don't need to check for the DB file

pmac: I guess I just don't see the point of checking these anymore. I think if the instance needs the DB file and it isn't there it just won't work. We can keep this I guess but it just feels like something that could unexpectedly interfere with startup in some situations.

bin/cron.py
```diff
@@ -180,6 +188,9 @@ def main(args):
     if "file" in args:
         schedule_file_jobs()
         has_jobs = True
+    if "l10n" in args:
```
pmac, line 191: we should file an issue to remove the rest of the jobs from this file. We could do it now, but no need to increase the size of this diff.

pmac (approved):

Issue #126, opened 2025-05-13
Bring additional pages over from bedrock


slightlyoffbeat:
## Description

After the decision to keep the MVP launch closer to 1:1 to bedrock, there are additional pages to bring over:

/firefox/browsers/mobile/
/firefox/windows/
/firefox/mac/
/firefox/browsers/mobile/ios/
/firefox/browsers/mobile/android/
/firefox/browsers/mobile/focus/
/firefox/linux/

alexgibson:
What should the URL slugs for these additional pages transition to? Should we simply omit `/firefox/`, or should some be prefixed with `/download/` for example?

slightlyoffbeat:
Browser comparison pages need to be brought over

to do: @slightlyoffbeat will update the issues

slightlyoffbeat:
Instead of updating this issue, I created a new one for the rest of the pages:

https://github.com/mozmeao/springfield/issues/128

@stevejalim a doc was created [here](https://docs.google.com/spreadsheets/d/1kIc8umb2D-3cVSytYM1DrHEphiZpes7jDKKPGTnP8P8/edit?usp=sharing):

PR #131, opened 2025-05-15
Migrate Firefox platform download pages (Issue #126)

alexgibson:
## One-line summary

Migrates Firefox platform download pages from bedrock.

I had to move a few image / JS assets into /browsers/desktop/platform/, which were being loaded from random places elsewhere. A spot check to make sure I'm not missing anything else would be good.

I also tried to update as many hard coded URLs as i could find to use `url()` now the pages are internet to springfield. A spot check of these would be useful too.

## Issue / Bugzilla link

#126

## Testing

- http://localhost:8000/en-US/browsers/desktop/windows/
- http://localhost:8000/en-US/browsers/desktop/mac/
- http://localhost:8000/en-US/browsers/desktop/linux/

Successful test run: https://github.com/mozmeao/springfield/actions/runs/15047524991


maureenlholland (approved):
r+wc 👍 

one rogue www.mozilla.org utm value

alexgibson:
Thanks @maureenlholland updated

Issue #128, opened 2025-05-15
Batch of pages to bring over from Bedrock


slightlyoffbeat:
## Description

After a full review of pages to bring over from bedrock, and in an effort to keep a close 1:1 for launch, additional pages have been identified to bring over from bedrock.

Below, I'll include the pages to bring over, and the destination URL. In every case, it should be assumed that /firefox/ should be omitted. If I indicate "no change", that means that the URL is the same as it would have been on bedrock, sans /firefox/. Otherwise, a new URL will be requested.

This list does not include items from https://github.com/mozmeao/springfield/issues/126

---

- [ ] `/browsers/best-browser/` -> `/more/best-browser/`
- [ ] `/browsers/browser-history/` -> `/more/browser-history/`
- [ ] `/browsers/chromebook/` -> `/browsers/desktop/chromebook/`
- [ ] `/browsers/compare/` -> `/compare/`
- [ ] `/browsers/compare/brave/` -> `/compare/brave/`
- [ ] `/browsers/compare/chrome/` -> `/compare/chrome/`
- [ ] `/browsers/compare/edge/` -> `/compare/edge/`
- [ ] `/browsers/compare/ie/` -> `/compare/ie/`
- [ ] `/browsers/compare/opera/` -> `/compare/opera/`
- [ ] `/browsers/compare/safari/` -> `/compare/safari/`
- [ ] `/browsers/incognito-browser/` -> `/more/incognito-browser/`
- [ ] `/browsers/mobile/compare/` -> `/browsers/mobile/`
- [ ] `/browsers/quantum/` -> `/browsers/desktop/quantum/`
- [ ] `/browsers/update-your-browser/` -> `/more/update-your-browser/`
- [ ] `/browsers/what-is-a-browser/` -> `/more/what-is-a-browser/`
- [ ] `/browsers/windows-64-bit/` -> `/more/windows-64-bit/`
- [ ] `/facebookcontainer/` -> `/more/facebookcontainer/`
- [ ] `/family/` -> `/landing/family/`
- [ ] `/faq/` -> `/more/faq/`
- [ ] `/mobile/get-app/` -> `/browsers/mobile/get-app/`
- [ ] `/more/` -> `/more/` (No change)
- [ ] `/more/misinformation/` -> `/more/missinformation/`
- [ ] `/set-as-default/` -> `/landing/set-as-default/`
- [ ] `/set-as-default/thanks/` -> `/landing/set-as-default/thanks/`
[...]

PR #139, opened 2025-05-19
Migrate /firefox/mobile/get-app to springfield

maureenlholland:
## One-line summary
Move `firefox/mobile/get-app` to springfield, rename url to `browsers/mobile/get-app`

I changed the ftl file name, from `mobile` to `get-app`, since it seems this is the only place it is used (checked in bedrock). Does this create any issues for l10n? Should I leave FTL name as it is in bedrock?

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/issues/128

## Testing

http://localhost:8000/en-US/browsers/mobile/get-app/

url helper links work
- [x] from http://localhost:8000/en-US/download/all/mobile-release/
Screenshot 2025-05-19 at 6 01 52 PM

- [x] from http://localhost:8000/en-US/
Screenshot 2025-05-19 at 5 59 58 PM

in comparison with https://www.mozilla.org/en-US/firefox/browsers/mobile/get-app/


maureenlholland:
@stevejalim I'm not sure about the Fluent rename here. It's more tied to the fact I don't see it used anywhere else than trying to match new path. I haven't renamed any of the strings in the file (and don't plan to). Waiting for review to confirm this is a useful rename (and then can add your name to commit as well)

stevejalim:
> I changed the ftl file name, from mobile to get-app, since it seems this is the only place it is used (checked in bedrock). Does this create any issues for l10n? Should I leave FTL name as it is in bedrock?

Changing the name is fine as long as I make sure I rename the translated versions, too - I have a script for that, so it should be fine

stevejalim:
One thing we'll need to do after all this work is merged is circle back and check that all the pages have the language switcher enabled (which is a sign that the fluent files are in place)

stevejalim (approved):
R+ from me - works locally and the fluent rename is on my list to handle. Thank you!

Issue #116, opened 2025-05-07
update homepage of firefox.com


slightlyoffbeat:
## Description

The current homepage of firefox.com is the old /firefox/ page.

- the current index page (formerly `/firefox` on bedrock) page will be [moved and repurposed](https://github.com/mozmeao/springfield/issues/132)
- update index `/` to be the `/new/` page (previously `/firefox/new` on bedrock, currently `/download/` on springfield)
- if we need to redirect /download/, we can redirect to the homepage. 

see: 
https://github.com/mozilla/bedrock/issues/16160
https://github.com/mozilla/bedrock/issues/16257

stephaniehobson:
We're going to use this page as the[ /desktop page ](https://github.com/mozmeao/springfield/issues/132) so it could be be moved and slightly edited as part of this issue if it makes sense to bundle the work.

PR #142, opened 2025-05-20
(feat) Port unsupported-browsers page from Bedrock

stevejalim:
From www.mozilla.org/firefox/unsupported-systems/ to /browsers/unsupported-systems

Note that this page was not localised in Bedrock, so isn't localised here.

Instead of pointing to /new/ (which doesn't exist in Springfield), I'm pointing to the root page, ahead of https://github.com/mozmeao/springfield/issues/116 landing

## Issue / Bugzilla link

Part of #128

## Testing

Compare http://localhost:8000/en-US/browsers/unsupported-systems/ with https://www.mozilla.org/en-US/firefox/unsupported-systems/


maureenlholland (changes requested):
URL conflict needs resolving, otherwise this is good to merge

maureenlholland (approved):

PR #143, opened 2025-05-20
Set GH deployment env based on CI env outputs

janbrasna:
## One-line summary

Attempt to reinstate individual deployment environments on GitHub and their reporting from Actions (+to properly categorize pushes based on dev-, stage-, test-, and prod-triggered events).

Before trying on bedrock https://github.com/mozilla/bedrock/issues/16266 figured I should attempt being brave here;) It's based on prior art from the CD pipeline author elsewhere [mozilla/kitsune#L128](https://github.com/mozilla/kitsune/blob/1.2.10/.github/workflows/build-and-push.yml#L128) and makes sense comparing the two repos' deployment reporting.

The value comes from the `*_image_tag` steps above:
https://github.com/mozmeao/springfield/blob/0175650c453f448ebd29cfa6f656917fc520f1b0/.github/workflows/build-and-push.yml#L78-L81

and is consumed in this job anyways on the cli:
https://github.com/mozmeao/springfield/blob/0175650c453f448ebd29cfa6f656917fc520f1b0/.github/workflows/build-and-push.yml#L211-L212

which already has the preceding job in `needs:` for good measure.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/16266

## Testing

With the envs basically not used for anything here on GH (all the CD is done elsewhere), being empty, having no rules etc., the idea is the CI should be able to create them and not trip over some branch protections and such. — Basically upon landing in main the "dev" env should be listed for the trigger instead of "build" once the CI finishes pushing the images.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -162,7 +162,7 @@ jobs:
   upload_static_assets:
     name: Upload static assets
     runs-on: ubuntu-latest
-    environment: build
+    environment: ${{ needs.push_image_to_gar.outputs.deployment_env }}
     needs: [build_and_publish_public_images, push_image_to_gar]
     permissions:
       contents: read


stevejalim (commented):
Let's give this a go and if it works as expected, we can clone it to Bedrock too

stevejalim:
@Basma1912 Does this look like it would break anything in Argo, please?

Basma1912:
@stevejalim it won't impact anything in Argocd .. after reviewing this .. the environment argument is not being used  in the upload assets

janbrasna:
@stephaniehobson This is still valid as–is. Just the timing was never great. In theory this is just a cosmetic label to better tell announcements, deployments and statuses apart. On the other hand formally deployment protections on the named environments then have the ability to stop folks _(non-admin I guess; some presumably have the privileges to circumvent protections)_ from publishing if not configured properly and this will push its own implicit environments, just for GitHub use — so I'd say it would be great to see this work from dev through test and stage to prod, but ideally at some point when there are no pressing deployment around though;)

janbrasna:
Yeah that means it has to be part of the claim even on the other GCP login — which makes it different here from bedrock.

While I'd love to have the environments labelled correctly on GH side +their notifications, it's more of IAM work that just might not be a priority.

janbrasna:
E.g. ideally there would be, say, `test` access claim to only write `nonprod-test-media` and `stage` claim for `nonprod-stage-media` etc. for tighter control.

Currently I see the `test` stage logs in as artifact-writer@moz-fx-springfield-**<mark>prod</mark>**.iam.gserviceaccount.com — this is correct/expected?

PR #144, opened 2025-05-20
(fix) Stop trying to push to the non-existent 'master' branch

stevejalim:
Should fix https://github.com/mozmeao/springfield-fluent-update/actions/runs/15136798855/job/42550930911


stevejalim (commented):
self-review: r+

PR #157, opened 2025-05-22
Update appstore test to gb URLs

janbrasna:
## One-line summary

Following up from #154, the locale would change the appstore links, too.

To point to a different version of the page, a different locale is used. This renders some URLs with different values.


stephaniehobson (approved):
R+ 🇬🇧

PR #159, opened 2025-05-22
(feat) Move set-as-default pages to below a landing/ path

stevejalim:
URL path change for the already-ported set-as-default pages

L10N files are also already set up.

## Issue / Bugzilla link

#128

## Testing

* http://localhost:8000/en-US/landing/set-as-default/ links to http://localhost:8000/en-US/landing/set-as-default/thanks/
* http://localhost:8000/it/landing/set-as-default/ links to http://localhost:8000/it/landing/set-as-default/thanks/


stevejalim:
@slightlyoffbeat Let me know if this was more than just a URL change, but it seems as simple as that

slightlyoffbeat:
@stevejalim That is correct.

stevejalim:
Self review: r+

PR #163, opened 2025-05-22
update tests for new homepage variation and new site navigation

wen-2018:
## One-line summary

This PR fixes the failed tests here: https://github.com/mozmeao/springfield/actions/runs/15190373793/job/42721363082

1. Test for homepage variation based on locale
2. Tests for new site navigation

### DIFF
--- a/tests/functional/firefox/test_home.py
+++ b/tests/functional/firefox/test_home.py
@@ -6,9 +6,12 @@ import pytest
 
 from pages.firefox.home import FirefoxHomePage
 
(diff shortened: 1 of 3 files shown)


stephaniehobson (approved):
This fixes several of the tests 🔥 There's still one whining about the app store on the homepage in en-GB because apparently the URL is different when you're not in en-US. That can be fixed in a follow up PR. In fact, it will probably be easier to fix it when we've finished moving pages around.

R+

PR #175, opened 2025-05-23
Remove download view (since it's now homepage)

stephaniehobson:
Leaving this as a draft for now because there's enough stuff broken already.

## One-line summary

Remove download view (since it's now homepage)


janbrasna:
With other PRs removing any use of /download and relocating /download/thanks… would it eventually make sense to move the actual /download/all just to /download then?

PR #177, opened 2025-05-23
Update selenium tests for desktop landing page

stephaniehobson:
These tests were intended to test the page that is now at browsers/desktop I've updated the locations and names accordingly.

Also, since the en-US page is now the same as all other locales, we can test with the default locale.


stephaniehobson:
R+ That's a nice green pipe.

PR #164, opened 2025-05-22
Migrate /more section from bedrock

stephaniehobson:
## One-line summary

Migrate /more section from bedrock

## Significant changes and points to review

- Migrate /more landing page and section and associated CSS and images
- consolidate css files into a shared `more-article` bundle
- make bottom CTA into a shared include
- removed icon from sub nav (because it looks funny to see the two side by side)
- also removed sub nav from a number of pages
- snuck a navigation alignment tweak in

### L10n notes

- I removed some strings that are no longer in use
- I changed brand-name-* references for other companies and products to just use the word
- The more/shared.ftl file is new
- The following ftl files came from bedrock:
  - l10n/en/firefox/more.ftl  → l10n/en/firefox/more/more.ftl
  - l10n/en/firefox/faq.ftl  → l10n/en/firefox/more/faq.ftl
  - l10n/en/firefox/browsers/best-browser.ftl →l10n/en/firefox/more/best-browser.ftl
  - l10n/en/firefox/browsers/windows-64-bit.ftl →l10n/en/firefox/more/windows-64-bit.ftl
  - l10n/en/firefox/browsers/history/browser-history.ftl → l10n/en/firefox/more/browser-history.ftl
  - l10n/en/firefox/browsers/history/what-is-a-browser.ftl → l10n/en/firefox/more/what-is-a-browser.ftl
  
## Issue / Bugzilla link

#128

## Testing

http://localhost:8000/en-US/more/
- the chromebook link is still to be migrated

http://localhost:8000/en-US/more/what-is-a-browser/
http://localhost:8000/en-US/more/browser-history/
http://localhost:8000/en-US/more/incognito-browser/
http://localhost:8000/en-US/more/update-your-browser/
http://localhost:8000/en-US/more/windows-64-bit/

PR #181, opened 2025-05-24
Restore original Bedrock strings for /more

flodolo:
From #164

> I changed brand-name-* references for other companies and products to just use the word

We are copying over existing localizations from www-l10n (Bedrock), this doesn't work.

We need to treat changes to strings like we would do for changes within Bedrock (if the string changes, we need a new ID and fallback). Given the time constraints, we should keep strings as is for now, and park these cleanups for later.


stephaniehobson (approved):
R+

PR #214, opened 2025-05-29
(fix) Amend Release Notes links in Nav and Footer to point to Springfield, not www.mozilla.org

### LINKED ISSUE #172: Release notes links
Menu and footer Release notes links point to WMO.

Homepage US/CA content: "See Release Notes" points to FXC.

Download/all has for every channel a link under the description, that also point locally to FXC.

Channel/desktop page has a link in description of every channel, this one currently links also locally to FXC.

Nit:

The banner In notes footer while correct on WMO, has a weird color =contrast on FXC:

> fxc

stevejalim:
The content in each site will be the same, so may as well keep people on www.firefox.com

## Issue / Bugzilla link

Resolves #172 - as reported by @janbrasna 

## Testing

Check the nav and footer on any page

### DIFF
--- a/springfield/base/templates/includes/footer/footer.html
+++ b/springfield/base/templates/includes/footer/footer.html
@@ -56,7 +56,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
           </h2>
           <ul class="mzp-c-footer-list" data-testid="footer-list-resources">
             <li><a href="{{ url('firefox.browsers.compare.index') }}" data-link-position="footer" data-link-text="Compare">{{ ftl('footer-compare') }}</a></li>
-            <li><a href="https://www.mozilla.org/firefox/releasenotes/" data-link-position="footer" data-link-text="Release Notes">{{ ftl('footer-release-notes') }}</a></li>
+            <li><a href="{{ url('firefox.notes') }}" data-link-position="footer" data-link-text="Release Notes">{{ ftl('footer-release-notes') }}</a></li>
             <li><a href="https://support.mozilla.org/?{{utm_params}}" data-link-position="footer" data-link-text="Support">{{ ftl('footer-support') }}</a></li>
             <li><a href="https://addons.mozilla.org/firefox/?{{utm_params}}" data-link-position="footer" data-link-text="Add ons">{{ ftl('footer-addons') }}</a></li>
             <li><a href="https://community.mozilla.org/?{{utm_params}}" data-link-position="footer" data-link-text="Community">{{ ftl('footer-community') }}</a></li>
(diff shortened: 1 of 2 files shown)


stephaniehobson (approved):
Usually internal links use the `url` helper. I think that helps with maintainability when we change link locations (not that I expect us to do that here but conventions and all. If we can use that here we should. Otherwise, R+

stevejalim:
> Usually internal links use the `url` helper. I think that helps with maintainability when we change link locations (not that I expect us to do that here but conventions and all). If we can use that here we should. Otherwise, R+

The dynamic release notes view that shows the latest one doesn't have a URL name to reverse so this was a quick fix

springfield/base/templates/includes/footer/footer.html
```diff
@@ -56,7 +56,7 @@ <h2 class="mzp-c-footer-heading" data-testid="footer-heading-resources">
           </h2>
           <ul class="mzp-c-footer-list" data-testid="footer-list-resources">
             <li><a href="{{ url('firefox.browsers.compare.index') }}" data-link-position="footer" data-link-text="Compare">{{ ftl('footer-compare') }}</a></li>
-            <li><a href="https://www.mozilla.org/firefox/releasenotes/" data-link-position="footer" data-link-text="Release Notes">{{ ftl('footer-release-notes') }}</a></li>
+            <li><a href="/firefox/releasenotes/" data-link-position="footer" data-link-text="Release Notes">{{ ftl('footer-release-notes') }}</a></li>
```
janbrasna, line 59: I think the `firefox_url()` helper was ported here too — could that be updated to the new URL and used universally?

NB: New home links there, but with `url()`: https://github.com/mozmeao/springfield/blob/d27793d05605e22ab59f80449bce1e975f3ab4f8/springfield/firefox/templates/firefox/download/desktop/download-en-us-ca.html#L104

> "I think that helps with maintainability when we change link locations (not that I expect us to do that here but conventions and all)."

IIUC the idea is to refactor the `/firefox/` out at some point, so it actually will change(?)

Generally, I wasn't sure what's the plan so I left just a general observation in the issue without any further details — understanding the in-product links will still be pointing to WMO (and nucleus attachments also publish there), it felt like that location will be used canonically. Makes sense to transitionally have the info in both locations; or start redirecting the WMO to FXC so no matter where in-product links point to, as long as any params are preserved and attributed if necessary(?) — just wanted to mention the risk of having duplicate content indexed, if served on both sites for some time.

stevejalim:
> Usually internal links use the `url` helper. I think that helps with maintainability when we change link locations (not that I expect us to do that here but conventions and all). If we can use that here we should. Otherwise, R+

Updated now to use a url helper and a named route `{{ url('firefox.notes') }}` which redirects to the same destination as `/firefox/releasenotes/` does

PR #217, opened 2025-05-29
Add release notes admin

pmac:
This moves release notes administration into Springfield and will thus mean we no longer need to maintain and run [Nucleus](https://nucleus.mozilla.org).

Replaces #161 since that one was based on a branch in my fork

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-104)
┆Priority: P3
┆Sprint: Backlog


pmac:
I think most of what is left to do is to fix up the tests. I think the ones that I have left in place are good tests, but we need a new way to load test data since the tests used the ability of the old model to load data from JSON files.

PR #229, opened 2025-06-02
222 Add refract replacement redirects

### LINKED ISSUE #222: Add redirects from refractr to springfield
At the moment, SRE's refractr service sends some requests for www.firefox.com/some/path to a different path on www.mozilla.org

When we start to point `https://firefox.com` and `https://www.firefox.com` we will basically lose the refractr behaviour (unless we commit to a much more complex solution, which we don't want to do)

So, we need to add some rediects to springfield for those redirects that really do still need to go from www.firefox.com/something to somwhere else that isn't handled by Springfield.

stevejalim:
## One-line summary

This changeset adds in-Django redirects that replace the ones we will pull out from refractr (when refractr no longer handles traffic for www.firefox.com)

See https://github.com/mozilla-it/refractr/pull/400

## Issue / Bugzilla link

Resolves #222

## Testing

`pytest -k test_global_conf_url` + manual testing of the paths added to `redirects.py`


robhudson (approved):
Good fixups. 👍

PR #240, opened 2025-06-03
Clean up references to /download/ path

### LINKED ISSUE #212: Clean up tests following change to redirect from /download/ to just /
## Description

While the tests are passing following the [change to redirect](https://github.com/mozmeao/springfield/pull/165/files#diff-f777a1022e563f9267af6d5e58212cb00d7bfa89f604aa00ca1be6d661564244R44) from `/download/` to the newer homepage at `/` (which also contains a Download Firefox button), we should clean up tests that try to hit `/download/` so that the redirect is not creating a false positive or having some other impact on the test

stevejalim:
## One-line summary

This changeset removes references to the `/download/` URL path and view, which is now replaced with `/`. 

Up till now, tests were passing because of the redirect that was added a few weeks ago. Now, tests don't rely on that redirect.

There's also been a small amount of other cleaning up around `/download/thanks/`

## Issue / Bugzilla link

Resolves #212 

## Testing

Checking links work and pages render with `DEBUG=False` in your `.env`

- [x] http://localhost:8000/en-US/features/adblocker/
- [x] http://localhost:8000/en-US/features/block-fingerprinting/
- [x] http://localhost:8000/en-US/features/translate/
- [x] http://localhost:8000/en-US/features/pdf-editor/
- [x] http://localhost:8000/fr/features/free-pdf-editor/
- [x] http://localhost:8000/fr/features/pdf-editor/
- [x] http://localhost:8000/fr/features/pdf-editor/kjdhfglkdhfgljfkdhgljkd -- to get a 404

### DIFF
--- a/profiling/hit_popular_pages.py
+++ b/profiling/hit_popular_pages.py
@@ -27,7 +27,6 @@ paths = [
     "/en-US/channel/desktop/",
     "/en-US/channel/desktop/?reason=manual-update",
     "/en-US/channel/desktop/developer/",
-    "/en-US/download/",
     "/en-US/thanks/",
     "/en-US/thanks/?s=direct",
     "/en-US/browsers/enterprise/",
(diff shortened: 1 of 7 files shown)


robhudson (approved):
Local testing looks good. Code looks good.

Only question I have is that the url `/download/all/` might suggest there's something at `/download/` since it's a sub-folder of it.

PR #241, opened 2025-06-03
Update bug_report template label

janbrasna:
## One-line summary

Seems the template minds the case difference in the issue label referenced.

Or, the other way around, fixing this by changing the actual issue label itself to Title-case to match bedrock for consistency, in which case this PR is moot.

## Issue / Bugzilla link

Many! ;)

## Testing


robhudson:
The casing consistency of our labels are the wild west. I kinda like lowercase and this fixes the issue. 👍

Issue #189, opened 2025-05-26
Fix header nav "Resources" button


stephaniehobson:
Right now the "Resources" heading in the nav has an empty `href` value. We can't just remove it because the code is set up to require a link to open the drop down menu. 

Some fix options are:
- stop using the Protocol nav code and code something new
- create a page at /resources that includes the links in the drop down.

janbrasna:
I was considering e.g. /more or similar kitchen sink as its href. 🤷

stephaniehobson:
We explicitly don't want to give /more the authority that a link in the navigation would give it.

I'm thinking we update #todo to  #m24-c-menu-panel-resources and ship it. What do you think @janbrasna?

janbrasna:
I've also suggested /releases index that's not linked very visibly; but it doesn't really matter what's the content. A fragment would do just well.

janbrasna:
Or also https://www-dev.springfield.moz.works/en-US/features/tips/ seems orphaned now, so that’s another candidate.

PR #247, opened 2025-06-03
Fix JS TypeError in nav menu and remove caret on mobile (Fix #186)

### LINKED ISSUE #186: Menu event throws JS TypeError breaking nav menu for touch devices
## Description

Clicking the "Resources" menu item itself raises an error: https://github.com/mozmeao/springfield/pull/146#discussion_r2103125949

For pointer devices this is only a little nuisance as any hover still allows for accessing the items in the subnav panel, but touch devices rely on the click handlers opening the subnav for them.

### Steps to reproduce

Desktop: https://github.com/mozmeao/springfield/pull/146#discussion_r2103125949
Mobile:

https://github.com/user-attachments/assets/bd9c5a3f-4d00-46f4-a5c8-dbaf6134178d

(It is demonstrated no matter how the menu is dismissed, either by tapping away, scrolling, or using the X affordance, any subsequent trigger is fruitless. The recording should display the tap attempts afterwards.)

### Expected result

No error regardless of input mode, subnav accessible even scripted on clicks/taps, not only on hover.

### Actual result

On desktop, clicking the nav logs the JS error. (Subnav still displays on hover.)

On mobile, clicking the nav opens the subnav correctly the first time, but apparently breaks right afterwards _(haven't wired up the device for tethered introspection to see if the culprit is the same or the two are unrelated, but my guess would be it's the same issue.)_ making the subnav items inaccessible once the panel closes unless the page is reloaded/navigated away.

### Environment

FXDE, WebKit TP on desktop (pointer).
FxiOS beta, Safari stable on mobile (touch).

### LINKED ISSUE #196: Mobile Menu "Resources" does not collapse when tapped (P2)
## Description

Mobile Menu "Resources" does not collapse when tapped

### Steps to reproduce

Image

### Expected result

The menu drawer should either not contain a dropdown caret (preferrable) - OR should collapse when tapped

### Actual result

There was no interactivity triggered

### Environment

Mozilla/5.0 (Linux; Android 11; SAMSUNG SM-G973U) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/14.2 Chrome/87.0.4280.141 Mobile Safari/537.36

### LINKED ISSUE #233: Resources dropdown arrow appears stacked on mobile
## Description

Resources dropdown arrow appears stacked on mobile when it should be side by side with text

### Steps to reproduce

https://www-dev.springfield.moz.works/en-US/ on mobile device. Open mobile menu.

### Expected result

Dev tools

Image

### Actual result

iPhone
Image

### Environment

iPhone 12 iOS 18.5 across Firefox/Chrome/Safari

stephaniehobson:
## One-line summary

Fix JS TypeError in nav menu and remove caret on mobile

- The added `h2` was causing the error because we were looking at the `target` instead of the `currentTarget`. 
- I made some additional changes because we don't actually expand/collapse the menu drop downs on mobile
- I also removed the caret next to Resouces which wrongly indicates there is an action to take on mobile.
- Note that: #189 is not addressed by this PR.

## Issue / Bugzilla link

Fix #186
Fix #196 
Fix #233 

## Testing

Manipulate the menu multiple times on both desktop and mobile. See if you can get it to throw an error or stop functioning again. You can also keep an eye on the HTML attributes that change to make sure aria-roles and classes are being updated as appropriate.

### DIFF
--- a/media/js/base/protocol/menu.es6.js
+++ b/media/js/base/protocol/menu.es6.js
@@ -296,7 +296,6 @@ MzpMenu.setAria = () => {
     }
 };
 
-
 /**
  * Unset initial ARIA menu panel states.
  */


janbrasna:
(Needs resolving conflicts with the bidi fixes.)

janbrasna:
_(Prettier is complaining about some formatting.)_

I can open the menu repeatedly meaning it’s no longer breaking execution. I however can not close it by clicking the menu item again, not sure if it was designed to do so, I can only dismiss it with X or by losing the focus elsewhere:

https://github.com/user-attachments/assets/57a1b59a-641e-46e9-b261-3f721d602016

stephaniehobson:
I don't think it was designed to since it doesn't work on mozorg either. It would be a nice enhancement though since I also expected tapping it again to close.

janbrasna:
Yeah I must have misread the code thinking it was a feature, but that's right it doesn't behave that way on bedrock. (Where, I wasn't really expecting it to, given the wider panels, one nav item closing out the other, etc., the X close button draws the attention there; it must have been the chevron here lending itself to being the open/close affordance 🤷 … I'll put it in my backlog for now, and eventually raise it as an enhancement once all the priority stuff is resolved / after the other dropdown panel is added etc.)

The only "issue" I found is on Android 8.0 out of the box (Chromium 103) with the nav hard to navigate being transparent; sharing that issue with bedrock. But since that OS can still update to present-day chromium versions this is really a non-issue. (Didn't get a chance to test on 7.0 that capped out at chromium 119; but since that one loses Play services this month, I pronounce it dead and unsupported anyways.)

LGTM. :shipit: 🚢

PR #253, opened 2025-06-04
remove 3rd party brand names from flt strings

stephaniehobson:
remove 3rd party brand names from flt strings

https://github.com/mozilla-l10n/www-firefox-l10n/pull/26#discussion_r2125075783


flodolo:
Unfortunately it's too late for this, we have already exposed strings in both Pontoon and Smartling.

You would need to change the string ID. At this point I suggest keeping these as is, and fix them in the future with the rest of string.

flodolo (changes requested):
See comment: we need new IDs, but at this point I would just accept that we have strings with extra Fluent terms, and fix them in the future.

flodolo:
Changes are possible while the PR importing strings in the l10n repository is still open.

In this case, I merged assuming these were stringed we ported from Bedrock, not brand-new strings.

Issue #180, opened 2025-05-23
Remove eyebrow from homepage


slightlyoffbeat:
## Description

"Firefox Browser" is redundant on the homepage and should be removed.

Before:

![Image](https://github.com/user-attachments/assets/727834f4-b851-4fc4-93db-dab07c697a1c)

after:

![Image](https://github.com/user-attachments/assets/007db347-9a21-4122-b239-f7d8ac595e28)

janbrasna:
Also note this is not only a part of new and old home, but also in mobile aside shown conditionally:

> ![fxios](https://github.com/user-attachments/assets/5acf6163-c1cd-4ec0-ada3-385c4b25fed7)

`aside#mobile-banner.show-android.show-ios`

https://github.com/mozmeao/springfield/blob/958a1cc81361c948e0a929e67a5a9e70fdbb605f/springfield/firefox/templates/firefox/download/desktop/download-en-us-ca.html#L215

https://github.com/mozmeao/springfield/blob/958a1cc81361c948e0a929e67a5a9e70fdbb605f/springfield/firefox/templates/firefox/download/desktop/download.html#L564

tesseheinricks:
Some additional options that satisfy the SEO concerns

![Image](https://github.com/user-attachments/assets/43b26d54-cb9b-42b5-a795-e2e3513f6d01)
![Image](https://github.com/user-attachments/assets/554101db-3c6c-411a-9d9b-26c63d16043f)

janbrasna:
Until a demo instance is available I'm attaching a few before/after screencaps of #259 for demonstration since the mockups don't exactly match the current type scale, base protocol theme etc., so to verify:

### US+CA home:

https://github.com/user-attachments/assets/f96c732a-249b-47c4-83d0-24af79c4009f

### GB-DE-FR-RoW home (and US+CA landing/get):

https://github.com/user-attachments/assets/cd5d15e7-8adf-4fb2-9a28-6c1cba8b69d9

### Mobile overlay: _(this is the only one impacting some dimensions)_

https://github.com/user-attachments/assets/e8e670e2-fb49-475a-9547-31ff588f9de7

### Basic experience: _(here the h1 is the actual heading there already)_

https://github.com/user-attachments/assets/579361f0-89a3-4492-b3c3-529604aad816

slightlyoffbeat:
Thank you Janbrasna. This all looks good to me.

Issue #184, opened 2025-05-25
New US/CA home jumps to footer on mobile #next tap


janbrasna:
## Description

Tapping a CTA to explore the desktop browser shown on mobile home leads to footer for US/CA locales:

> ![fxios](https://github.com/user-attachments/assets/5acf6163-c1cd-4ec0-ada3-385c4b25fed7)

### Steps to reproduce

(To repro set e.g. a fake Android UAstring to have the mobile aside shown, which includes both the link to trigger the fragment nav and the target id span as well.)

### Expected result

Old GB home: https://www-dev.springfield.moz.works/en-GB/ tapping on "Learn about the Firefox desktop browser below" jumps just to the MR106 section below: https://www-dev.springfield.moz.works/en-GB/#next

### Actual result

New US home: https://www-dev.springfield.moz.works/en-US/ tapping the same CTA/button tries to jump to https://www-dev.springfield.moz.works/en-US/#next which ends up scrolling to footer.

### Environment

Safari iPadOS 18.5

janbrasna:
Ah, it's the natural order that has changed. The og has

1. aside#mobile-banner
2. section#desktop-banner
3. section.t-releases

so the anchor jump "just works". The refresh page has the #mobile-banner aside moved to the end of the page, with the skip element also being implicitly positioned there as the target, hence why it scrolls to the bottom of the page.

If you prefer to keep the mobile aside content in the page order after the main content, the fix may be adding a display:block to the empty span, which would (invisibly) render it at the coords of the aside, making the anchor target scroll to that spot between the aside and the section.

(Nit: with the mobile aside at the end, the references to "below" and link to further down the page when the content section is actually way above that aside makes the outlined/unstyled view and flow feel somewhat broken. But understand the intention to push the actual desktop content as high as possible.)

stephaniehobson:
The mobile blurbs should continue to be shown, the link should continue to go to the release notes. 😬

stephaniehobson:
Replicated in Safari on desktop so this is a Safari bug based on content order - I think.

PR #259, opened 2025-06-05
Eliminate repetitive wordmarks

janbrasna:
## One-line summary

Turns home h1 into plaintext, removes mobile h2 (there are better h2s right after it) and removes decoration from basic experience.

These are the ~6 occurrences, most of which benefit from the shorter hero, however for mobile aside this makes the contained screenshot smaller, so make sure that's ok.

(I haven't removed the one on mobile product pages as that one needs to alternate between Firefox and Focus wordmarks so it's kept for consistency. If there were "Firefox for Android" or "Firefox for iOS" actual wordmarks, these would be a better fit there, to make it immediately obvious it's a mobile product…)

## Issue / Bugzilla link

#180

## Testing

http://localhost:8000/en-US/
http://localhost:8000/en-GB/
http://localhost:8000/en-US/landing/get/
http://localhost:8000/he/landing/get/
http://localhost:8000/skr/
http://localhost:8000/ur/?xv=basic
http://localhost:8000/ro/?xv=basic
(+ the same in mobile /=spoof android UA/ for aside#mobile-banner)

### DIFF
--- a/l10n/en/firefox/download/desktop.ftl
+++ b/l10n/en/firefox/download/desktop.ftl
@@ -109,7 +109,6 @@ firefox-desktop-download-screenshots = Screenshots
 # Enhanced Tracking Protection is a feature name and so is capitalized in English
 firefox-desktop-download-enhanced-tracking-protection = Enhanced Tracking Protection (ETP)
 firefox-desktop-download-from-watching-a = From watching a web tutorial to keeping an eye on your favorite team, your video follows you while you multitask.
-firefox-desktop-download-firefox-browser = { -brand-name-firefox-browser }
 firefox-desktop-download-get-firefox-android = Get { -brand-name-firefox-browser } for <strong>{ -brand-name-android }</strong>
 firefox-desktop-download-get-firefox-ios = Get { -brand-name-firefox-browser } for <strong>{ -brand-name-ios }</strong>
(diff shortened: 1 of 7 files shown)


janbrasna:
@stephaniehobson Thanks, haven't noticed it at first being pushed with a ref outside of this PR but thanks for including any updates from main with it, that was very thoughtful, appreciate it. I'm not sure if this is expected to get some feedback from stakeholders first, so I included some screencaps in the tracking issue to demonstrate the difference there, if anyone cares that much.

That said, the only questionable thing is the mobile alt hero, as it has the image background sized based on the length of the content, so with a shorter content the image gets smaller. I'm not even sure the purple hero `aside#mobile-banner` was intended to stay (see https://github.com/mozmeao/springfield/issues/184#issuecomment-2921536565) so chances are it will get removed eventually(?) — noting that before digging deeper into moving the background image out of the content container and positioning it somehow to the parent hero directly to gain some more height back.

stevejalim:
Based on this comment, I think we can approve and merge this

https://github.com/mozmeao/springfield/issues/180#issuecomment-2961236409

janbrasna:
@stevejalim The overall style got approved, so this only awaits a FE review to check there's not something horrendously forgotten somewhere, a fluent ref removed too early etc. … or ofc you can do the honours if you feel like, in case a strict `build-prod` (done already for demo1 above) and a push to `test` is enough to verify this viable for now.

Issue #17, opened 2025-02-12
Review Switches and set to On where relevant


stevejalim:

janbrasna:
Looking at _dev_ & _demo_ (where any switch is evaluated as enabled, so there only disabled switches make sense, e.g. newsletter maintenance mode etc.), _stage_ and _prod_ according to their `healthz-cron` lists, these all seem to have been cleaned up in the recent past.

What is generally missing:

1. `SHOW_FIREFOX_APP_STORE_BANNER`

Some [screenshots](https://github.com/mozilla/bedrock/pull/14942) what they are for context.
These are currently shown only on _dev_ implicitly, as _stage/prod_ doesn't enable them. _(Spoof Android UAstring in devtools to check.)_

2. `DOWNLOAD_FIREFOX_MR106_PROMO`  ([context](https://github.com/mozilla/bedrock/issues/14863))

Comparing _dev_ that enables this by default, and _prod_ keeping this as disabled, yields different results: _(compare "Latest Firefox features" section:)_
https://www-dev.springfield.moz.works/en-GB/ (look for "Edit your PDFs directly")
https://www.springfield.moz.works/en-GB/ (look for "Expanded Dark Mode")

3. `BROWSER_MR2_PROMO`

Related to above, when the MR106 is intended to be disabled, then this one changes the middle card "Expanded Dark Mode" to newer "Choose your color", which you can see is also the older on prod.

The 2+3 affect e.g. GB and DE home, and landing/get even for US/CA.

Instead of bringing the 2.) & 3.) switches back, my guess is that for the locales listed in the template it'd be ok to hardcode the conditions like these are active for now — as these pages will be most likely replaced with what's currently designed for US+CA.

stevejalim:
`SHOW_FIREFOX_APP_STORE_BANNER` doesn't appear in code in Springfield, and #264 handled the other two, so i think we can close this.

PR #264, opened 2025-06-06
Fix outdated product features dropping MR106/MR2 switch logic

janbrasna:
## One-line summary

Fixes outdated feature listing, originally driven by switches, when in prod mode.

The switches are now gone, changing the content shown when `if DEV` or not (as waffle default for _dev_ is True, but _stage_ and beyond is False)…

Not worth opening for general l10n as the page is most likely going away for all locales.

(Removing content & assets now inaccessible behind the versioning.)

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/17#issuecomment-2946007181

## Testing

verify that "Latest Firefox features" box has the same content for `DEV=True` and `DEV=False`:
http://localhost:8000/en-GB/
http://localhost:8000/de/
http://localhost:8000/cy/ (no box)
http://localhost:8000/en-US/landing/get/?geo=us
http://localhost:8000/fr/landing/get/?geo=ca


stephaniehobson (approved):
R+

PR #271, opened 2025-06-09
(fix) Tidy up 30x status codes for redirects ported from Refractr

### LINKED ISSUE #270: Vanity `firefox.com/pair` redirect should serve 301 or 302 status?
Since 301s are heavily cached, and this doesn't really appear on the web per se — the link being primarily shown in mobile app onboarding screens as a step to follow manually on another device:

https://github.com/mozilla-mobile/firefox-ios/blob/v139.0/firefox-ios/Shared/en.lproj/Localizable.strings#L1178

([Screenshots of the flow](https://redirect.github.com/mozilla/fxa/issues/18061#issuecomment-2629515601))

there might not be a reason to actually return it as 301 permanent (as added here), but just 302 temporary (as currently served):

_(prod refract:)_
```http
~ $ https -v HEAD firefox.com/pair
HEAD /pair HTTP/1.1
Host: firefox.com

HTTP/1.1 302 Found
location: https://accounts.firefox.com/pair/
server: nginx
```

_(stage springfield:)_
```http
~ $ https -v HEAD www.springfield.moz.works/pair
HEAD /pair HTTP/1.1
Host: www.springfield.moz.works

HTTP/1.1 301 Moved Permanently
X-Served-By: cache-lga21987-LGA, cache-lga21945-LGA
cache-control: max-age=43200
expires: Mon, 09 Jun 2025 20:52:18 GMT
location: https://accounts.firefox.com/pair/
```

I can fold the status code change with some other updates in #269 if favourable, as that one also adds some code checks for vanity URLs and this one is commented out for now.

stevejalim:
Because by default our redirect() helper uses 301s and there may be reasons why refractr was 302ing, so this changeset makes sure that some things are still 302ed where it makes sense, particularly if we're likely to update them further as they won't be cached in the same way.

Most importantly ensures that /pair/ -> accounts.firefox.com/pair/ is a 302

## Issue / Bugzilla link

Resolves #270

## Testing

Check any of the URLs amended in `springfield/firefox/redirects.py`


janbrasna (approved):
r+

_(whatever jumps out at me in the url PR upon expanding on these I'll follow up individually there when the time comes)_

Issue #235, opened 2025-06-03
Add top 3-4 Features to the Features dropdown + "view all"


slightlyoffbeat:

janbrasna:
For now I have this:

> ![nav](https://github.com/user-attachments/assets/d95c5d54-eefa-496d-8bc9-1d0cf3a3b59c)

I'll continue tweaking the spacing and when I'm happy with it I'll post it for review to get it in front of more folks 👀 

(Also! I still want to animate the chevron 180° rotating on hover, see if I get to it too…)

slightlyoffbeat:
I'll tag @tesseheinricks to help with above questions

janbrasna:
Ah, that was not color change, but a font weight change, see it now. got it. Thanks for the clarification. I think we had to scrape the font-weight change on hover for now as it's not supported by the fallback font for other languages — but I've seem some preliminary work to get more countries supported in the Mozilla family so we can perhaps enable that selectively across locales at a later point. (Nerdy detail: The new font comes in "uniwidth" design, meaning it doesn't change dimensions when changing styles. That is nonetheless not the norm, so it makes the navigation jump in languages that don't have enough glyph support to be set in Mozilla type. /This can be already seen in such languages when loading/…)

I'll look into the changes to all the menu panels first, and then circle back to the new content. Thank you.

janbrasna:
@stephaniehobson I'll see how far I'll get eventually; the only thing I don't know looking at it would be the hover state for the whole "View all Firefox features", guessing it's a color change, and using something we have in the palette on m24 root vars. All ask if I get that far;)

janbrasna:
I've added a hover animation for the arrow so check it's appropriate to keep:

https://github.com/user-attachments/assets/b80e2886-0c13-442e-b197-3f1911255b83

I'm unsure about the mobile layout. This is how it renders currently:

> <img width="608" height="497" alt="Image" src="https://github.com/user-attachments/assets/139f40da-9c46-472c-bfaa-6d31fe15fbca" />

Besides the logomark and × close symbol not having the correct margins (preexisting issue) the new header color now make the first and second level headings render pretty much the same. So I'm wondering about next steps for these.

The design calls for encapsulating the submenu in a grey box — is that intended only for "Features" subnav, or both panels? If that's intended for both, maybe that separation would help distinguishing the two heading levels.

PR #288, opened 2025-06-16
Add "Features" menu dropdown

janbrasna:
## One-line summary

Add menu items for "Features" navigation item.

Some headings and colors were updated with that design.

New treatment was designed for mobile panels.

Q: mobile design calls for grey inset background (only for "Features" or any nav panel?) — not being sure now how to distinguish top nav heading link vs. panel heading without a link both looking the same currently in the mobile nav, I applied the grey background to both panels to make the two heading levels more distinct.

(Fixed some whitespace in surrounding files with it.)

Changed the chevron from sprite to single icon & rotate -180deg on hover.

## Issue / Bugzilla link

#235

## Testing

### DIFF
--- a/media/css/m24/components/navigation-refresh.scss
+++ b/media/css/m24/components/navigation-refresh.scss
@@ -190,7 +190,7 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
 
     a:link,
     a:visited {
-        color: $color-black;
+        color: $token-color-off-black;
         text-decoration: none;
     }
 
@@ -318,15 +318,13 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
 }
 
 .m24-c-menu-category-link {
-    display: flex;
-    font-size: $text-body-md;
-    justify-content: flex-end;
-    margin: $spacer-md 0 0;
-    padding: 0;
-
     a:link,
     a:visited {
         display: flex;
+        justify-content: flex-start;
+        margin: $spacer-md 0 0;
+        padding: $spacer-sm;
+        background: var(--m24-light-orange);
         font-weight: 600;
         position: relative;
         width: fit-content;
@@ -348,15 +346,17 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
         transition-timing-function: $bezier;
         width: auto;
     }
-}
 
-.m24-c-menu-category-link a:hover,
-.m24-c-menu-category-link a:focus {
-    svg {
-        @include bidi((
-            (right, -$spacing-sm, auto),
-            (left, auto, -$spacing-sm),
-        ));
+    a:hover,
+    a:focus {
+        background: var(--m24-orange);
+
+        svg {
+            @include bidi((
+                (right, -$spacing-sm, auto),
+                (left, auto, -$spacing-sm),
+            ));
+        }
     }
 }
 
@@ -432,7 +432,7 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
 }
 
 .m24-c-menu-subtitle {
-    color: $m24-color-dark-mid-gray;
+    color: $m24-color-black;
     margin: 0;
     padding: $spacing-md 0 $spacing-sm;
 }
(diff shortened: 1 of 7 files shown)


janbrasna (commented):
I've marked as ready for review to start getting some feedback on the design, and mostly to approve the content of the menu items first.

It's **_DNM_** for now as it still lacks wrapping the strings in ftl.

@stephaniehobson I've added the mobile panel treatment (grey inset box) in a separate commit that you can reset/revert if that's not what was designed/intended — feel free to push to the PR as you see fit.

media/css/m24/components/navigation-refresh.scss
```diff
@@ -221,16 +221,18 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
...
             @include bidi(((margin-left, $spacing-sm, margin-right, 0),));
             content: '';
-            background: url('/media/img/icons/caret-sprite.svg') no-repeat bottom left / auto 12px;
+            background: url('/media/protocol/img/icons/arrow-down.svg') no-repeat center center / auto 16px;
             display: inline-block;
             height: 12px;
             vertical-align: baseline;
             width: 12px;
+            transform-origin: center;
+            transition: transform 0.25s ease;
```
janbrasna, line 230: Ideally these would only apply based on prefers-reduced-motion media query…

media/css/m24/components/navigation-refresh.scss
```diff
@@ -318,21 +324,48 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
...
-    margin: $spacer-lg $spacer-lg 0;
-    padding: 0;
-
     a:link,
     a:visited {
         display: flex;
+        justify-content: space-between;
+        margin: $spacer-md 0 -1em;
+        padding: $spacer-sm;
+        background: var(--m24-light-orange);
```
stephaniehobson, line 333:
```suggestion
        background: $m24-color-light-orange;
```

Let's use the tokens instead of CSS variables for now. I know we're planning to go all in on CSS vars but I don't want a mix of them to manage until we make the switch

media/css/m24/components/navigation-refresh.scss
```diff
@@ -318,21 +324,48 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
...
+        position: relative;
+        top: 0.15em;
+        transition-duration: $fast;
+        transition-timing-function: $bezier;
+        width: auto;
+    }
+
+    a:hover,
+    a:focus {
+        background: var(--m24-orange);
```
stephaniehobson, line 357:
```suggestion
        background: $token-color-orange;
```

stephaniehobson (approved):
R+ wc

Looks good. Thanks for following through on this one.

media/css/m24/components/navigation-refresh.scss
janbrasna: This definitely was not a design choice, most likely just a WIP — or I wasn't able to locate the right token from the right lib that would be loaded at that point;) Thanks for getting the correct ones for me.

media/css/m24/components/navigation-refresh.scss
janbrasna: Hmm I did remove the transition on the chevron rotation, but forgot about the slower arrow movement on the orange category link hover…

Do we know a good best practice on frequency/length etc. that might be preferable to avoid, and/or if some movement is OK… or I should better follow up with removing any movement on the arrow hover to be on the safe side?

media/css/m24/components/navigation-refresh.scss
```diff
@@ -221,16 +221,22 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
...
+
+            @media (prefers-reduced-motion: reduce) {
+                transition: none;
+            }
         }
 
         &.mzp-is-selected {
             h2::after {
-                background-position: bottom right;
+                transform: rotate(-180deg);
```
janbrasna, line 239: Now I'm realizing the previous WMO nav had the rotation "on the inside" instead:

https://github.com/mozilla/bedrock/blob/2025-06-25/media/css/protocol/components/_menu.scss#L186

_(clockwise vs. counterclockwise)_

(Ideally that should get flipped the other direction for RTL but I admittedly skipped that here for simplicity.)

Is there a preference for one or another, or attempts at consistency even?

stephaniehobson: It's the previous nav, I don't think there's any point trying to be consistent with it.

PR #318, opened 2025-06-19
Hotfix ci tests

stevejalim:
* Fixes CDN test failure
* Re-enables scheduled Download and CDN tests

### DIFF
--- a/springfield/firefox/redirects.py
+++ b/springfield/firefox/redirects.py
@@ -105,4 +105,6 @@ redirectpatterns = (
         r"/(?P<channel>release|aurora)notes/(?P<page>[\/\w\.-]+)?$",
         "http://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-US/{prod}/{vers}/{channel}notes/{page}",
     ),
+    # Bug 868182
+    redirect(r"^mobile/faq/?$", firefox_mobile_faq, query=False),
 )
--- a/tests/redirects/map_all_redirects.py
+++ b/tests/redirects/map_all_redirects.py
@@ -61,5 +61,7 @@ URLS = flatten(
         # bug 988746, 989423, 994186, 1153351
         url_test("/mobile/{23,28,29}.0/releasenotes/", "/firefox/android/{23,28,29}.0/releasenotes/"),
         url_test("/mobile/{3,4}2.0beta/{aurora,release}notes/", "/firefox/android/{3,4}2.0beta/{aurora,release}notes/"),
+        # bug 868182
+        url_test("/mobile/faq/?os=firefox-os", "https://support.mozilla.org/products/firefox-os"),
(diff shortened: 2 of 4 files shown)


springfield/firefox/redirects.py
```diff
@@ -105,4 +105,6 @@ def mobile_app(request, *args, **kwargs):
         r"/(?P<channel>release|aurora)notes/(?P<page>[\/\w\.-]+)?$",
         "http://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-US/{prod}/{vers}/{channel}notes/{page}",
     ),
+    # Bug 868182
+    redirect(r"^mobile/faq/?$", firefox_mobile_faq, query=False),
```
janbrasna, line 109: `firefox_mobile_faq` seems actually quite bizarre to keep; but since it's been in the refracts, it might be a hint it actually was perhaps used at some point with fxc. Can't really find any source that would confirm these URLs were ever only on mozorg or also on fxdotcom … so they're probably to stay?

tests/redirects/map_all_redirects.py
```diff
@@ -61,5 +61,7 @@
         # bug 988746, 989423, 994186, 1153351
         url_test("/mobile/{23,28,29}.0/releasenotes/", "/firefox/android/{23,28,29}.0/releasenotes/"),
         url_test("/mobile/{3,4}2.0beta/{aurora,release}notes/", "/firefox/android/{3,4}2.0beta/{aurora,release}notes/"),
+        # bug 868182
+        url_test("/mobile/faq/?os=firefox-os", "https://support.mozilla.org/products/firefox-os"),
```
janbrasna, line 65:
```suggestion
        url_test("/mobile/faq/?os=firefox-os", "https://support.mozilla.org/products/firefox-os"),
        url_test("/mobile/faq/", "https://support.mozilla.org/products/mobile"),
```

The other one to double check too?

janbrasna (approved):
I have seen these somewhere in the past and still wondering where, and why were omitted from porting. But makes sense given the helper is actually still there.

springfield/firefox/redirects.py
janbrasna: I always pause a bit when I see `query=False` ~ it does something else than what it may seem: it means "add these query parameters: none", not "remove any query string when redirecting" (it is the default value IIUC), so it being often added on redirects that actually work with some query params is a little puzzling. ~ Q: does this serve any purpose I failed to infer, or would it be okay if, when I'm in the neighborhood (as I have a few more basic redirect tests that I ought to add) and removing the defaults doesn't break anything, if I removed these in cases they actually make understanding the rules harder?

janbrasna: yes, I gotta try and test that assumption first anyway.

PR #319, opened 2025-06-19
Fix homepage navigation tests with wildcards and un-wildcard some others

janbrasna:
## One-line summary

Allow cases where e.g. extra `utm_campaign` param is being added to the location.

```
specs/footer.spec.js:25:9 › /en-US/ footer (mobile) › Footer language change @firefox 
    =========================== logs ===========================
    waiting for navigation to "**/de/?automation=true" until "commit"
      navigated to "https://stage.springfield.nonprod.webservices.mozgcp.net/de/?automation=true&utm_campaign=SET_DEFAULT_BROWSER"
```

Right now we don't exactly care how the query string looks like — and since the home is also download page now, it can have a varying degree or parameters added and removed etc.

Ideally it should not happen with `?automation=true` but this is a hotfix for now.

(This exact failure is weird given it comes from a Linux UA that should not be triggering said campaign, but that's another question.)

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/actions/runs/15740357974/job/44406845294#step:5:532

### DIFF
--- a/tests/playwright/specs/404.spec.js
+++ b/tests/playwright/specs/404.spec.js
@@ -25,7 +25,7 @@ test.describe(
             const goBackLink = page.getByTestId('link-go-back');
             await expect(goBackLink).toBeVisible();
             await goBackLink.click();
-            await page.waitForURL('**/?automation=true', {
+            await page.waitForURL('**/?automation=true*', {
                 waitUntil: 'commit'
             });
         });
--- a/tests/playwright/specs/404.spec.js
+++ b/tests/playwright/specs/404.spec.js
@@ -8,10 +8,10 @@
 
 const { test, expect } = require('@playwright/test');
 const openPage = require('../scripts/open-page');
-const url = '/en-US/';
+const url = '/en-US/compare/';
 
 test.describe(
-    `${url} page`,
+    `404 page`,
     {
         tag: '@firefox'
     },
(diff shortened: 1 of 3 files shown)


stevejalim:
Let's wait for integration test to pass first then we can merge

tests/playwright/specs/404.spec.js
```diff
@@ -25,7 +25,7 @@ test.describe(
             const goBackLink = page.getByTestId('link-go-back');
             await expect(goBackLink).toBeVisible();
             await goBackLink.click();
-            await page.waitForURL('**/?automation=true', {
+            await page.waitForURL('**/?automation=true*', {
```
janbrasna, line 28: If there's no rush would this be an actual improvement while there:

```suggestion
            await page.waitForURL('**/en-US/?automation=true*', {
```

to not pass literally "any" page, incl. 404s and others?

stevejalim: Actually, this is enough here - have tested locally against Staging
```suggestion
            await page.waitForURL('/en-US/?automation=true*', {
```

stevejalim: But let's go with your suggestion, np

stevejalim (approved):

PR #322, opened 2025-06-24
Update mobile QR codes to FXC URLs

### LINKED ISSUE #277: Update URLs in QR codes for mobile redirects to reference www.firefox.com not www.mozilla.org/firefox/
## Description

In https://github.com/mozmeao/springfield/pull/266 we moved to programatically-generated QR codes for all.

During that work, we noticed that some QRs (for the mobile app redirect) point to www.mozilla.org. This works, but once www.firefox.com is live, we should change them to point to there instead. 

If we wanted to do it earlier, we'd have to inject the relevant non-production-domain hostname in (eg www-dev.springfield.moz.works) which feels like a bit more noise than necessary. 

The actual mobile redirection logic IS present in Springfield, so once it's already serving pages on www.firefox.com, changing over to the new domain will be easy to test locally.

---

#### Success Criteria
- [ ] QR codes don't send people to www.mozilla.org, they send to www.firefox.com or the relevant mobile store (as current behaviour)

janbrasna:
## One-line summary

Generates QR codes with links going through prod domain app redirector, or uses store helpers.

This means that where locale is a factor in the target URL, a new/different QR code would be generated, so a handful of them will be actually created and cached, given this outputs a wide variety of locales/hrefs.

A side effect of that is that the target store URL is not opened in the locale of the device scanning the code, but in the locale the user was already browsing the site on desktop that generated the QR code for them.

(Thoughts? The app store links could ofc be also hard-coded to be locale-neutral and do the content negotiation like before, this is just easier to manage incl. referal query params etc. all coming from the single source of truth in the helper setup. And the QR code link matches the URL of the button shown next to it, too.)

## Issue / Bugzilla link

Resolves #277

## Testing

http://localhost:8000/he/download/all/mobile-release/
http://localhost:8000/zh-TW/browsers/mobile/focus/

compare before/after (prod vs. PR) e.g. these two:
http://localhost:8000/pt-PT/browsers/mobile/android/
http://localhost:8000/el/browsers/mobile/android/
(these would show the same QR on prod, but differ here)

this one has language and differs:
http://localhost:8000/nl/browsers/mobile/ios/
whereas these two end up without language so are the same:
http://localhost:8000/cy/browsers/mobile/ios/
http://localhost:8000/dsb/browsers/mobile/ios/

### DIFF
--- a/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-android.html
+++ b/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-android.html
@@ -6,7 +6,7 @@
 
 {# Send-to-device widget for Firefox for Android #}
 
-{% set qr_url = "https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral%26utm_campaign%3Dfirefox-browsers-mobile-android" %}
+{% set qr_url = play_store_url('firefox', 'firefox-browsers-mobile-android') %}
 
 <div class="c-inline-form">
   <h2 class="c-inline-form-title">{{ ftl('mobile-android-get-firefox-for') }}</h2>
--- a/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-android.html
+++ b/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-android.html
@@ -6,7 +6,7 @@
 
 {# Send-to-device widget for Firefox for Android #}
 
-{% set qr_url = play_store_url('firefox', 'firefox-browsers-mobile-android') %}
+{% set qr_url = "https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral%26utm_campaign%3Dfirefox-browsers-mobile-android" %}
 
 <div class="c-inline-form">
   <h2 class="c-inline-form-title">{{ ftl('mobile-android-get-firefox-for') }}</h2>
--- a/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-ios.html
+++ b/springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-ios.html
@@ -6,7 +6,7 @@
 
 {# Send-to-device widget for Firefox for iOS #}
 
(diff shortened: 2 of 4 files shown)


stephaniehobson:
I do think we should hard-code them to be language neutral. The user's device probably knows best.

springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-android.html
```diff
@@ -6,7 +6,7 @@
 
 {# Send-to-device widget for Firefox for Android #}
 
-{% set qr_url = "https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral%26utm_campaign%3Dfirefox-browsers-mobile-android" %}
+{% set qr_url = play_store_url('firefox', 'firefox-browsers-mobile-android') %}
```
janbrasna, line 9:
```suggestion
{% set qr_url = "https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral%26utm_campaign%3Dfirefox-browsers-mobile-android" %}
```

springfield/firefox/templates/firefox/browsers/mobile/includes/s2d-ios.html
```diff
@@ -6,7 +6,7 @@
 
 {# Send-to-device widget for Firefox for iOS #}
 
-{% set qr_url = "https://apps.apple.com/app/apple-store/id989804926?pt=373246&ct=firefox-browsers-mobile-ios&mt=8" %}
+{% set qr_url = app_store_url('firefox', 'firefox-browsers-mobile-ios') %}
```
janbrasna, line 9:
```suggestion
{% set qr_url = "https://apps.apple.com/app/apple-store/id989804926?pt=373246&ct=firefox-browsers-mobile-ios&mt=8" %}
```

janbrasna (commented):
I haven't verified whether that's something that can be parametrized in the helpers. ~Going to look now. In case you know the answer already~ Yeah that'd have to be customized to accept arbitrary locales, or none etc.; probably not worth it, so…  (if you don't mind the ugly campaigns and app IDs sprinkled here and there) below are the buttons to just revert these two:

PR #326, opened 2025-06-24
Fix home mobile #next anchor for refresh locales

### LINKED ISSUE #184: New US/CA home jumps to footer on mobile #next tap

janbrasna:
## One-line summary

Moving the id on the actual element to land on.

No matter where in DOM the banners are repositioned, this keeps their target reasonable in the document outline.

## Issue / Bugzilla link

Fixes #184

## Testing

http://localhost:8000/en-CA/ with UA triggering mobile aside
(or http://localhost:8000/en-CA/#next in general)

### DIFF
--- a/springfield/firefox/templates/firefox/download/desktop/download-en-us-ca.html
+++ b/springfield/firefox/templates/firefox/download/desktop/download-en-us-ca.html
@@ -94,7 +94,7 @@ Set Firefox as your default browser.
   </section>
 
 
-<section class="mzp-l-content t-releases">
+<section class="mzp-l-content t-releases" id="next">
   <div class="mzp-c-emphasis-box js-animate">
     <h2 class="mzp-c-section-heading">Latest Firefox features</h2>
     <ul class="c-trio">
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Not all languages display the release notes section so the 3 of us should watch to make sure this does not regress when we add the translations.

janbrasna:
This will probably change over time before it lands here too, but judging from WIP [mozilla/bedrock#16350](https://github.com/mozilla/bedrock/pull/16350/files#diff-d394f5ad373cc1a39331391e842e672c16c06499d3e926246bf8021c95011463R98) I went with the assumption this block is not behind locale activation conditions — and given it will land as a whole I'm guessing it won't get anything fancy of that sort. (But a lot of parts will be moving one way or another so def 👓 measure twice; I'll keep an eye: I use weird locales on weird devices all the time anyway;D)

PR #327, opened 2025-06-25
Bump mobile aside screen size

janbrasna:
## One-line summary

Wastes less space when presenting mobile device screens on home for mobile useragent visitors.

Following up from #259 that made the home/landing aside#mobile-banner on mobile devices shorter, which also resulted in less tall =smaller device screen shots, this adds some of the size and legibility back.

Oh, also, didn't like the symmetrical whitespace around the device as that itself is somewhat unbalanced with the menubars at the bottom etc. — so I've taken the liberty to make the call here and actually make the bottom margin smaller than the top margin, because the alignment was bothering me all the time.

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/180#issuecomment-2950439527

## Testing

/en-CA
/en-GB

### DIFF
--- a/media/css/firefox/home-en-us-ca.scss
+++ b/media/css/firefox/home-en-us-ca.scss
@@ -641,11 +641,11 @@ button.mzp-c-cta-link {
             background-position: top center;
             background-repeat: no-repeat;
             border-radius: $border-radius-sm;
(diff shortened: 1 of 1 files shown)


stephaniehobson (approved):
R+ 📱

Issue #324, opened 2025-06-24
[Sentry Error] Error Cannot read properties of null (reading 'addEventListener')


stephaniehobson:
```
TypeError
Cannot read properties of null (reading 'addEventListener')
```

Stacktrace suggests it is coming from here:
https://github.com/mozmeao/springfield/blob/ba3f4869c3d7fbfccb5d0f315ab9cc844fe533cd/media/js/base/banners/m24-pencil-banner.es6.js#L81

**Location:**
Happening on `/`

**Browser:**
All

**OS:**
All

[Full error report.](https://mozilla.sentry.io/issues/6706102548/?project=4508966604242944&query=is%3Aunresolved&referrer=issue-stream&sort=freq&stream_index=1)

stephaniehobson:
Fixed in https://github.com/mozmeao/springfield/pull/329

PR #329, opened 2025-06-26
Only bind pencil-banner setup if it has close button

### LINKED ISSUE #328: JS breaks on some locales
## Description

While navigating between translations I noticed after some locales I'm no longer able to switch via JS select in footer. Turns out the page errors out on load with `TypeError: Cannot read properties of null (reading 'addEventListener')` stopping further execution.

**This seems to only affect DEV** 👷 as the same STR worksforme on stage+prod. So it's either switches, locales, or code not pushed further than main;]

### Steps to reproduce

Open https://www-dev.springfield.moz.works/skr/ and check console.

### Expected result

I should be able to change language or do whatever the UI script was attaching to.

### Actual result

Some missing element from the page unexpectedly?

Image

Could it be that pencil banner JS expects the banner to be there, but it's also gated by l10n threshold so it's not rendered?

https://github.com/mozmeao/springfield/blob/a34b9d45659c073e94d0b45996c7e3713f869bfe/media/js/base/banners/m24-pencil-banner.es6.js#L78-L85

### Environment

localhost, www-dev
FXDE, chromium

janbrasna:
## One-line summary

Only init pencil banner that is actually interactive.

This skips pushing the GA datalayer in case the banner is used only as a form of notification when can't be dismissed etc. — sounds OK to me.

## Issue / Bugzilla link

Fixes #328 (and possibly #324 too)

## Testing

http://localhost:8000/en
http://localhost:8000/skr
http://localhost:8000/eo

### DIFF
--- a/media/js/base/banners/m24-pencil-banner.es6.js
+++ b/media/js/base/banners/m24-pencil-banner.es6.js
@@ -96,7 +96,7 @@ M24PencilBanner.init = function () {
         typeof window.Mozilla.Cookies !== 'undefined' &&
         window.Mozilla.Cookies.enabled();
 
-    _pencilBanner = document.querySelector('.m24-pencil-banner');
+    _pencilBanner = document.querySelector('.m24-pencil-banner:has(.m24-pencil-banner-close)');
 
     // If the banner does not exist on a page then do nothing.
     if (!_pencilBanner) {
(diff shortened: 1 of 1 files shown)


stephaniehobson (approved):
R+

PR #331, opened 2025-06-26
Refactor pencil banner button bind for legacy IE

janbrasna:
## One-line summary

Rewrites #329 in a more conservative way to avoid _SyntaxError_ on legacy IE.

Not that the button would work in IE to start with. It just seems to choke badly on a selector it doesn't understand in the JS query… TIL

## Issue / Bugzilla link

#324

## Testing

[/actions/runs/15898256192/job/44834926423](https://github.com/mozmeao/springfield/actions/runs/15898256192/job/44834926423) :shipit: 🚢 ✅


stevejalim:
r+ let's merge this and try it on www-dev

PR #334, opened 2025-06-27
Mark missing SSL test results as `xfail`

janbrasna:
## One-line summary

If there’s no results from upstream API, these are expected to fail. Let’s mark them as such in that case, so they don’t fail the whole pipeline which is still very useful. 

Also triggers in a different time slot than bedrock so we’re not getting unnecessarily rate-limited etc.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/16339

### DIFF
--- a/.github/workflows/cdn_tests.yml
+++ b/.github/workflows/cdn_tests.yml
@@ -7,7 +7,7 @@ env:
   SLACK_BOT_TOKEN: ${{secrets.SLACK_BOT_TOKEN_FOR_MEAO_NOTIFICATIONS_APP}}
 on:
   schedule:
-    - cron: "0 11 * * *" # 11am daily
+    - cron: "50 11 * * *" # 11:50 AM UTC daily
   workflow_dispatch:
     inputs:
       springfield_service_hostname:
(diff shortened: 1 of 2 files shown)


stevejalim:
@janbrasna What situtation occurrs where those SSL results are not available but the rest of the tests are worth carrying on with? Just double-checking whether xfail here could actually give us a false positive

janbrasna:
@stevejalim Qualsys has hiccups so they often don’t return compete data. That’s for three or four SSL tests (about six parametrized).

But CDN tests have more things than just SSL, like really important CDN sanity checks, about six of them — checking that locales are negotiated properly and not cached without vary, statics are served from the GCS bucket not the app, statics are hooked up to the static URL content and served, query strings reach the app, cache is cached reasonably enough etc.

janbrasna:
I mean, if there ARE results from SSL, and then THESE don’t match, that’s a valid fail. I’m actually trying to make THIS case the visible amongst all the noise, and have a loud fail when the expectations disagree from reality. (But then, the expectations are not set by us, but by the server handshakes priming the Qualsys baseline list derived from its current TLS configuration, so if that one changes the probe won’t really be loud about it — happy to elaborate more one day. That’s why I’m contemplating moving this to a more simple and more hard coded check that sets the expectations explicitly, more in the original ticket.)

So this is just to distract from unsuccessful probe runs, for folks to not expect this to be that flaky failing every other day, to actually draw everyone’s attention in case it really fails.

The only “false positive” danger here is the service goes away completely and the test is never run again… but still reporting happy every day.

But then, would anyone notice?;) This is currently running expecting to not have TLS1.3, which def is available, but it’s still happy… So I’m giving this a little attention overall here and there, to improve a little bit over time.

PR #345, opened 2025-07-02
(security) Dep bump for urrlib primarily, plus other updates

stevejalim:
Selenium was blocking the urllib upgrade, so I've downgraded selenium and pytest-selenium to match the versions we use on Bedrock, which allows the upgrade to urllib3, surprisingly


stevejalim:
Integration tests passed fine for this branch, with the selenium change https://github.com/mozmeao/springfield/actions/runs/16020789297

PR #348, opened 2025-07-04
Fix `test_robots_no_redirect` with `@override_switch decorator`

### LINKED ISSUE #347: Failing test, `test_robots_no_redirect` when `DEV=True`
The `test_robots_no_redirect` test in Springfield is failing due to `django-waffle` switch behavior. The test expects `disallow_all` to be `False` when accessing robots.txt with hostname `www.firefox.com`, but it's getting `True` instead.

### Steps to reproduce

1) Set up Springfield locally with `DEV=True` in `.env`
2) Run the test: `pytest springfield/base/tests/test_views.py::TestRobots::test_robots_no_redirect -v`

### Expected result

The test should pass, with `response.context_data["disallow_all"]` being `False` when the hostname is `www.firefox.com`.

### Actual result

The test fails with:

```
FAILED springfield/base/tests/test_views.py::TestRobots::test_robots_no_redirect - AssertionError: True is not false
```

### Environment

Operating System: macOS (darwin 24.3.0)
Python: 3.13.3
Django: 5.2.2
Environment: Local development with `DEV=True` in .env

localjo:
Add the `@override_switch("ROBOTS_FORCE_DISALLOW_ALL", active=False)` decorator to the test to explicitly set the switch value, matching the pattern used by other tests.

## Issue / Bugzilla link

Fixes #347

### DIFF
--- a/springfield/base/tests/test_views.py
+++ b/springfield/base/tests/test_views.py
@@ -77,6 +77,7 @@ class TestRobots(TestCase):
         self.view.request = self.rf.get("/", HTTP_HOST="www.springfield.moz.works")
         self.assertTrue(self.view.get_context_data()["disallow_all"])
 
+    @override_switch("ROBOTS_FORCE_DISALLOW_ALL", active=False)
     def test_robots_no_redirect(self):
         response = self.client.get("/robots.txt", headers={"host": "www.firefox.com"})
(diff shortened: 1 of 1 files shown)


stevejalim (approved):
Makes sense - thanks for firming up that test

r+

PR #359, opened 2025-07-10
Fix Internet (… string case in fluent)

janbrasna:
## One-line summary

For consistency with other occurrences.

Using the same v2 string ID, as the update is still waiting in l10n review.

## Issue / Bugzilla link

https://github.com/mozilla-l10n/www-firefox-l10n/pull/49#pullrequestreview-2998904412


wen-2018 (approved):
r+

PR #368, opened 2025-07-15
fix nav content horizontal shift (fix #365)

### LINKED ISSUE #365: Nav still moves between pages with and without the shoulder CTA
## Description

Most likely spacer.gif wasn't updated to compensate for download CTA icon addition that made the space wider.

### Steps to reproduce

Navigate between home and some other page that adds the shoulder download button quickly.

### Expected result

Main menu being in pretty much the same spot.

### Actual result

https://github.com/user-attachments/assets/703cd06a-769f-4d01-94c3-613fbf8097e1

### Environment

FXDE, macOS 15

wen-2018:
## One-line summary

This PR fixes the issue that nav content shift when download button is not displayed.

Compare navigation content when download button is displayed or not

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/365

## Testing

http://localhost:8000/en-US/
http://localhost:8000/en-US/browsers/mobile/

### DIFF
--- a/media/css/m24/components/navigation-refresh.scss
+++ b/media/css/m24/components/navigation-refresh.scss
@@ -165,18 +165,6 @@ $margin-top: 58px; // top margin offset for mobile navigation menu
     }
 }
 
-// spacer to center nav items when CTA isn't there
-
-.spacer-gif {
-    display: none;
-
-    @media #{$mq-md} {
-        display: inline-block;
-        width: 120px; // matches width of Firefox logo
-    }
-}
-
-
 // * -------------------------------------------------------------------------- */
 // Menu
 
@@ -527,5 +515,6 @@ body:has(.m24-c-navigation-items.mzp-is-open) {
 
     @media #{$mq-md} {
         display: inline-block;
+        min-width: 156px; // prevent nav content shift when CTA is not displayed
     }
 }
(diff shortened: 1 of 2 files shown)


media/css/m24/components/navigation-refresh.scss
```diff
@@ -527,5 +527,6 @@ body:has(.m24-c-navigation-items.mzp-is-open) {
 
     @media #{$mq-md} {
         display: inline-block;
+        min-width: 156px; // prevent nav content shift when download button is not displayed
     }
```
janbrasna, line 531: Q: how does this overlap with previous work: ? https://github.com/mozmeao/springfield/blob/13446956c9171a58323d980e7979fbcac191284c/media/css/m24/components/navigation-refresh.scss#L168-L176

stephaniehobson: @wen-2018 I think we should update this line ⬆️ to be the 156px

wen-2018: I don't think we need a spacer if we give the CTA container a min-width.

janbrasna (approved):
LGTM.

Works as expected across a few viewport sizes / devices / platforms (and locales).

stephaniehobson (approved):
R+

Issue #381, opened 2025-07-18
Home trio emphasis box doesn't handle longer headings well


janbrasna:
## Description

With new content flowing in from l10n, there might be some places now to double check how bullet-proof the layout is.

### Steps to reproduce

https://www.firefox.com/pl/

### Expected result

Text wrapping hard / not overflowing into other content.

### Actual result

<img width="924" height="501" alt="Image" src="https://github.com/user-attachments/assets/b341be1b-fadc-4658-af8c-92857e40d128" />

### Environment

FXDE, chromium @ macOS

stephaniehobson:
😬 I guess we could use some direction from Tess on how to handle this.... unfortunately she's away.

Let's drop the title size to xs for non-English languages.

stephaniehobson:
I think we could also investigate adding `hyphens: manual;` either here specifically or globally. 

I've found browsers can get a little hyphenation happy if we set it to auto 😬 but this would give the l10n community a chance to add intelligent places to hyphenate words when needed.

stephaniehobson:
I'm also against scaling the text dynamically because it comes with some accessibility concerns and breaks the visual hierarchy if this text acts differently than other things on the page.

I think that brings us back to making the list wrap to the next line if the items are too large to fit on a single line. (And that might be a nice enhancement for users on all screen sizes)

PR #384, opened 2025-07-21
Decrease feature heading font-size for non-EN langs (#381)

stephaniehobson:
## One-line summary

Decrease feature heading font-size for non-EN langs

- Decreased font size for all languages on `lg` screens
- Increased font size to original value on `xl` screens for English only.

## Issue / Bugzilla link

#381  (not totally fixed)

## Testing

(run `make preflight` to update your local translations)

http://localhost:8000/pl/

If that doesn't work just copy/paste the long word from prod ;)

### DIFF
--- a/media/css/firefox/home.scss
+++ b/media/css/firefox/home.scss
@@ -457,6 +457,12 @@ button.mzp-c-cta-link {
     }
 
     @media #{$mq-lg} {
+         h3 {
+            /* decrease heading size */
+            @include font-size($title-sm-size);
+            line-height: $title-sm-line-height;
+        }
+
         li {
             float: left;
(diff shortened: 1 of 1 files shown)


janbrasna:
> `overflow-wrap: break-word;` seems to fix the issue and has good browser support. Not sure if it makes sense to actually break the word in `pl`.

Unfortunately not, grammar/dictionary calls for syllables:

Screenshot 2025-07-22 at 23 43 10

stephaniehobson:
We could look at hyphenating. Let's move the discussion to the issue: https://github.com/mozmeao/springfield/issues/381#issuecomment-3109739518

PR #393, opened 2025-07-28
Run l10n cleanup cron in even months

janbrasna:
## One-line summary

Open obsolete ftl reminders on different schedule to wmo (and refer the correct l10n repo).

Bedrock is run on odd months, so this aims to spread the workload/pings a bit.

- [x] Also: The workflow needs its "L10N" label created, it will be failing until that happens. (See link below for the error.)

## Issue / Bugzilla link

[/actions/runs/16262675495/job/45911310377#step:2:50](https://github.com/mozmeao/springfield/actions/runs/16262675495/job/45911310377#step:2:50)

### DIFF
--- a/.github/workflows/routine_l10n_obsolete.yml
+++ b/.github/workflows/routine_l10n_obsolete.yml
@@ -1,7 +1,7 @@
 name: Bimonthly l10n cleanup reminder
 on:
   schedule:
-    - cron: 0 09 14 */2 *
+    - cron: 0 10 15 */2 *
 
 jobs:
   create_issue:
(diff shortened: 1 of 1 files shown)


.github/workflows/routine_l10n_obsolete.yml
```diff
@@ -1,7 +1,7 @@
 name: Bimonthly l10n cleanup reminder
 on:
   schedule:
-    - cron: 0 09 14 */2 *
+    - cron: 0 10 15 2-12/2 *
```
janbrasna, line 4: Hmh while it's pretty well explained in the workflow editing UI, just reading this doesn't make much sense. Would a comment help?

```suggestion
    - cron: 0 10 15 2-12/2 *  # runs every even month on the 15th day at 10 AM UTC
```

stephaniehobson (commented):
Label created 🏷️

stephaniehobson (approved):
R+

PR #408, opened 2025-08-07
Enable writing webpack files to disk in dev mode

localjo:
This fixes an issue where `npm start` wasn't building/re-building CSS files in dev mode, sometimes causing CSS assets to return 404s or preventing changes from getting written to disk after re-building in watch mode

## One-line summary

Sets `devServer.devMiddleware.writeToDisk = true` in the webpack config, enabling CSS files to be written to disk in dev mode.

This only affects development, and writes files to the disk that were otherwise only stored in memory. I've only tested this locally, not in Docker, but I don't anticipate that it would change anything in Docker.

## Issue / Bugzilla link

None

## Testing

Run `npm start` to start the app, then make a CSS change. The CSS should get re-built and written to disk. Refresh in the browser to confirm that the changes worked.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -105,7 +105,8 @@ module.exports = {
         hot: false,
         static: false,
         devMiddleware: {
-            index: false // specify to enable root proxy'ing
+            index: false, // specify to enable root proxy'ing
+            writeToDisk: true // Write files to disk so Django can serve them
         },
         proxy: [
             {


janbrasna:
https://mozilla.github.io/bedrock/install/#run-a-local-server

>  Regardless of whether you run Bedrock via Docker or directly on your machine, the URL of the site is localhost:**8000** - not **8080**

localjo:
@stevejalim I think the documentation is correct, and the terminal logs also say that the site runs on localhost:8000. The reason I ended up on localhost:8080 was because that's typically the port that Django apps are accessible on, and it was already in my browser history from another project. The 8080 URL serves just the Django side without the webpack dev server for in-memory assets, so I didn't notice I was on the wrong port until I started having issues. So, TL;DR, it was just a mix-up on my end. No worries.

PR #431, opened 2025-08-20
Retain artifacts of Playwright test results in CI

### LINKED ISSUE #429: Playwright CI should upload test results
## Description

Integration testing failures point to test results that are not stored from the runner for retrieval.

Accessibility testing stores test results, but generates warnings there's nothing to store (perhaps successful runs don't emit results?)

### Steps to reproduce

A11y: https://github.com/mozmeao/springfield/actions/runs/17102437709/job/48502440317#annotation:6:19
Integration: https://github.com/mozmeao/springfield/actions/runs/16882868824/job/47822967099#step:5:634

### Expected result

We should probably consistently store artifacts for inspection, and ignore when there are none? https://github.com/actions/upload-artifact#customization-if-no-files-are-found

### Actual result

A11y: "No files were found with the provided path: tests/playwright/test-results-a11y/. No artifacts will be uploaded." on successful run.

Integration: "Error Context: test-results/products-firefox-firefox-f-082cc--page-Feature-page-private--firefox/error-context.md" generated on CI failure, but no artifact stored.

### Environment

CI:
https://github.com/mozmeao/springfield/actions/workflows/a11y_tests.yml
https://github.com/mozmeao/springfield/actions/workflows/integration_tests.yml

janbrasna:
## One-line summary

Uploads failures. Silences missing results on success.

Using distinct IDs so there should be no issues in conflicting names from different ref runs.

## Issue / Bugzilla link

Resolves #429

## Testing

💭

### DIFF
--- a/.github/workflows/a11y_tests.yml
+++ b/.github/workflows/a11y_tests.yml
@@ -37,9 +37,6 @@ jobs:
     env:
       PLAYWRIGHT_BASE_URL: ${{ github.event.inputs.springfield_service_hostname || 'https://dev.springfield.nonprod.webservices.mozgcp.net' }}
       CI: true
-      CI_JOB_ID: ${{ github.run_id }}
-      DRIVER: ""
-      LABEL: accessibility-tests
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-node@v4


.github/workflows/integration_tests.yml
```diff
@@ -111,6 +108,13 @@ jobs:
       run: cd tests/playwright && npm ci && npm run install-deps
     - name: Run Playwright tests
       run: cd tests/playwright && npm run integration-tests
+    - uses: actions/upload-artifact@v4
+      if: ${{ !cancelled() }}
+      with:
+        name: test-results-${{ github.event.inputs.git_sha }}-${{ github.run_id }}
```
janbrasna, line 114: Or if it gets too long, it can be shortened to just:

```suggestion
        name: test-results-${{ github.run_id }}
```

(but I left the triggering sha from cron in place too, if folks want to cross-reference the versions the results came from later… but it's the run_id making the artifact unique really.)

.github/workflows/integration_tests.yml
```diff
@@ -99,9 +99,6 @@ jobs:
     env:
       PLAYWRIGHT_BASE_URL: ${{ github.event.inputs.springfield_service_hostname }}
       CI: true
-      CI_JOB_ID: ${{ github.run_id }}
-      DRIVER: ""
-      LABEL: playwright-tests
```
janbrasna, line 104: FYI these are being removed because they are not used by the CI, or Playwright here. It's leftovers from functional tests, mainly to configure Selenium, pytest and codecov, which is not relevant for these jobs.

.github/workflows/a11y_tests.yml
```diff
@@ -37,9 +37,6 @@ jobs:
     env:
       PLAYWRIGHT_BASE_URL: ${{ github.event.inputs.springfield_service_hostname || 'https://dev.springfield.nonprod.webservices.mozgcp.net' }}
       CI: true
```
janbrasna, line 39: FYI the runner sets [`CI=true`](https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables) itself by default, so it's superfluous here.

I haven't removed it as it might be used in Playwright configs to make decisions, so it's a little more explicit that such value is available in `process.env.CI` conditionals for visibility.

Issue #323, opened 2025-06-24
[Sentry Error] Event CustomEvent (type=unhandledrejection) captured as promise rejection


stephaniehobson:
`Event 'CustomEvent' (type=unhandledrejection) captured as promise rejection`

**Location:**
Happening on pages which trigger downloads in all languages .

**Browser:**
No reports of this error from Firefox but Chrome, Edge, and Safari trigger it.

**OS:**
Reported in both Windows and Mac.

[Full Sentry report](https://mozilla.sentry.io/issues/6691214032/?project=4508966604242944&query=is%3Aunresolved&referrer=issue-stream&sort=freq&stream_index=0). (Mozilla only)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-112)
┆Priority: P3
┆Sprint: Backlog

stephaniehobson:
I had a look around our JS and could not find any place where we don't handle promise rejections.

janbrasna:
The updates above improved UITour checks for Firefox — however the OP mentioned:

> "No reports of this error from Firefox but Chrome, Edge, and Safari trigger it."

Have the reports improved, or `CustomEvent` rejections are still being logged, esp. for other browsers?

stephaniehobson:
It's still a problem 🙁

janbrasna:
Hmh there might be more waiting just around the corner:

<img width="1156" height="526" alt="Image" src="https://github.com/user-attachments/assets/c2fbdce9-85a1-4de3-a7fc-58df6ed3928d" />

_(I have similar type of endpoints blocked on the DNS level, hence the loading error.)_

I have not followed the TypeError further, but the unhandled rejection seems to be in their code. 🤷

PR #436, opened 2025-08-26
Wrap dataLayer pushes in try/catch #323

stephaniehobson:
## One-line summary

Wrap `TrackProductDownload` `dataLayer` pushes in try/catch #323

- This is an attempt to address the (so far) unreproducible error in #323.
- Wraps both dataLayer calls in a try/catch for promise rejection

## Issue / Bugzilla link

#323 

## Testing

Make sure the events still fire as expected.

### DIFF
--- a/media/js/base/datalayer-productdownload.es6.js
+++ b/media/js/base/datalayer-productdownload.es6.js
@@ -245,9 +245,20 @@ TrackProductDownload.sendEventFromURL = (downloadURL) => {
  * @param {Object} - product details formatted into a product_download event
  */
 TrackProductDownload.sendEvent = (eventObject) => {
-    window.dataLayer.push(eventObject);
-    // we also want to keep the old event name around for a few months to help with the transition
-    // this can be deleted as part of the UA cleanup
+    // Treating this as the source of mozmeao/springfield#323 and wrapping it in a try/catch
+    try {
+        const result = window.dataLayer.push(eventObject);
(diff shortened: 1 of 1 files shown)


localjo:
It might make things a little bit DRYer and more readable to put the try/catch block into it's own function instead of repeating it. Something like this:

```js
function safePushEventToDataLayer(eventObject) {
    // Treating this as the source of mozmeao/springfield#323 and wrapping it in a try/catch
    try {
        const result = window.dataLayer.push(eventObject);
        // Handle if dataLayer.push returns a promise (some GTM configurations do this)
        if (result && typeof result.then === 'function') {
            result.catch(() => {
                // Silently handle promise rejections to prevent unhandled rejection errors
            });
        }
    } catch (error) {
        // Handle synchronous errors
    }
}

TrackProductDownload.sendEvent = (eventObject) => {
    safePushEventToDataLayer(eventObject);
    // we wanted to keep the old event name around for a few months to help with the transition
    // now there are a bunch of dashboards built to use it so it gets to live on forever
    TrackProductDownload.sendOldEvent(eventObject);
};

TrackProductDownload.sendOldEvent = (eventObject) => {
    const oldEventObject = JSON.parse(JSON.stringify(eventObject));
    // replace event name with old event name
    oldEventObject['event'] = 'product_download';
    safePushEventToDataLayer(oldEventObject);
};
```

Looks good to me, r+

PR #441, opened 2025-08-28
Add basic WNP logic and Nightly WNP

maureenlholland:
## Testing

Nightly: http://localhost:8000/en-US/whatsnew/140.0a1 
- check the Firefox Labs UITour link correctly goes to about:preferences#experimental (https://mozmeao.github.io/platform-docs/uitour/#local-development)


maureenlholland:
TODO:
- finish default evergreen (add variants to view logic)
- add tests
- general QA

springfield/firefox/views.py
```diff
@@ -821,3 +822,91 @@ class PlatformViewWindows(L10nTemplateView):
...
+            if version.endswith("a1"):
+                return "nightly"
+            if version.endswith("a2"):
+                return "developer"
+
+    return "unknown"
+
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1399276
+# https://github.com/mozilla/bedrock/pull/5132
+def show_57_dev_whatsnew(version):
```
maureenlholland, line 841: check if we still need this

l10n/en/firefox/whatsnew/send-to-device.ftl
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.springfield.moz.works/whatsnew/60.0/
+
+whatsnew-s2d-download-firefox-for-android = Download { -brand-name-firefox } for { -brand-name-android } and { -brand-name-ios }
```
maureenlholland, line 7: check if Android and iOS should be hardcoded now

springfield/firefox/views.py
```diff
@@ -821,3 +822,111 @@ class PlatformViewWindows(L10nTemplateView):
...
+
+        return ctx
+
+    def get_template_names(self):
+        version = self.kwargs.get("version") or ""
+
+        oldversion = self.request.GET.get("oldversion", "")
+        # old versions of Firefox sent a prefixed version
+        if oldversion.startswith("rv:"):
+            oldversion = oldversion[3:]
```
janbrasna, line 909: I don't recall the `oldversion` is actually being used for anything. (But it would be cool if that's tracked somehow. _Which probably is, on product level or on the actual URL params hitting that page…_)

media/img/firefox/whatsnew/qrcode-firefox-whatsnew-variant-b.png
janbrasna, line 1: When you're in the mood to get rid of these, they could probably be replaced with the qrcode() helper — to actually gives us some more visibility into where they really point to, and with what params — and to be able to tweak them if needed.

janbrasna (commented):
Oooh, this in nice!

In case you're actually not on a Nightly, that about:preferences#experimental will lead you just to #general (as you won't have the Labs feature available, so don't be surprised if you're testing from a different channel;)…)

l10n/en/firefox/whatsnew/evergreen.ftl
stevejalim, line 1: @maureenlholland Do we need to translate these via Pontoon? If so, we probably need to add to `l10n/configs/pontoon.toml` I think...

(I could be wrong, so push back if I am!)

stevejalim (approved):
Backend r+wc - nice work!

l10n/en/firefox/whatsnew/evergreen.ftl
maureenlholland: ah, good catch! yes, these should be Pontoon translations and config needs update

PR #459, opened 2025-09-08
16488 fix linting

localjo:
## One-line summary

Fixes linter errors from the pre-commit hook.

- Auto-fixed lots of formatting
- Excluded Wagtail CMS CSS from linting
- Re-generated migrations file


springfield/cms/migrations/0006_whatsnewpage.py
```diff
@@ -1,27 +1,173 @@
...
+import wagtail.fields
+
 
 class Migration(migrations.Migration):
-
     dependencies = [
-        ('cms', '0005_springfieldimage_description'),
-        ('wagtailcore', '0094_alter_page_locale'),
+        ("cms", "0005_springfieldimage_description"),
+        ("wagtailcore", "0095_groupsitepermission"),
```
stephaniehobson, line 16: This update removes `('wagtailcore', '0094_alter_page_locale')` as a dependency. Is that intentional?

stephaniehobson (commented):
R+ ?

Question about the migration dependencies but otherwise good to merge.

springfield/cms/migrations/0006_whatsnewpage.py
SmileyChris: It doesn't remove the previous migration, it's just updating this migration's dependency to the newer wagtailcore dependency (i.e. wagatil 0095 will get run before our cms 0006)

PR #481, opened 2025-09-16
Port bedrock work from July 1 to current

maureenlholland:
## One-line summary
Cherry picks work from WMO that is relevant to FXC

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/issues/452

## Testing
* [x] 404 page: http://localhost:8000/en-US/404/
* [x] dev edition: http://localhost:8000/en-US/channel/desktop/developer/
* [x] testflight (email autocomplete): http://localhost:8000/en-US/channel/ios/testflight/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-153)
┆Priority: P3
┆Sprint: Backlog

### DIFF
--- a/media/css/base/page-not-found.scss
+++ b/media/css/base/page-not-found.scss
@@ -12,12 +12,6 @@ $image-path: '/media/protocol/img';
         padding-top: $layout-sm;
         padding-bottom: $layout-sm;
     }
-
-    .header-image {
-        .logo-mozilla {
-            display: inline-block;
-        }
-    }
 }
 
 .message-section {


l10n/en/send_to_device.ftl
```diff
@@ -7,6 +7,8 @@ send-to-device-your-download-link = Your download link was sent.
 send-to-device-please-enter-an-email = Please enter an email address.
 send-to-device-an-error-occured = An error occurred in our system. Please try again later.
 send-to-device-enter-your-email = Enter your email
+# Only localize "yourname". Do not change "@example.com".
+send-to-device-email-placeholder = yourname@example.com
```
janbrasna, line 11: _(I did not pay too much attention to this on the pages slated for removal before, but will be checking here now…)_

I think we have several ways l10ns adjust the placeholders — some countries have national equivalents of the example domains, some use "bogus" email formats etc. — I think we generally leave these to the localizers as they know what to do with them. Adding instructions here might make it seem that particular format is important for functionality… which it is not.

(I'll look again how some placeholders end up translated in the consumed locales.)

PR #483, opened 2025-09-16
Remove inactive variation JS in mobile/ios template

### LINKED ISSUE #443: Remove inactive variation JS in `mobile/ios`
## Description

https://github.com/mozmeao/springfield/blob/45eda4c0edd72ff6e9b0eaaac74fd83216c02af0/springfield/firefox/templates/firefox/browsers/mobile/ios.html#L169-L173

The ?v is not exposed so it's not prone to crashing, but the bundle with the ID `firefox-browsers-ios-sms` does not exist here.

(There might be more leftovers from that presumably send-to-device experiment elsewhere?)

---

#### Success Criteria
- [ ] Nonexistent bundle reference removed
- [ ] Variant conditions for deleted experiments are removed

adi-ray:
## One-line summary
Removes leftover code that referenced a non-existent JavaScript bundle.

## Issue / Bugzilla link
Fixes #443 

## Testing
- [x] Nonexistent bundle reference removed
- [x] Variant conditions for deleted experiments are removed

### DIFF
--- a/springfield/firefox/templates/firefox/browsers/mobile/ios.html
+++ b/springfield/firefox/templates/firefox/browsers/mobile/ios.html
@@ -166,11 +166,7 @@
 {% endblock %}
 
 {% block js %}
-{% if variation == '2' %}
-  {{ js_bundle('firefox-browsers-ios-sms') }}
-{% else %}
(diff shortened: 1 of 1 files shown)


stephaniehobson (approved):
R+

TY again!

PR #609, opened 2025-09-26
CMS card blocks

maribedran:
## One-line summary
Create all card patterns as blocks on the CMS.

Organize CSS files and create the `fl-` prefix for CSS classes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-219)
┆Priority: P2
┆Sprint: Backlog


l10n/en/cms.ftl
```diff
@@ -0,0 +1 @@
+block-step-card-step = Step
```
maribedran, line 1: @stevejalim I'm not sure if this is the best place for the FTL strings for the CMS blocks, but my thinking is that we'll only have a handful of those since almost none of the strings in the CMS will be hardcoded. I thought of prefixing them with `block-BLOCK_NAME-TOKEN`. What do you think?

PR #617, opened 2025-10-01
Move cron CI to "latest" image tag and store results on failure

janbrasna:
## One-line summary
Avoids failed runs when HEAD ref SHA is still building and unavailable from the registry.

Porting https://github.com/mozilla/bedrock/pull/16569

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/16568
https://github.com/mozilla/bedrock/issues/16498

## Testing

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-248)
┆Priority: P2
┆Sprint: Backlog

### DIFF
--- a/.github/workflows/cdn_tests.yml
+++ b/.github/workflows/cdn_tests.yml
@@ -1,4 +1,4 @@
-# Workflow that runs download-link tests on live infra
+# Workflow that runs CDN and SSL tests on live infra
 
 name: CDN tests
 run-name: CDN tests for ${{ github.sha }}
@@ -23,7 +23,7 @@ jobs:
     if: github.repository == 'mozmeao/springfield'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Notify via Slack that tests are starting
         uses: ./.github/actions/slack
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cdn_tests.yml
```diff
@@ -1,4 +1,4 @@
-# Workflow that runs download-link tests on live infra
+# Workflow that runs CDN and SSL tests on live infra
 
 name: CDN tests
 run-name: CDN tests for ${{ github.sha }}
```
janbrasna, line 4: FYI I was also contemplating what to do with the naming / run label, as the result is definitely not "for github.sha" — that would be the main HEAD — which may or may not be used as the source of the test code to hit the deployments, but is most likely not the ref deployed in prod or any of the hosts this is run against. (The github.sha passed as a trigger to cron events is the default branch head, even when that's not something you then ever use in the CI — in our case the only thing that ref influences is the slack notifier code version;) and the initial shell script to pull containers and kick it all off…) — But any git-rev-parse–fu to try to parse the ref from prod branch, or curl revision.txt for the deployment hit can only be run way after the workflow has started, and can't be really renamed on the top level. (There's an ugly low level API that could be patched with run name even during the workflow execution, but it's just like super fugly ugly to call it, authenticated of course, for that correct run id, and be tweaking it from inside the run while it's running…)

So for now I've left it as is, but the sha there is misleading regarding the content tested. Wondering whether to not put the manual trigger there, and if not, just some defaults — like `for ${{ github.event.inputs.springfield_service_hostname || 'www.firefox.com' }}` maybe with "on [day of week]" added, but quickly checking I don't see a context with any meaningful info to be available at this top level before any job is run, so there's not much to be useful there… or maybe incorporate `github.run_attempt` if greater than 1 for visibility etc. somehow? Ran out of ideas.

stevejalim (approved):

.github/workflows/cdn_tests.yml
janbrasna: After a little time away from this I realized we actually want to save the deployed revision.txt for record keeping, in case something doesn't look alright from the test results — so I'm adding it somewhere to my bucket list — while unable to surface it to the run-name, to add a logging step somewhere inside the job to just scrape the deployment's ref and keep it in the run log for us.

PR #649, opened 2025-10-09
[WT-308] CMS - Add Ability To Add UI Tour Buttons In Wagtail

dchukhin:
## One-line summary
Wagtail users can now add UI Tour Buttons wherever they have a choice to add (regular) buttons.

 - A new `UITourButtonBlock` (almost the same as the `ButtonBlock`, except link URL is replaced with a button type) is added to all of the places that user `ButtonBlock` for buttons.
- templates are updated to show UI Tour buttons directly after (regular) buttons
- a new JavaScript file (`media/js/cms/ui-tour-buttons.js`) handles directing the user to the appropriate place based on the button

Note for local testing:
the UI tour functionality only works on mozilla domains, so to make it work locally (on http://localhost:8000/), you will need to:
1. open Firefox and go to about:config
2. change browser.uitour.requireSecure to false
3. change browser.uitour.testingOrigins to http://localhost:8000/
4. maybe restart Firefox

### DIFF
--- a/springfield/cms/templates/components/uitour_button.html
+++ b/springfield/cms/templates/components/uitour_button.html
@@ -5,7 +5,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #}
 
 <div class="ui-tour">
-  <button class="button {{ theme_class }}" {% if disabled %}disabled{% endif %}>
+  <button class="button {{ theme_class }}" type="button" {% if disabled %}disabled{% endif %}>
     {% if icon_name and icon_position == "left" %}
       {% set extra_class = "icon-left" %}
       <include:icon


maribedran (dismissed):

springfield/cms/templates/components/uitour_button.html
```diff
@@ -0,0 +1,27 @@
+{#
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+<div class="ui-tour">
+  <button class="button {{ theme_class }}" {% if disabled %}disabled{% endif %}>
```
stephaniehobson: Let's add a `type="button"` on here just to be safe.
```suggestion
  <button class="button {{ theme_class }}" type="button" {% if disabled %}disabled{% endif %}>
```

PR #659, opened 2025-10-14
Update app-dev newsletters to kwargs

### LINKED ISSUE #658: Replace `email_newsletter_form` positional arguments with kwargs
The `email_newsletter_form` template tag helper calls for 'app-dev' (Mozilla Developer) newsletter id leaning on the `newsletters` value as the implicit first argument. This can break if the order of things shifts.

janbrasna:
## One-line summary

Swaps positional arguments for keywords to make reordering less fragile.

This also makes it more obvious what the value signifies, given it has to be passed to Basket and doesn't occur anywhere else on the site.

## Issue / Bugzilla link

Fixes #658

## Testing

/en-US/channel/desktop/developer/

_(NB: I wasn't able to trace back the alt-newsletter include use years back; it seems it was extracted out of the page once, but never actually used…)_

### DIFF
--- a/springfield/firefox/templates/firefox/developer/includes/alt-newsletter.html
+++ b/springfield/firefox/templates/firefox/developer/includes/alt-newsletter.html
@@ -17,7 +17,7 @@
 ) %}
   {% if LANG.startswith('en-') %}
   {{ email_newsletter_form(
-    'app-dev',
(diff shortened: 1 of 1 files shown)


stephaniehobson (approved):
R+ 🍎

PR #660, opened 2025-10-14
Remove "global" from global root title

janbrasna:
## One-line summary

Simplifies root nonlocale page title for robots & link previews.

The root title has been already disambiguated from other home pages so this extra bit is perhaps not necessary.

## Issue / Bugzilla link

#453

## Testing

http://localhost:8000/en-US/404-locale/
http://localhost:8000/ (in prod mode, and with Accept-Language empty)


stephaniehobson (approved):
R+ 🌐

PR #663, opened 2025-10-14
[WT-340] CMS revisions

maribedran:
## One-line summary

A few revisions on CMS blocks.

- Add the "Your Firefox has been updated." notification at the top of the What's New Page
- Add Page chooser to the button and link blocks as an option to the link text input
-

### DIFF
--- a/media/css/cms/flare-banner.css
+++ b/media/css/cms/flare-banner.css
@@ -1,3 +1,9 @@
+/*
+* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
 .fl-banner {
     position: relative;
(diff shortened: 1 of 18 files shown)


springfield/cms/models/pages.py
```diff
@@ -168,6 +168,8 @@ class WhatsNewPage(AbstractSpringfieldCMSPage):
     parent_page_types = ["cms.WhatsNewIndexPage"]
     subpage_types = []
 
+    ftl_files = ["firefox/whatsnew/evergreen"]
```
stevejalim: There may be a risk here that we're speccing a single FTL file, but in doing so, lose the use of the ones in settings.DEFAULT_FLUENT_FILES. It might be that the WNP doesn't use any of the default strings (eg in the footer), but worth a check in case it does

maribedran:
@stephaniehobson Added validation to prevent the external option from being selected when a page is chosen. We can add further validation for the domains in a future PR if we define the rules for that.

PR #671, opened 2025-10-17
notification refactor to fix theme combinations

kkellydesign:
## One-line summary
This fixes the themes for the notifications when icons are in use. This refactors the components to rely more heavily on CSS variables, in order to see all of the variations in one place in the code. 

image

### DIFF
--- a/media/css/cms/flare-notification.css
+++ b/media/css/cms/flare-notification.css
@@ -5,8 +5,6 @@
 */
 
 
-/* Negative margin used to reduce the top spacing on the next block */
-
 :root {
     --notification-bg: light-dark(var(--blue-light-blue), var(--blue-light-blue));
     --notification-color: var(--neutrals-charcoal);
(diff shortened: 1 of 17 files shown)


media/css/cms/flare-notification.css
```diff
@@ -1,3 +1,17 @@
+/*
+* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
+
+/* Negative margin used to reduce the top spacing on the next block */
```
stephaniehobson: Can this comment be moved to be next to the negative margin declaration it references?

media/css/cms/flare-notification.css
```diff
@@ -125,57 +142,11 @@
...
-
-.notification-black {
-    background: var(--neutrals-charcoal);
-}
-
-.notification-white {
-    background: light-dark(var(--neutrals-white), var(--neutrals-ash));
-}
-
 .notification-outline {
```
stephaniehobson: The tests are complaining because you defined this twice. Possible fixes: move the border declaration up to line 91. Add a comment ignoring this rule.

I'm not usually a fan of ignoring rules but sometimes it makes sense to because it make the file more human readable but I think in this case grouping them together on line 91 makes sense.

PR #677, opened 2025-10-20
Add UTM parameters to CMS block links

maribedran:
## One-line summary

Add UTM query parameters to all CMS external links to Mozilla domains.

Free form pages can now define a context variable to build UTM query parameters.
A new template tag was added to add the UTM parameters to urls when the variable is set to the request context.

## Testing

See the test cases in `springfield/cms/tests/test_templatetags.py` and add links following those patterns to different buttons and CTAs in a What's New Page and check if the resulting URL has the correct parameters.

### DIFF
--- a/springfield/cms/templates/cms/free_form_page.html
+++ b/springfield/cms/templates/cms/free_form_page.html
@@ -10,10 +10,8 @@
 {% block content %}
 <div class="fl-page">
   <main class="fl-main">
-    <include:inline-notification
-      message="{{ ftl('whatsnew-update-notification') }}"
-      color="purple"
-    />
+    {% block pre_content %}
+    {% endblock %}
 
     {% set ns = namespace(headings=0) %}
     {% for block in page.content %}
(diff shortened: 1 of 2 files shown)


stephaniehobson (approved):
Right UTMs are being added to the right links 👍 

I ran into an error using a Section CTA outside of the whats new page template.

R + if you can confirm that error is unrelated to this PR.

springfield/cms/templates/cms/free_form_page.html
```diff
@@ -10,12 +10,22 @@
 {% block content %}
 <div class="fl-page">
   <main class="fl-main">
+    <include:inline-notification
```
stephaniehobson: Non-what's new pages should not display the update notification ;)

maribedran: Thanks @stephaniehobson ! I've been looking so much at the page with the notification that I didn't even noticed it there.

PR #691, opened 2025-10-23
Add new PDF features to the PDF feature page

marco-c:
Fixes part of https://github.com/mozilla/bedrock/issues/15977

This is at least adding the missing features to the page, but I think we should definitely refresh this page.


l10n/en/firefox/features/pdf-editor-2023.ftl
```diff
@@ -18,9 +18,13 @@ pdf-editor-weve-all-faced-this-you-need-to = We’ve all faced this: you need to
 pdf-editor-no-more-now-all-you-need-to-do = No more! Now, all you need to do is edit the PDF online with { -brand-name-firefox }, save it, and email it from your computer.
 pdf-editor-add-text = Add text
 pdf-editor-open-the-pdf-in-firefox-click-text = Open the PDF in { -brand-name-firefox }. Click the Text button to choose a color and text size before selecting where on the document you wish to add text. It’s that easy!
-pdf-editor-add-drawings-or-your-signature = Add drawings (or your signature)
-pdf-editor-open-the-pdf-in-firefox-click-draw = Open the PDF in { -brand-name-firefox }. Click the Draw icon to choose a color, thickness and opacity before then being able to draw on the document. It probably won’t be any messier than your usual signature!
+pdf-editor-add-drawings = Add drawings
+pdf-editor-open-the-pdf-in-firefox-click-draw2 = Open the PDF in { -brand-name-firefox }. Click the Draw icon to choose a color, thickness and opacity before then being able to draw on the document.
```
maureenlholland:
```suggestion
pdf-editor-open-the-pdf-in-firefox-click-draw-to-choose = Open the PDF in { -brand-name-firefox }. Click the Draw icon to choose a color, thickness and opacity before then being able to draw on the document.
```

**blocking**: we either version strings to provide a fallback or remove strings and provide new non-versioned strings https://mozmeao.github.io/platform-docs/l10n/?h=fluent#obsolete-strings 

In this case, it looks like we want to remove, so let's provide a non-versioned ID

**blocking**: we either version strings to provide a fallback or remove strings and provide new non-versioned strings https://mozmeao.github.io/platform-docs/l10n/?h=fluent#obsolete-strings 

In this case, it looks like we want to remove, so let's

maureenlholland (approved):
r+ thanks again

PR #703, opened 2025-10-27
Prevent Wagtail from creating migrations that don't change the DB

maribedran:
## One-line summary

Don't create migrations when CMS blocks have changes that don't affect the database.

Create a new `StreamField` class that ignores block definitions when used in Django's check to detect changes in models.
Definition took from: https://github.com/wagtail/wagtail/issues/4298#issuecomment-1159515247 

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-378

## Testing

Make changes locally to any CMS block and run `./manage.py makemigrations` to check that no new migrations are created.


springfield/cms/migrations/0018_alter_freeformpage_content_and_more.py
```diff
@@ -0,0 +1,28 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name="freeformpage",
+            name="content",
+            field=springfield.cms.fields.StreamField(block_lookup={}),
+        ),
+        migrations.AlterField(
+            model_name="whatsnewpage",
+            name="content",
+            field=springfield.cms.fields.StreamField(block_lookup={}),
```
stevejalim: This all looks really promising, thanks! 

Just to check I'm on the same page: these two models have new migrations because this is the first time we're skipping the StreamField's content, yep?

maribedran:
@stevejalim, the custom migration operations have to be manually created anyway. We generate an empty migration and have to write the code that specifies the changes. I haven't thought of a good way to validate this except for code reviews. Anything that changes the schema of the page's data needs one of those, but that's something the developer needs to know.

PR #716, opened 2025-10-30
[WT-387] Firefox Account link in rich text editor

maribedran:
## One-line summary

Add a new feature to Wagtail's rich text editor to add Firefox Account links.

The feature is implemented based on Wagtail's doc example at https://docs.wagtail.org/en/stable/extending/extending_draftail.html#creating-new-entities:
- Register the new feature at `springfield/cms/wagtail_hooks.py` with the appropriate handlers to convert data from and to the database
- Implement the JS to power the editor UI at `media/js/cms/wagtailadmin-fxa.js`
- The feature uses a custom `<fxa>` HTML tag that is parsed by a new `richtext` template filter at `springfield/cms/templatetags/cms_tags.py`
- The filter replaces the custom tag with the link HTML with all the necessary parameters built from the request context

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-387

## Testing

[1min Demo video](https://www.loom.com/share/3d3134b42386490ca32bc6a86648b26f)

- Edit any rich text field in a What's New or Free Form page and add an FXA link to a text selection
- Check that the resulting link takes you to FXA and has all the appropriate analytics attributes and UTM parameters
image
- Translate the page and check that the translated link has the same `data-cta-uid`

### DIFF
--- a/springfield/cms/templatetags/cms_tags.py
+++ b/springfield/cms/templatetags/cms_tags.py
@@ -154,8 +154,7 @@ def richtext(context, value: str) -> str:
     See springfield/cms/wagtail_hooks.py for the <fxa> tag registration.
     """
     rich_text = wagtail_richtext(value)
-    text = rich_text.source if isinstance(rich_text, RichText) else str(rich_text)
-    soup = BeautifulSoup(text, "html.parser")
+    soup = BeautifulSoup(str(rich_text), "html.parser")
 
     for fxa_tag in soup.find_all("fxa"):
         label = fxa_tag.text


kkellydesign (changes requested):
We need to be able to remove links as well, if possible. If we have to just backspace through it, that's ok, but it'd be great if we could figure out this bit of UI consistency. 

https://www.loom.com/share/fce6bc44a4464b07ada227d36f073dbc

kkellydesign (approved):
All of these revisions look good to me.

stephaniehobson (changes requested):
`[role=button]` hover styles are complicating the hover display of these links.

https://github.com/user-attachments/assets/319d0c30-824a-44cd-945f-e1c5a00f5bb0

PR #717, opened 2025-10-31
Add light-dark fallbacks for old Safari and split token/theme variables

localjo:
## One-line summary

Adds fallbacks for light-dark rules that were ignored by Safari 17.4 and earlier, and splits CSS variables into separate token and theme stylesheet with prefixes.

Unfortunately, Safari 17.4 and lower recognize the light-dark syntax as valid, and parse the rules, but are unable to resolve the values. Because they recognize the syntax as valid, a simple fallback rule before the light-dark rule doesn't work, so this `@supports` syntax was the most straightforward way I could find to add fallbacks.

This PR also splits CSS variables into tokens/constants in flare-tokens.css with `--token-` prefixes and variables that change depending on context (viewport media queries, dark mode, etc) in flare-theme.css with `--fl-theme-` prefixes.

## Testing

Verify that light mode styles work correctly in Safari 17.4 and earlier.

### DIFF
--- a/media/css/cms/flare-newsletterform.css
+++ b/media/css/cms/flare-newsletterform.css
@@ -132,9 +132,15 @@
     border-color: var(--token-neutrals-charcoal);
 }
 
-.fl-section-subscribe .fl-button:hover {
+.fl-section-subscribe .fl-button:hover,
+.fl-section-subscribe .fl-button:focus {
     background: var(--token-neutrals-black);
     color: var(--token-neutrals-white);
+}
+
+.fl-section-subscribe .fl-button:active {
+    background: var(--fl-theme-button-primary-active-bg);
+    color: var(--fl-theme-button-primary-active-text);
 }
 
 .fl-section-subscribe .fl-newsletterform-error-list {


stephaniehobson:
We're going to have this problem with every `light-dark()` usage on the site, eh? What are your thoughts on re-factoring to support dark mode in a different way?

localjo:
I suggest that we use the prefix `--token-` for custom CSS properties that don't change, and `--theme-` for the ones that change depending on context. That naming seems to align with conventions that Figma and Adobe are using. We could do something like this:

```css
:root {
  /* TOKENS - Static constants */
  --token-color-white: #fff;
  --token-color-black: #15141a;
  --token-color-purple: #7543e3;
  --token-font-headline: 'Mozilla Headline', sans-serif;
  --token-scale-16: 16;
  --token-scale-32: 32;
  --token-width-mobile: 375;
  --token-width-tablet: 768;
  --token-width-desktop: 1024;
  --token-width-max: 1440;

  /* THEME - Start with mobile light mode defaults */
  --theme-bg-color: var(--token-color-white);
  --theme-text-color: var(--token-color-black);
  --theme-spacing: calc(var(--token-scale-16) * 1px);
  --theme-font-size-h1: 32px;
  --theme-container-width: calc(var(--token-width-mobile) * 1px);
}

/* Set variables for dark mode */
@media (prefers-color-scheme: dark) {
  :root {
    --theme-bg-color: var(--token-color-black);
    --theme-text-color: var(--token-color-white);
  }
}

/* Set variables for desktop */
@media (min-width: 1024px) {
  :root {
    --theme-spacing: calc(var(--token-scale-32) * 1px);
    --theme-font-size-h1: 48px;
    --theme-container-width: calc(var(--token-width-desktop) * 1px);
  }
}
```

@stephaniehobson So far, to avoid being verbose, we haven't been prefixing CSS custom properties with a flare-related prefix like we are for classnames, but if you think it's useful, we could also use `--fl-token-` and `--fl-theme-` as the prefixes.

stephaniehobson:
@localjo Agree on `token` and `theme` prefixes. Let's leave the tokens plain but add `fl` to the theme ones so `--token-` and `--fl-theme-`

kkellydesign:
> I should clarify that the tokens aren't necessarily 1:1 with the Figma tokens

That's totally fine. We can backport our conventions back toward Figma if necessary.

kkellydesign (commented):
EDIT: I meant to make this an inline comment. Pasting there now so it references real code. Sorry for the noise. 

@stephaniehobson a couple of things here: 
* We should err on the side of getting this merged quickly, as merge conflicts will be painful if this sits too long. However, this is a good PR to rip off some bandaids and  discuss some structural opinions. 
* I'm not crazy about this method for styling, but we should agree on a recommendation before I ask for a change. So these are from Figma. My problem with using them like this is that they're not meaninful numbers without units. 40px ≠ 40em ≠ 40rem. My suggestion is to use rem units globally in these variables, and use ems when it makes sense, but to keep them out of the variables. (for example, I can imagine a default em value for `<small>`. We can use px values in the variable names if that makes sense to you. (I know personally, I still think in pixels, and have to sit with a calculator open when working in rems.) So we'd have this: 

```css
:root {
  --scale-14: 0.875rem;
  --scale-16: 1rem;
}
.fl-banner {
  padding-block: var(--scale-16);
}
```

media/css/cms/flare-theme.css
```diff
@@ -5,178 +5,180 @@
...
-This file was generated using the Figma MCP tool to export all variables
-from the Figma design file. The MCP tool provides comprehensive access to
-all design tokens that aren't easily visible in Figma's standard UI or
-other Figma plugins (unfortunately)
-
-To update this file in the future, use the MCP tool with a prompt that
-asks the agent use mcp_Figma_get_variable_defs to get all variables
-from the selected Figma file and update this file. The process is
-non-determinative so you need to check and adjust the results as needed.
+Theme - Contextual CSS custom properties that can change based on
```
maureenlholland: 🎉 love this

maureenlholland (approved):
r+ if the newsletter button focus state is not a regression

PR #732, opened 2025-11-04
Include Wagtail Docs URLs in supported nonlocales

maribedran:
## One-line summary

Don't add a locale prefix to Wagtail document URLs.


stevejalim (approved):
LGTM!

PR #741, opened 2025-11-06
Add full width image/video support to CMS blog page layout

localjo:
## One-line summary

Add full width image/video support to CMS blog page layout

This uses the existing video.html template (which doesn't appear to be used by any Wagtail block yet) and adds an aspect_ratio parameter which is important for YouTube embeds since they don't have an intrinsic height ratio. The default is 16 / 9 which should cover 95% of YouTube videos, but we can add CMS support to set other aspect ratios if needed.


media/css/cms/flare-media.css
```diff
@@ -0,0 +1,87 @@
...
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
+.media img,
+.media video,
+.media iframe {
+    display: block;
+    width: 100%;
+    height: auto;
```
localjo: @maribedran Re: the missing border radius, I think the important thing is consistency within the intro component. Right now, the media (images, video) don't have a border radius, but the images looked like they did because the placeholder image has the rounded borders inside the image file. Assuming that both images and videos in the intro component should have a border radius, adding it here should fix both.

```suggestion
    height: auto;
    border-radius: calc(var(--scale-16) * 1px);
```

Issue #697, opened 2025-10-27
Support post-deployment check for assets


stevejalim:
## Description

Whenever we do a new release, we need to be 100% sure that all of our static assets are available. If a CSS or JS file is missing, this can break a page, making it harder or even impossible for users to find or download Firefox. 

We had an incident recently where this was not the case and while the root cause has been identified and remedied, there is still value in checking that the new assets are available.

As part of our post-deployment checks, we should check that all the assets are available on 
a) the origin server(s) for the relevant environment
b) the CDN for that environment

It's worth remembering that if an asset is missing, the CDN will cache that 404 for many, many years. So, if we detect this situation, the error messaging needs to very clearly show which paths are 404ing on the CDN so that they can be manually invalidated ASAP. (Stretch goal: Ideally, we would automatically invalidate that path on the CDN and re-run the checks until we end up with 200 OKs for every static assets)

---

#### Success Criteria
- [ ] After a deployment is complete, we run an automated check that the expected assets are all available from the origin server and the CDN
- [x] This check should also be able to be manually run
- [x] If an asset is missing, we are told which path(s) to invalidate on the CDN
- [ ] Stretch: if an asset is missing, we automatically re-run the check, and keep doing so until all is well.

PR #746, opened 2025-11-06
(feat) Add retries and backoff to the asset-check command

stevejalim:
## One-line summary

Makes the asset-check more resilient if the CDN comes back with a timeout because it's currently fetching the asset.

## Issue / Bugzilla link

#697

## Testing

Unit tests are enough for this


stephaniehobson (approved):
R+

PR #748, opened 2025-11-06
[WT-401] Article pages

maribedran:
## One-line summary

Create the Article Index and Detail page types replacing the unused Feature pages.

- Replace the existing unused Feature Index and Detail page types with Article Index and Detail
- The Article page classes used to be abstract models and were made concrete models
-  Replace the Feature CTA snippet with a new Download Firefox CTA snippet. Take note of the conditional display here; `media/css/cms/flare-utilities.css` contains a handful of classes that can be applied anywhere. We also have a FE component that can be wrapped around anything, and can be populated by the CMS where that makes sense. We're already using this in the WNP components for conditionally-displayed cards: 

```
<include:conditional-display condition="{{ value.settings.show_to }}">
...
</include:conditional-display>
```

In this PR, we're using it for the "Do more with Firefox" block at the bottom of the page. This is a reusable CMS snippet that can be used as a global CTA on non-firefox browsers. Open this page in Chrome to see it: 

https://www-demo1.springfield.moz.works/en-US/firefox-browser-features/

## Minor design deviations
- The body background of the page is now `ash` instead of white, with the article container itself sticking to white. This was necessary not so much for the article page, but for the new index page. The card designs on the index page require a non-white background, because otherwise the hover state of those outlined cards looked broken. 
- You'll also notice that the "do more" image is slightly bigger. We used the same image as you had before, but stuck with the new banner component so this isn't a special snowflake in the new system. I felt like using that as-is made the image way too big, so I constrained the whole component to a smaller max-width. That helped get it back down to a reasonable size. 
- There is a feature image in the CMS that is currently only used on the index page, if a page is marked as "is featured" in the CMS. In the 2026 design version, that will also be displayed on the detail page, to match the Figma mockups. 

We've stood up a handful of pages on demo1 to give you a feel for the index page. If you want to recreate these locally, build an "Article index page", call it "features", and then build feature pages that are children of that page. 

https://www-demo1.springfield.moz.works/en-US/firefox-browser-features/
https://www-demo1.springfield.moz.works/en-US/firefox-browser-features/bookmark-manager/

CMS direct links: 
features index: https://www-demo1.springfield.moz.works/cms-admin/pages/16/edit/
feature page: https://www-demo1.springfield.moz.works/cms-admin/pages/21/edit/
 
## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-401

## Testing locally

- Make sure to run migrations (this will delete any existing feature pages you may have on your DB)
- Create new Articles as children in the index page
- Set some of the articles as featured and check that the index page displays them in bigger cards
- Create a Download Firefox snippet and add it to an Article
  - The snippet should not be displayed on Firefox
  - Check the download links in older OSs:
    - Windows <= 8.1
    - Mac <= 10.14
 

Screenshot 2025-12-09 at 10 41 08 PM

Screenshot 2025-12-09 at 10 40 04 PM

Screenshot 2025-12-09 at 10 40 12 PM

### DIFF
--- a/media/css/cms/flare-theme.css
+++ b/media/css/cms/flare-theme.css
@@ -27,7 +27,7 @@ runtime conditions. Design tokens live in flare-tokens.css.
     --fl-theme-color-text-body: var(--token-neutrals-black);
 
     /* Core surface colors */
-    --fl-theme-surface-page: var(--token-neutrals-white);
+    --fl-theme-surface-page: var(--token-neutrals-ash);
     --fl-theme-surface-panel: var(--token-neutrals-ash);
     --fl-theme-surface-elevated: var(--token-neutrals-white);
     --fl-theme-surface-hover: var(--token-neutrals-white);


kkellydesign (changes requested):
Some things need merged before this one: 

* The template doesn't _quite_ support dark mode yet, so the light-dark refactor PR should go up before this one, so my fixes won't create merge conflicts: https://github.com/mozmeao/springfield/pull/717

* We need to support the fonts that exist on the current site. My understanding is the sans-serif headlines are WNP-only. Let's confirm today though.

stevejalim (approved):
Backend R+ on this, but still needs a FE review

PR #752, opened 2025-11-07
Remove hard-coded inline notification from What's New Page template

kkellydesign:
## One-line summary
The hard-coded "Your Firefox has been updated" notification has been removed from the template, so that it can be added manually with different color options. 
(Local screenshot) 

Screenshot 2025-11-07 at 6 30 12 AM

## Testing
When looking at this on production, the WNP 145 data will only have one notification at the top after this is merged in.


stevejalim (approved):
Makes sense to me, thanks! r+

(I'm assuming the `pre_content` block has not default content that will bubble up instead)

PR #755, opened 2025-11-07
Fix bg image inline styles

kkellydesign:
## One-line summary

The QR code no longer relies on inline style attributes

Screenshot 2025-11-07 at 2 25 03 PM

Additionally, I added some help text to the CMS to help with sizing and placement of the QR code. 

Screenshot 2025-11-07 at 3 46 57 PM


maribedran (approved):

stephaniehobson (commented):
I think I need more information about what this PR is trying to change. I'm guessing:

- [x] remove the simple background option (and associated inline style)
- [ ] place QR code appropriately when a 1200x675 image is provided
- [ ] place QR code appropriately when no image is provided.


I see an adorable tiny QR code when no image is picked. 

I see that this is existing code but we should still address it - here or in a separate PR - We should be giving screen reader users the context that this is an image with a QR code rather than just hiding it from the screen readers (maybe they want to open it on their phones 😉 ) Let me know if you want a recommendation on how to structure that.

springfield/cms/blocks.py
```diff
@@ -840,10 +840,13 @@ def BannerBlock(allow_uitour=False, *args, **kwargs):
 
     class _BannerBlock(blocks.StructBlock):
         settings = BannerSettings()
-        image = ImageChooserBlock(required=False)
+        image = ImageChooserBlock(
+            required=False,
+            help_text="To use as a QR Code background, this image should be 1200x675, expecting a 300px square directly in the center",
```
maureenlholland: 👏 great use of help text

springfield/cms/templates/components/banner.html
```diff
@@ -25,12 +25,11 @@
 
   <div class="fl-banner-layout">
     {% if contents.media or qr_code %}
-      <div class="fl-banner-media" aria-hidden="true" {% if media_url and qr_code %} style="background-image: url('{{ media_url }}');" {% endif %}>
+      <div class="fl-banner-media {% if qr_code %} has-qr-code{% endif %} {% if qr_code and not contents.media %} has-large-qr-code{% endif %}" aria-hidden="true">
         {% if qr_code %}
           <div class="fl-banner-qr" role="img">{{ contents.qr }}</div>
```
maureenlholland: We have some existing translations for "Scan this QR code" on the evergreen WNP

https://github.com/mozilla-l10n/www-firefox-l10n/blob/164f29690690aeb56b852ae46f7f2dc34b04dee9/fr/firefox/whatsnew/evergreen.ftl#L16

https://github.com/mozilla-l10n/www-firefox-l10n/blob/164f29690690aeb56b852ae46f7f2dc34b04dee9/de/firefox/whatsnew/evergreen.ftl#L16

maureenlholland (approved):
r+wc

going to merge so we know we have removed the inline style attribute that will trigger a million sentry reports tomorrow

a separate PR should address the aria-hidden and qr sizing issues (WNP145 is using a QR code embedded in an image already, so is not affected by the qr code styling here)

PR #763, opened 2025-11-10
Properly Render Translations For Mixed-Case Locales

dchukhin:
## One-line summary
Translations in mixed-case locales (like "en-GB") should now be properly rendered.

It looks like mixed case locales get lowercased by Django, so translations of Wagtail pages end up looking for a page in a lowercase locale (such as "en-gb"). When they are not found, the English Page is rendered. The fix here is to override `Locale`'s `get_active()` method to return the locale in mixed-case ("en-GB").

## Testing
To recreate the issue locally, check out the `main` branch, and create a translation in the "en-GB" locale (you need to create the locale first if you don't have it locally). Then, translate a string in the en-GB translation, publish the en-GB translation, and go to the en-GB page, and observe that the en-US text shows up.
Then, switch to this branch, and reload the en-GB page and observe that the eb-GB text shows up.


stevejalim (approved):
r+ - makes sense in how the patch is being applied, and the tests back that up.

I'm still wondering why the monkey-patch here https://github.com/mozmeao/springfield/blob/main/springfield/base/middleware.py#L152-L176 isn't already fixing up translation.get_language(), because it should, via https://github.com/mozmeao/springfield/blob/main/springfield/base/middleware.py#L113-L149 - that's code which works for `es-ES` in Bedrock (and is the same as the code here)

I think for now we'll definitely use this fixup, but it'd be interesting to explore why the existing monkey-patch didn't work @dchukhin

Issue #466, opened 2025-09-10
remove linux32 download links for firefox 145+


jcristau:
## Description
Per https://blog.mozilla.org/futurereleases/2025/09/05/firefox-32-bit-linux-support-to-end-in-2026/, we're planning to stop 32-bit linux support starting with firefox 145.  We should therefore remove the corresponding download links etc.

On the product side the removal will happen in https://bugzilla.mozilla.org/show_bug.cgi?id=1987132

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-142)
┆Priority: P3
┆Sprint: Backlog

janbrasna:
This seems to start with nightlies today:
https://github.com/mozmeao/springfield/actions/runs/18099754054/job/51499393419#step:3:14542

As all the randomly sampled URLs are 145.0a1 "i686" ones.
https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central-l10n/firefox-145.0a1.es-MX.linux-i686.tar.xz
https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central-l10n/firefox-145.0a1.sk.linux-i686.tar.xz
https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central-l10n/firefox-145.0a1.ko.linux-i686.tar.xz

janbrasna:
Some places that hardcode the OS to get the idea: [/search?q=repo:mozmeao/springfield+os=linux](https://github.com/search?q=repo%3Amozmeao%2Fspringfield%20os%3Dlinux&type=code)

janbrasna:
For now this covers user-visible surfaces. There will be some changes necessary to work around ESR train overlap in 2026 (possibly supporting three different major versions, each having its own set of platform support, both additions and removals — and the complexity it brings in allowing to navigate to appropriate downloads at such point…), its timing TBA around March 2026.

For the things mentioned in the thread above, some followups were created:
- improving https://github.com/mozmeao/springfield/issues/648
- & dropping conditionals from https://github.com/mozmeao/springfield/pull/640
- and also somehow sorting out accessing old direct links to the linux32 drilldown pages (for products no longer built for the arch) https://github.com/mozmeao/springfield/issues/650

PR #769, opened 2025-11-11
Update primary Linux platform CTA to x64

janbrasna:
## One-line summary

Linux platform landing page shows smaller selection of archs when accessed from non-Linux browser, the default being i686 32bit that's no longer shipping. This uses x64 as the default.

This looks reasonably and by design self-contained to that page, so should not influence any other occurrence in case it could complicate any layout elsewhere.

This change adds a little more space that was most likely supposed to be separating the buttons from the APT/RPM message, but was only applied in cases when both buttons were shown. This now applies the same spacing no matter what all is shown (as the spacing is tied to the x64 variant specifically, that's now there always).

## Issue / Bugzilla link

#466

## Testing

1. with non-Linux user-agent
2. with "Linux" user-agent

### DIFF
--- a/tests/playwright/specs/products/firefox/firefox-platforms.spec.js
+++ b/tests/playwright/specs/products/firefox/firefox-platforms.spec.js
@@ -97,7 +97,7 @@ test.describe(
             browserName
         }) => {
             const downloadButton = page.locator(
-                '#download-button-desktop-release-linux'
+                '#download-button-desktop-release-linux-64'
             );
             const repoLink = page.locator('.c-linux-debian a');
 


stevejalim (approved):
Tested locally and this works.

Trying to think through whether the change in default will frustrate anyone, but then again, if 32-bit downloads aren't available via bouncer, they aren't available.


@stephaniehobson It might be that there's some code cleanup we can do to drop all 32-bit support now; let's check in later. I'm also pinging the bouncer team to check

janbrasna (commented):
Given we no longer have platform bitness https://bugzil.la/1861847 as of v123, there can't be any reasonable messaging through, say, unsupported macros to point folks towards ESR for the next year — so this needs to rely on the wording around all the CTA places: "download this and this" … "or other platforms and languages" — and have folks figure out something there, even when they need some preexisting knowledge to be able to find the platform they may need. But, realistically, the builds are dropped for no real interest in them. And, also, really, Linux folks get their software through other means than binary or tarball downloads.

I'm tackling the platform-specific occurrences in a followup #776

janbrasna:
> "[…] code cleanup we can do to drop all 32-bit support now"

No, ESR will support it for another year, so that's a task for November 2026.

(With the next few ESR cycles there will be more special-casing in cards, to allow platforms correctly for situations where a combination of esr-old, esr-latest and esr-next are offered by the download functionality, at [March, July and October](https://whattrainisitnow.com/release/?version=esr) release milestones.)

> "I'm also pinging the bouncer team to check"

… whether they plan to update staging and prod bouncer results for more testing parity?;] Or what was the idea that got truncated @stevejalim? O:)

PR #770, opened 2025-11-11
Refine page titles as suggested and introduce `block page_title_region` (#385)

janriokrause:
Introduced a `block page_title_region` to cleanly append region names to selected English pages, preventing duplicate content (issue 13019). Also changed "Firefox" to "Firefox.com".

Testing:

```
$ curl --silent "http://localhost:8000/en-US/browsers/"{"mobile/ios","desktop","mobile"}"/" | grep "<title>"
    <title>Get Firefox for iOS (iPhone and iPad) from Mozilla — Firefox.com (US)</title>
    <title>Firefox for Desktop available for Windows, Mac and Linux — Firefox.com (US)</title>
    <title>Firefox mobile browsers put your privacy first — Firefox.com (US)</title>
```

### DIFF
--- a/springfield/base/templates/base-protocol.html
+++ b/springfield/base/templates/base-protocol.html
@@ -28,7 +28,7 @@
     {% block extra_meta %}{% endblock %}
 
     {% block shared_meta %}
-    <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox{% endblock %}{% endfilter %}</title>
+    <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox.com{% endblock %}{% block page_title_region %}{% endblock %}{% endfilter %}</title>
     <meta name="description" content="{% filter striptags %}{% block page_desc %}{% endblock %}{% endfilter %}">
     <meta property="og:type" content="website">
     <meta property="og:site_name" content="Firefox">
--- a/springfield/firefox/templates/firefox/browsers/desktop/index.html
+++ b/springfield/firefox/templates/firefox/browsers/desktop/index.html
@@ -8,13 +8,16 @@
 {% from "macros.html" import google_play_button, apple_app_store_button, fxa_email_form with context %}
 
 {# Issue 13019: Avoid duplicate content for English pages. #}
+{%- block page_title_region -%}
+  {% if LANG == 'en-US' %} (US){%- endif -%}
+  {% if LANG == 'en-GB' %} (UK){%- endif -%}
+{%- endblock -%}
+
 {%- block page_title -%}
-  {%- if LANG == 'en-US' -%}
-    Firefox for Desktop — Firefox (US)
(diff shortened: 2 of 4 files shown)


springfield/firefox/templates/firefox/browsers/desktop/index.html
```diff
@@ -8,13 +8,16 @@
...
+  {% if LANG == 'en-GB' %} (UK){%- endif -%}
+{%- endblock -%}
+
 {%- block page_title -%}
-  {%- if LANG == 'en-US' -%}
-    Firefox for Desktop — Firefox (US)
-  {%- elif LANG == 'en-GB' -%}
-    Firefox for Desktop  — Firefox (UK)
+  {%- if LANG == 'en-US' or LANG == 'en-GB' -%}
+    Firefox for Desktop available for Windows, Mac and Linux
```
janriokrause: After outsourcing the region, we could go back to using the FTL files here, right? Or what was the reason they weren't used here in the first place?

springfield/firefox/templates/firefox/browsers/mobile/ios.html
```diff
@@ -22,17 +22,20 @@
...
-{# Issue 13019: Avoid duplicate content for English pages. #}
 {%- block page_title_full -%}
-  {%- if LANG == 'en-US' -%}
-    Get Firefox for iOS (iPhone and iPad) — Mozilla (US)
-  {%- elif LANG == 'en-GB' -%}
-    Download Firefox for iOS (iPhone and iPad) — Mozilla (UK)
+  {%- if LANG == 'en-US' or LANG == 'en-GB' -%}
+    Get Firefox for iOS (iPhone and iPad) from Mozilla
   {%- else -%}
     {{ ftl('mobile-ios-firefox-browser-ios') }} - Mozilla
```
janbrasna: Here BTW the format for RoW seems to keep the old style though then. For not introducing a new l10n string just for potential "from Mozilla" here I guess?

springfield/base/templates/base-protocol.html
```diff
@@ -28,7 +28,7 @@
     {% block extra_meta %}{% endblock %}
 
     {% block shared_meta %}
-    <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox{% endblock %}{% endfilter %}</title>
+    <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox.com{% endblock %}{% block page_title_region %}{% endblock %}{% endfilter %}</title>
```
janbrasna: Given this is a global base template, I'd ask around first if that is a pattern worth encouraging in the first place.

My understanding is when a customization is necessary, there's a reason, and it should be explicit at that exact page, with ideally tailored titles fully.

(Not sure if the concern is still valid after these years though.)

janbrasna (commented):
There definitely are some changes necessary for those repetitive titles, and I would like to see them simplified or rewritten personally.

I'm not sure the questions raised in the tracking ticket were even being considered yet though, they may need to go through martech folks or search agency first.

(But hopefully there will be at least a part of these changes reviewable without external input.)

maureenlholland:
closing in favour of https://github.com/mozmeao/springfield/pull/900

PR #771, opened 2025-11-12
Add `noindex` to pages excluded from sitemap (#370)

janriokrause:
Add `noindex` meta tag to `404` and `/browsers/unsupported-systems/` pages.

### DIFF
--- a/springfield/base/templates/404.html
+++ b/springfield/base/templates/404.html
@@ -15,7 +15,7 @@
 {% endblock %}
 
 {# 404 pages must not include canonical and hreflang x-default: https://github.com/mozilla/bedrock/issues/5895 #}
-{% block canonical_urls %}{% endblock %}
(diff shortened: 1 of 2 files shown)


springfield/base/templates/404.html
```diff
@@ -15,7 +15,7 @@
 {% endblock %}
 
 {# 404 pages must not include canonical and hreflang x-default: https://github.com/mozilla/bedrock/issues/5895 #}
-{% block canonical_urls %}{% endblock %}
+{% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
```
janbrasna: Is indexing of error pages a concern?

(That said, should this canonical block be shared right from the parent `base-error` template?)

stevejalim (approved):

PR #774, opened 2025-11-12
All a11y 2

Ingi-Hong:
## One-line summary
Ported from https://github.com/mozilla/bedrock/pull/16172

The [speech reader areas commit](https://github.com/mozilla/bedrock/pull/16172/commits/4f2bf1355a9c73628eeacd99a02ef540d83f552c) was not ported over from the original branch. 

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/15082

## Testing
See original pr

### DIFF
--- a/springfield/firefox/templates/firefox/all/includes/main.html
+++ b/springfield/firefox/templates/firefox/all/includes/main.html
@@ -16,7 +16,7 @@
               <ul>
                 <li><a href="{{ firefox_url('desktop', 'sysreq', 'release') }}">{{ ftl('firefox-all-check-the-system-requirements') }}</a></li>
                 <li><a href="{{ firefox_url('desktop', 'notes', 'release') }}">{{ ftl('firefox-all-release-notes') }}</a></li>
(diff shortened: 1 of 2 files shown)


janbrasna:
Would it be possible to deploy this to a demo for a wider review? Appreciate it.

Ingi-Hong:
Nice. Thanks for taking care of that. Was travelling for a little bit

janbrasna:
This also confuses any domready-triggered scripts, that only fire on the first step, but do not affect anything further:
- https://github.com/mozmeao/springfield/issues/959

I'm not sure if there's a good way to have all the Mozilla.Utils.onDocumentReady stored somewhere to be able to trigger the callbacks again when the partials change — or a more realistic approach is to special case just the attribution now and specifically trigger its binding logic to invoke it after the partials are replaced…?

However as both this and the focus scroll position are currently affecting the download flow in production rather seriously, I'd say reverting for now before solutions are found is the safer path — unless anyone has a couple of quick fixes for these in mind.

Issue #648, opened 2025-10-09
Consider adding Linux ARM download CTAs in place of removed 32-bit


janbrasna:
## Description

With Linux32 buttons going away #466 — and the ask to add Linux ARM64/Aarch64 to channels https://github.com/mozilla/bedrock/issues/14613 — the easiest path could be using the same tandem buttons:

<img width="982" height="413" alt="Image" src="https://github.com/user-attachments/assets/fb94a0ce-2298-45f8-8c40-1289527108d1" />

and just make them from 32+64 that currently have to adapt based on channel availability to 64+ARM going forward.

---

#### Success Criteria
- [ ] Consider moving from Linux32 + Linux64 dual buttons to Linux64 + LinuxARM64
- [ ] Update tests tracking presence of CTA buttons for the new platform

PR #776, opened 2025-11-13
Hide explicit `linux-32` download options

janbrasna:
## One-line summary

These are the remaining, hard-coded, and special-cased places that manually include `os=linux&product=firefox-latest` in their download flows.

I do realize this is the ugly thing to do, and leaving commented out code in production is sloppy — I just ideally see #648 taking its place there. (But might need some wording help first.)

## Issue / Bugzilla link

#466

## Testing

(with a Linux UAstring)
http://localhost:8000/en-US/thanks/?xv=basic
http://localhost:8000/en-US/thanks/

### DIFF
--- a/tests/playwright/specs/products/firefox/firefox-platforms.spec.js
+++ b/tests/playwright/specs/products/firefox/firefox-platforms.spec.js
@@ -116,7 +116,7 @@ test.describe(
             await downloadButton.click();
 
             const download = await downloadPromise;
-            expect(download.url()).toContain('linux64');
+            expect(download.url()).toContain('linux-x86_64');
 
             // Cancel download
             await download.cancel();


janbrasna (commented):
For now I'm keeping them commented out, with the idea to reuse that for #648

_I'm unable to run springfield at the current head, and trying to test this on a version from back when I had a working image was super flaky, the more I re-run, the more errors/timeouts Playwright had, so I could not verify this fully — a demo deploy to take it for a spin from a few platforms, and a push to run integration tests would be helpful here._

stevejalim:
> @stevejalim IIUC the password-protected demos seem to stand up the basic auth; however subsequent mere-mortal demo pushes do not clear that auth requirement apparently ("Unauthorized. This site is in private demo mode.")
> 
> No rush, just a fun kink.

Oh, that is a surprise -- the Cloud Build logs don't pass those to Cloud Run, and i can see `_ENABLE_BASIC_AUTH: false` logged in the output... Have opened https://github.com/mozilla/bedrock/issues/16878

stevejalim:
> Actually a `test` deployment push would just do, as a demo of sorts for a bit before the tests finish anyways;D ty

Just pushed to springfield-test - as noted in the GHAs above

stephaniehobson (approved):
R+ 

I can't test in Linux but it sounds like @janbrasna did a through job.

stephaniehobson:
Successful integration test run: https://github.com/mozmeao/springfield/actions/runs/20178611521

PR #778, opened 2025-11-13
New waitlist

stevejalim:
Already signed off via https://github.com/mozmeao/springfield-private/pull/10

### DIFF
--- a/springfield/firefox/templates/firefox/ai/waitlist.html
+++ b/springfield/firefox/templates/firefox/ai/waitlist.html
@@ -30,7 +30,7 @@
     <h2 class="c-heading">AI Window: Built for choice &amp; control</h2>
     <div class="c-body">
       <p><strong>The web is changing, and sitting it out doesn’t help anyone.</strong></p>
-      <p>We were founded under the mission that the web should be open, fast, secure and people first. As AI becomes a larger part of the web, we see a clear role: to shape it in ways that supports choice and stays true to the values that make the internet open and accessible for everyone. With AI Window, we’ll do this through:</p>
+      <p>We were founded under the mission that the web should be open, fast, secure and people first. As AI becomes a larger part of the web, we see a clear role: to shape it in ways that support choice and keep the internet open and accessible to all. With AI Window we’ll do this through:</p>
       <ul class="mzp-u-list-styled">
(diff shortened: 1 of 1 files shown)


springfield/firefox/templates/firefox/ai/waitlist.html
```diff
@@ -0,0 +1,146 @@
...
+      <input type="hidden" name="source_url" value="{{ request.build_absolute_uri() }}">
+      <input type="hidden" name="lang" id="id_lang" value="en">
+      <label class="visually-hidden" for="newsletter-email">{{ email_label|d(ftl('newsletter-form-your-email-address'), true) }}</label>
+      <div class="c-input-wrapper">
+        <input data-testid="newsletter-email-input" id="newsletter-email" class="mzp-js-email-field" type="email" name="email" placeholder="email@example.com" autocomplete="email" class="c-input" />
+        <button data-testid="newsletter-submit-button" id="newsletter-submit" type="submit" class="c-submit">
+          <span class="c-submit-text">Join the waitlist</span>
+          <svg class="c-submit-loading" fill="currentColor" width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+            <title>Sending</title>
+            <style>.s1{animation:s .8s linear infinite;animation-delay:-.8s}.s2{animation-delay:-.65s}.s3{animation-delay:-.5s}@keyframes s{93.75%,100%{opacity:.2}}</style>
```
janbrasna: CSP blocks this inline style. This kind of SVG only works linked as an image (it's available as [loading.svg](https://github.com/mozmeao/springfield/blob/main/media/img/icons/loading.svg) in icons.)

janbrasna: (e.g. if this output, with just "bubbles", not animating, is acceptable, then just the `<style/>` block can be removed to not submit the violations, and left as-is for this page…)

So the easy thing here would be to just externalize that to an img, but it probably will only start loading once shown, it would have to be preloaded in some way for the intended effect. (It was really added to be used as a CSS background and swapped in-place ~ however I'm not familiar with the FDS markup to be looking into where to place it for the button to have it preloaded best.)

PR #780, opened 2025-11-13
feat: add Slack pings for the static-assets check progress

### LINKED ISSUE #697: Support post-deployment check for assets

stevejalim:
Now that our static-asset checks are running once a deployment is done, we should alert Slack about it. This changeset adds that support, in the same way we do for our Integration Tests. 

Also, while in the area, we drop an unnecessary codebase checkout for the actual assets-check step.

## Issue / Bugzilla link

Resolves #697 

## Testing

Eyeballing for now - we can test as soon as it hits dev

### DIFF
--- a/.github/workflows/post_deploy_asset_check.yml
+++ b/.github/workflows/post_deploy_asset_check.yml
@@ -24,13 +24,15 @@ jobs:
   notify-of-checks-start:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          sparse-checkout: |
+            .github/actions/slack
       - name: Notify via Slack that tests are starting
         uses: ./.github/actions/slack
         with:
           channel_id: ${{ env.SLACK_CHANNEL_ID }}
           env_name: test
-          fetch-depth: 1
           label: "Static-assets check for [${{ inputs.git_sha }}]"
           message: "Static-assets check started"
           ref: ${{ inputs.branch }}
@@ -61,14 +63,16 @@ jobs:
     runs-on: ubuntu-latest
     needs: [asset-check]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          sparse-checkout: |
+            .github/actions/slack
       - name: Notify via Slack of test-run success
         if: ${{ needs.asset-check.result == 'success' }}
         uses: ./.github/actions/slack
         with:
           channel_id: ${{ env.SLACK_CHANNEL_ID }}
           env_name: test
-          fetch-depth: 1
           label: "Static-assets checks for [${{ inputs.git_sha }}]"
           message: "Static-assets check completed. Status: success"
           ref: ${{ inputs.branch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/post_deploy_asset_check.yml
```diff
@@ -19,15 +21,25 @@ on:
...
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
+      - name: Notify via Slack that tests are starting
+        uses: ./.github/actions/slack
         with:
-          ref: ${{ inputs.git_sha }}
-          fetch-depth: 0
+          channel_id: ${{ env.SLACK_CHANNEL_ID }}
+          env_name: test
+          fetch-depth: 1
```
janbrasna:
```suggestion
```

janbrasna (commented):
Just some extra copypasta artifacts sneaked in:

.github/workflows/post_deploy_asset_check.yml
janbrasna: _(not familiar with the slack action code, but assuming it doesn't need to access any paths other than its own code so it could be the only folder checked out instead of the whole tree, but didn't really study up what it ends up running.)_

.github/workflows/post_deploy_asset_check.yml
```diff
@@ -39,3 +51,46 @@ jobs:
...
+          ref: ${{ inputs.branch }}
+          slack_bot_token: ${{ env.SLACK_BOT_TOKEN }}
+          status: "success"
+      - name: Notify via Slack of test-run failure
+        if: ${{ needs.asset-check.result == 'failure' }}
+        uses: ./.github/actions/slack
+        with:
+          channel_id: ${{ env.SLACK_CHANNEL_ID }}
+          env_name: test
+          fetch-depth: 1
```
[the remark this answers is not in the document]
janbrasna: test-run `failure` step still has the fetch-depth ^

janbrasna (approved):
r+wc _(There are two more occurrences of fetch-depth, on a different action context, to be removed)_

Let's see if the slack action is happy this way, it can always be tweaked if it shows any rough edges when in use. LGTM :shipit: 🚢

PR #781, opened 2025-11-13
WT-441 Video Block + Banner Refactoring 

maribedran:
## One-line summary

Add a new Video field to the Intro, Media + Content, and Banner blocks, and refactor the Banner block.

This PR is based on @localjo's work on #741, which adds a new video component, and implements it on the CMS. It also refactors the Banner block splitting it into the regular outline banner and the filled banner with brand image.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-441

## Testing

- Make sure to run migrations
- Check the Intro, Media + Content and Banner variations on the Patter Library
- Test all the variants from those blocks in a Free Form or What's New Page

Intro
image

Media + Content

image

Banner
image

Filled Banner

image

Small Curious  Kit

image

Large Curious Kit

image

Sitting Kit

image

Kit Tail

image

### DIFF
--- a/springfield/cms/fixtures/intro_fixtures.py
+++ b/springfield/cms/fixtures/intro_fixtures.py
@@ -18,8 +18,7 @@ def get_intro_variants() -> list[dict]:
             "type": "intro",
             "value": {
                 "settings": {"media_position": "after"},
-                "image": None,
-                "dark_image": None,
+                "media": [],
                 "heading": {
                     "superheading_text": '<p data-block-key="ybdoh">Superheading text</p>',
                     "heading_text": '<p data-block-key="uzief">Simple Intro with text and button</p>',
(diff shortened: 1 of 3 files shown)


kkellydesign:
@maribedran I just found the draft of WNP 146 that they're working on, so we'll need to be able to support the purple background in the normal banner component. I just updated the CSS to handle it, it just needs the BE implemented. Lets go ahead and make it a theme dropdown to be consistent. The default should be "Outline" and the other option should be "Purple Background". It should toggle the classes on the banner between `fl-banner-outlined` and `fl-banner-purple`. 

Screenshot 2025-11-20 at 5 00 57 PM

stephaniehobson:
This is a great simplification :)

Links in the filled/purple banners are showing up blue and so they're difficult to read.

stephaniehobson (commented):
I don't have time to do a super through review but I had a quick look and noted a few things.

kkellydesign (changes requested):
The change request is in my comment, but there is some additional context from Stephanie a couple of notes above mine: 
https://github.com/mozmeao/springfield/pull/781#issuecomment-3564339648

maribedran:
@stephaniehobson I'm trying to understand how this works on bedrock since there aren't any examples in springfield. The class is used there in `video` elements, but bedrock also uses YT embeds. I didn't spend much time on it, but I couldn't figure out how this applies to embedded videos. Could it be the case that we're not tracking this at all for embeds?

maureenlholland:
@localjo I can confirm this is correct

>  it looks like tracking analytics for YouTube embeds is handled separately, and we only need to handle CDN video analytics here

YouTube video embeds are handled automatically by GA4

For CMS-loaded CDN videos, we may have to revisit the class + js bundle requirements, I've haven't looked into that yet

PR #782, opened 2025-11-14
use env to pass input parameters

fkiriakos07:
## One-line summary

use env to pass input github parameters

make sure the jobs are running without errors


stevejalim:
Thanks - this'll need doing over on Bedrock too

stevejalim (approved):
r+

Issue #765, opened 2025-11-11
Add shorter URL for app redirector to simplify QR code density


janbrasna:
## Description

The longer the URL, the more data stored in the QR code, the "busier" the output image — needing either larger physical dimension to be legible, or less correction data to stay smaller, (and/or) risking worse scalability from different displays.

When added manually, folks often link through mzl.la bitly shortlinks for very simple, clean and super scannable QR codes for that short URL content — but in cases the QRs are auto–generated from within code, including e.g. page or campaign details, the links tend to get longer, and the codes less coarse.

Luckily the current qrcode() helper renders very legible codes without any extra fluff or overhead complicating the resulting matrix, so there's no need to optimize the rendering itself, but storing shorter URLs would help uses when the resulting image is smaller, and needs less pixel amount to stay legible.

(The current URL is rather complicated, and seemingly unnecessarily; not sure if it is actually bearing any meaning or needs structuring that way, but the easiest is to keep the existing rule for current param verbosity, and adding the new endpoint with reduced param length, to co-exist.)

---

#### Success Criteria
- [ ] An alternative URL is mapped to the mobile store redirector functionality, with shorter path, shorter param names, and explicit default to be left out most of the time.
- [ ] Consumers of the qrcode() helper update to shorter input URLs mainly in places where space / resolution of the output image is a factor.
- [ ] It is verified that using e.g. apex domain, few slashes etc. for simplification does not have adverse effects on performance regarding more redirects (incl. different services with apex coming through refractr in case we wanna omit the www. for extra brevity when used…)

PR #792, opened 2025-11-19
Add vanity /qr URL aliasing mobile_app redirector

janbrasna:
## One-line summary

The shorter the URL, the more readable the QR code gets. This adds some aliasing to be able to link shorter mobile_app redirector URLs.

Shorter params could also be used on the original endpoint, there's no restriction.

Tests were added to demonstrate mixing'n'matching the patterns between the approaches.

## Issue / Bugzilla link

#765

## Testing

`pytest -k test_redirects.py`
http://localhost:8000/qr
http://localhost:8000/qr?c=foo
http://localhost:8000/fy-NL/qr/?product=quux
http://localhost:8000/de-BE/browsers/mobile/app/?p=klar&campaign=qr-focus (will show store error outside germany unless using android ua)
http://localhost:8000/browsers/mobile/app/?product=focus&campaign=qr-focus (control)


janbrasna (commented):
Wondering where to document this endpoint alias best so it's not just tribal knowledge. Ideally in some code using it, but there's no consumer yet 😅 🤷

PR #796, opened 2025-11-20
WT-431: Flare navigation

lucianoratamero:
## One-line summary
This PR reimplements the navbar using styles and theming from flare.

- light/dark modes
- mobile
- keyboard navigation

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-431

### DIFF
--- a/media/css/cms/flare-navigation.css
+++ b/media/css/cms/flare-navigation.css
@@ -17,8 +17,6 @@
    ================================================ */
 
 .fl-header {
-    --fl-theme-font-family-headline: var(--token-font-family-mozilla-headline);
-    --fl-theme-font-family-body: var(--token-font-family-mozilla-text);
     padding: 18px calc(var(--token-scale-40) * 1px);
     border-bottom: 2px solid var(--fl-theme-border-color-subtle);
     background: var(--fl-theme-navigation-bg);


media/img/cms/close.svg
localjo: There's already a close icon that's referenced in flare-icon.css that relies on a single SVG that I think we could use instead of adding two more redundant SVGs.

springfield/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -18,10 +19,17 @@
...
-      <!-- TODO: replace with new protocol-assets version -->
-      <img src="{{ static('img/icons/m24-small/download-white.svg') }}" alt="" />
-    </span>
+    {% if flare_styles %}
+      <include:icon
+        icon_name="arrow-down"
+        hidden=true
+        extra_class="fl-icon-right" />
+    {% else %}
+      <span class="mzp-c-button-icon-end">
```
localjo: The `mzp-c-` prefix is from Mozilla Protocol, and this shouldn't have any Protocol dependencies.

springfield/firefox/templates/firefox/includes/download-unsupported.html
```diff
@@ -23,13 +23,33 @@
   <p>{{ ftl('download-button-please-download-esr') }}</p>
   <div class="download-platform-list">
     <p>
-      <a class="mzp-c-button mzp-t-product download-link os_win64 ga-product-download" href="https://download.mozilla.org/?product=firefox-esr115-latest-ssl&os=win64&lang={{ win_download_lang }}" data-download-version="win64" data-cta-text="Download Firefox Extended Support Release 64-bit" data-cta-type="firefox">
+      <a class="{% if flare_styles %}fl-button button-secondary{% else %}mzp-c-button mzp-t-product{% endif %} download-link os_win64 ga-product-download"
```
localjo: I don't think the mzp class should be here either.

springfield/firefox/templates/firefox/includes/download-unsupported.html
```diff
@@ -23,13 +23,33 @@
...
+          <include:icon
+            icon_name="arrow-down"
+            hidden=true
+            extra_class="fl-icon-right" />
+          {% endif %}
       </a>
     </p>
     <p>
-      <a class="mzp-c-button mzp-t-product download-link os_win ga-product-download" href="https://download.mozilla.org/?product=firefox-esr115-latest-ssl&os=win&lang={{ win_download_lang }}" data-download-version="win" data-cta-text="Download Firefox Extended Support Release 32-bit" data-cta-type="firefox">
+      <a class="{% if flare_styles %}fl-button button-secondary{% else %}mzp-c-button mzp-t-product{% endif %} download-link os_win ga-product-download"
```
localjo: Another instance of the mzp classes that might not belong. I think there are a few more. Rather than commenting on each one, it might be good to search/replace.

localjo (dismissed):
Mostly just feedback about classnames. There's also a CSS pattern that we're trying to use with overriding `--fl-theme` variables for media queries, rather than wrapping a bunch of rules inside media query blocks. Here's an example from Slack: https://lincolnloop.slack.com/archives/CHYTGGMPF/p1762298368804319?thread_ts=1762298363.450639&cid=CHYTGGMPF

springfield/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -6,9 +6,10 @@
 
 {% set params = '?utm_source=www.firefox.com&utm_medium=referral&utm_campaign=download-button-thanks' %}
 
-<div id="{{ id }}" class="mzp-c-button-download-container c-button-download-thanks">
+<div id="{{ id }}" class="{% if flare_styles %}fl-fx-download-container{% else %}mzp-c-button-download-container{% endif %} c-button-download-thanks">
+
   <a href="{{ transition_url }}"
-     class="download-link c-button-download-thanks-link mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
+     class="download-link c-button-download-thanks-link {% if flare_styles %}fl-button button-secondary{% else %}mzp-c-button mzp-t-product{% endif %} {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
```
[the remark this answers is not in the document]
maribedran: @localjo this is the download firefox button that's used across the entire website. The only modification here is that if it receives the `flare_styles` flat, it uses flare styles instead of the `mzp-c` ones.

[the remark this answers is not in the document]
lucianoratamero: this template is used everywhere, so I'll mark all comments of flare prefixes on these related templates as done, since we can't really do much here.

media/css/cms/flare-theme.css
```diff
@@ -68,11 +68,11 @@ runtime conditions. Design tokens live in flare-tokens.css.
     --fl-theme-code-surface: var(--token-neutrals-grey-3);
 
     /* Typography variables */
-    --fl-theme-font-family-headline: var(--token-font-family-metropolis);
+    --fl-theme-font-family-headline: var(--token-font-family-mozilla-headline);
```
localjo: Changing these font family variables here is also changing which font families are used on the Whats New Page

media/css/cms/flare-theme.css
```diff
@@ -26,7 +26,7 @@ runtime conditions. Design tokens live in flare-tokens.css.
     --fl-theme-color-text-body: var(--token-neutrals-black);
 
     /* Core surface colors */
-    --fl-theme-surface-page: var(--token-neutrals-ash);
+    --fl-theme-surface-page: var(--token-neutrals-white);
```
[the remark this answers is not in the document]
lucianoratamero: yep, I discussed this with Mari, and this will be reverted on the article PR.

media/css/cms/flare-navigation.css
```diff
@@ -0,0 +1,488 @@
...
+   Navigation
+   ================================================ */
+
+.fl-nav {
+    display: none;
+    flex-grow: 1;
+    background: var(--fl-theme-navigation-bg);
+}
+
+.fl-nav a[href^='http']::after {
```
localjo: We've already got icon components that can be used here, instead of duplicating the CSS.

media/css/cms/flare.css
```diff
@@ -48,6 +48,7 @@ https://caniuse.com/?compare=safari+15.6&compareCats=CSS
 @import 'flare-tag.css' layer(components);
 @import 'flare-typography.css' layer(components);
 @import 'flare-intro.css' layer(components);
+@import 'flare-navigation.css' layer(components);
```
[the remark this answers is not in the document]
localjo: GitHub diffs might be playing tricks with me, but it looks like you changed the layer for `flare-notification.css` to `template` while `flare-navigation.css` is still `components`, it should be `flare-notification.css` in `components` and `flare-navigation` in `template`.

springfield/cms/templates/cms/includes/menus/features.html
```diff
@@ -0,0 +1,86 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public License,
```
localjo: I've noticed that on the HTML files you added in this branch, when the pre-commit hook runs, the format of the license doesn't match what it's looking for, so it tries to add it again. If you edit the html files, stage them, then run `pre-commit run`, you'll see what I mean. It might be good to update those now so it doesn't cause issues with the precommit hooks in the future.

kkellydesign (changes requested):
I just added some visual revisions for 1:1 parity with the existing system. We need a couple more things: 

- [ ] Add headroom.js (I assume) for the sticky behavior instead of `position: sticky`. 
- [ ] Check for the behaviors that show or hide the download button, and make sure they're implented here. If you look at the homepage, there's no download button in the masthead. Looking at prod, it might just be "it should be everywhere except the homepage." But We ought to make sure we can honor that in our version if necessary.  
- [x] Add the hover animation for the arrow on the "View all Firefox features" button
- [x] When you're testing in mobile and then expand the browser again, the X icon doesn't go away on the top right next to the download button. 
- [x] hard-code the orange color from the button, as it deviates from the "light orange" color in the pattern library.

stephaniehobson:
When navigating the mobile layout with a keyboard -- focus is not being passed to the open navigation menu and can escape the the open navigation menu.

Steps to reproduce:
1) open site on a small enough screen to trigger mobile layou
2) tab until the "Menu" button has focus
3) press enter to open the menu
4) attempt to tab to "Fingerprint blocking"
5) discover you are instead tabbing around the page content you can't see.

I discover a bug in our current implementation on firefox.com while trying to make a video of the desired behavior. You can see how it should work on www.mozilla.org.

stephaniehobson:
This heading seems a bit close to the top of the page now. I think this occurs on pages when they don't start with an inline notification.

Screenshot 2025-12-04 at 1 27 44 PM

media/css/cms/flare-template.css
```diff
@@ -479,3 +255,69 @@
...
+    text-align: center;
+    margin-bottom: 8px;
+}
+
+.fx-unsupported-message p {
+    margin-bottom: 8px;
+}
+
+/* Show the Windows unsupported message if on Windows with unsupported version for Firefox */
+.windows.fx-unsupported .fx-unsupported-message.win {
```
stephaniehobson: Download buttons styles should be broken out into a separate file and then we need to make sure we're not breaking anything.

media/css/cms/flare-base.css
```diff
@@ -78,6 +78,87 @@ See https://mozmeao.github.io/platform-docs/browser-support/
...
+@font-face {
+    font-display: swap;
+    font-family: 'Mozilla Text';
+    font-style: italic;
+    font-weight: bold;
+    src:
+        url('/media/fonts/mozilla-text/MozillaText-BoldItalic.woff2') format('woff2');
+}
+
+@font-face {
```
stephaniehobson: This font-family needs to be added to the stack if you want it used as a fallback.

stephaniehobson (changes requested):
Sorry, two more.

1) the download button should not display to current Firefox users.
2) the navigation menu shifts so that Features/Resources do not display centered when the button is removed from the DOM.

kkellydesign:
@stephaniehobson quick question about this one:

>  1. the download button should not display to current Firefox users.

Should that always be true? If so, I can just add modifiers for is-firefox  to the nav component. But our work on the nav has been in service of this page, which current has the download button in firefox:
https://www.firefox.com/en-US/features/block-fingerprinting/

Currently, the download button is disabled conditionally on a per-template basis – you can see this in several different base templates throughout: 

```
  {% with hide_nav_cta=True %}
    {% include 'includes/navigation/navigation.html' %}
  {% endwith %}
```

In Luciano's code, when that's triggered, the layout doesn't shift, in an identical way to protocol. So if we want to keep the logic consistent to the current site, I think we're good. But we can revise with the is-firefox class if you want.
```
.fl-header:has(.hide-cta-on-desktop),
.fl-header:has(.hide-nav-download-on-desktop) {
   padding-right: 156px;
}
```

PR #797, opened 2025-11-20
WT-432: Flare footer

lucianoratamero:
## One-line summary
This PR reimplements the footer using flare's styles and theming.

- mobile support
- compatibility with the WNP template

## To test
* **Create a new Free form page .** _This accepts the same components of a What's new page, but uses the default base template. That's what we're testing here._
    * Go to the pages list, and drill into the root/homepage. Likely at http://localhost:8000/cms-admin/pages/1/
    * hit the green plus button, then create a " Free form page"
    * Add an intro component, and fill out just enough info to allow you to publish the page. 
    * Add a translation or two. (Just like on the existing site, a page will only receive the translation chooser if there are available translations for that URL.)  
    Screenshot 2025-12-09 at 6 53 35 PM

    * Hit the "View Live" button. 

## Notes
* This template isn't used anywhere on production yet, so if you're looking in dark mode, the fact that the body uses black-black and the footer uses a lighter black doesn't matter. This implementation is in service of getting pages like https://www.firefox.com/en-US/features/block-fingerprinting/ in the CMS, so just compare our version with that page on production. Also note that we will be disabling dark mode on those new pages, in line with the existing site's support. So both versions are identical, even though it looks a little off with a dark mode page right now. We can cross that bridge later. 
* We've discussed some CSS refactors. Many of those are finished, but in a future PR, and pointing at the 2026 design only. (It doesn't make sense to mix in the new breakpoints in this version, for example.) So, ignore units and media queries for now. 

## Optional Test #2: 
The What's New Page footer is different than this one, but it's unchanged. If you'd like to compare them, do the same thing as above, but instead of a freeform page, create a "What's new Index", and then add a child page. That should match the footer here: https://www.firefox.com/en-US/whatsnew/145/

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-432

## Screenshots
Screenshot 2025-12-09 at 7 05 39 PM

Screenshot 2025-12-09 at 7 05 09 PM

### DIFF
--- a/media/css/cms/flare-footer.css
+++ b/media/css/cms/flare-footer.css
@@ -210,7 +210,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 /* Footer Headings */
 
 .fl-footer-heading {
-    text-align: left;
+    text-align: start;
     color: var(--token-neutrals-grey-3);
     font-weight: bold;
     margin-bottom: calc(var(--token-scale-8) * 1px);
@@ -238,7 +238,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 /* Subheadings */
 .fl-footer-subheading {
     color: var(--token-neutrals-grey-3);
-    text-align: left;
+    text-align: start;
     font-weight: bold;
     font-size: calc(var(--token-scale-16) * 1px);
     line-height: 1.5;


media/css/cms/flare-footer.css
```diff
@@ -0,0 +1,626 @@
...
+        margin: 0;
+    }
+}
+
+/* -------------------------------------------------------------------------- */
+
+/* Secondary Section */
+
+.fl-footer-secondary {
+    max-width: 1440px;
```
localjo: I think we can use a variable from flare-theme.css or flare-tokens.css here.

kkellydesign (commented):
Just noting to make sure that we're toggling the correct versions of the templates on and off with the environment variable. If it's a CSS-only switch, great. Here's Jo's example: 

https://github.com/mozmeao/springfield/commit/41a3f39032ae5aa550705a454c77f21b792c2145

kkellydesign (approved):
lgtm, I made a couple of minor changes.

springfield/cms/templates/cms/includes/wnp-footer.html
```diff
@@ -38,7 +38,9 @@
           </li>
         </ul>
       </nav>
-      <div class="fl-footer-logo" aria-hidden="true"></div>
+      <div class="fl-footer-logo" aria-hidden>
```
stephaniehobson:
```suggestion
      <div class="fl-footer-logo">
```

This made more sense when the logo was not a link. We should not hide this new functionality from users.

media/css/cms/flare-footer.css
```diff
@@ -0,0 +1,693 @@
...
+.fl-footer-branding p {
+    margin: 0;
+    font-size: calc(var(--token-scale-16) * 1px);
+    color: var(--token-neutrals-white);
+    line-height: 1.5;
+}
+
+.whats-new-page .fl-footer-branding p {
+    color: var(--fl-theme-color-text-body);
+}
```
stephaniehobson: I don't see the class `.fl-footer-branding` `.fl-footer-more` referenced by any HTML?

stephaniehobson (approved):
R+ functionality matches prod in LTR languages.

I left some optional suggestions for code clean up.

I suggest that RTL testing and updates happen for both the footer and nav in a follow up PR. I forgot to review it when I did the nav review. The biggest change needed for that is the switch to `padding-inline` and `margin-inline`.

media/css/cms/flare-footer.css
```diff
@@ -0,0 +1,693 @@
...
+.fl-footer-list li {
+    list-style-type: none;
+    margin-bottom: calc(var(--token-scale-8) * 1px);
+}
+
+.fl-footer-list li:last-child {
+    margin-bottom: 0;
+}
+
+.fl-footer-list.is-closed[aria-hidden='true'] {
```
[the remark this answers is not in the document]
kkellydesign: agreed, I just checked. Removed 👍

PR #804, opened 2025-11-24
Add CMS video analytics

localjo:
## One-line summary

Add event listeners to track CMS video metrics

 instead of adding the `data_videoengagement` bundle to the page, this imports `../base/datalayer-videoengagement.es6` in the `firefox-cms-new` bundle, and adds the event listeners to video elements when the user clicks on the poster image and the `video` element is dynamically added to the page. This solution doesn't involve a `ga-video-engagement` class.
 
I'm also not sure if this makes the `data_videoengagement` bundle redundant or not. I don't see another place that bundle is used.

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/pull/781

### DIFF
--- a/media/js/cms/new.es6.js
+++ b/media/js/cms/new.es6.js
@@ -4,6 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+import VideoEngagement from '../base/datalayer-videoengagement.es6';
+
 (function () {
     function initNewsletterForm() {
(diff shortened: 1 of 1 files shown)


maribedran (approved):

maribedran:
The event handlers were called when I interacted with the video. I'm assuming the video engagement is doing its work, and GA is getting what it needs. If @stephaniehobson doesn't have anything, we can merge it to the banners work or move #781 first and get this in afterwards.

PR #807, opened 2025-11-28
Wt 474 homepage

maribedran:

### DIFF
--- a/media/css/cms/flare26-tokens.css
+++ b/media/css/cms/flare26-tokens.css
@@ -72,8 +72,7 @@ Static values extracted from Figma that do not change based on context.
     --token-border-radius-xl: 125px;
 
     /* Width and viewport variables */
-    --token-width-desktop-content: 697px;
-    --token-width-max-content: 725px;
+    --token-width-desktop-content: 725px;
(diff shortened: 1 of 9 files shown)


media/css/cms/flare26-tokens.css
```diff
@@ -71,6 +73,7 @@ Static values extracted from Figma that do not change based on context.
 
     /* Width and viewport variables */
     --token-width-desktop-content: 697px;
+    --token-width-max-content: 725px;
```
localjo: I think this is semantically what `--token-width-desktop-content` was intended for.

springfield/cms/templates/cms/blocks/light-dark-image.html
```diff
@@ -5,16 +5,14 @@
  file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #}
 
+{% set sizes = sizes or "(min-width: 768px) 50vw, (min-width: 1440px) 680px, 100vw" %}
```
localjo: Should these sizes match the new tshirt size `@custom-media` breakpoints?

PR #812, opened 2025-12-02
fix: ensure that WNP path can only start with whatsnew/

stevejalim:
This changeset tightens up the regex for our whatsnew pages

Testing

- [x] http://localhost:8000/whatsnew/200/ will be 200 OK
- [x] http://localhost:8000/firefox/whatsnew/200/ will be 404
- [x] http://localhost:8000/anything/whatsnew/200/ will be 404


maureenlholland (approved):
r+ 🟢

PR #820, opened 2025-12-03
Pattern library: Add page titles to full page view. 

stephaniehobson:
## One-line summary

Add page titles to full page view of pattern library patterns.

- Created custom `RenderPatternView` that extends the base pattern library view to pass name to the base template context
- Added custom URL configuration to use the extended view instead of the default pattern library render view
- Updated `base-pattern.html` to display name in page title

## Issue / Bugzilla link

n/a

## Testing

Open a full page view from the pattern library e.g. http://localhost:8000/pattern-library/render-pattern/pattern-library/base-styles/typography/typography_examples.html and see if title matches the content.

### DIFF
--- a/springfield/cms/pattern_library_urls.py
+++ b/springfield/cms/pattern_library_urls.py
@@ -1,0 +1,28 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+from django.urls import path, re_path
+
+from pattern_library import get_pattern_template_suffix, views
+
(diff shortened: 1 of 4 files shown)


kkellydesign (changes requested):
This works well for me, but rather than put it here in Springfield, lets put it in the Pattern Library code itself. (we're already running a LincolnLoop fork of the project) 

@dchukhin can you take a look at this and see if it's a good candidate for a global revision to the pattern library?

dchukhin:
Yeah, that seemed like a good idea, so I moved the logic there, made a new release, and opened a pull request here to use the newest version of our django-pattern-library fork: https://github.com/mozmeao/springfield/pull/826.
I think that pull request would supercede this one.

PR #844, opened 2025-12-08
Youtube video loader

maribedran:
## One-line summary

Load YouTube iframes using the YouTube IFrame API to allow GTM to trigger events on video interaction.

This PR is an attempt to guarantee that the appropriate GTM events are fired on YT videos. It adapts the code from `media/js/base/video-inline-embed.es6.js` that loads the IFrame API and calls it to replace the placeholder image with the video iframe.

We still couldn't verify that the code works, but it seems to be a GTM configuration issue. According to GTM's documentation, loading the video through the API should be enough: https://support.google.com/tagmanager/answer/7679325?hl=en

## Issue / Bugzilla link

Interacting with the video on this page doesn't trigger the GTM events: https://www.firefox.com/en-US/whatsnew/146/

## Testing

The `dataLayer` should have video interaction events like observed here: https://www.mozilla.org/en-US/firefox/136.0/whatsnew/

### DIFF
--- a/media/js/cms/new.es6.js
+++ b/media/js/cms/new.es6.js
@@ -221,15 +221,62 @@ import VideoEngagement from '../base/datalayer-videoengagement.es6';
         });
     }
 
-    function loadScript() {
+    function loadScript(button) {
         const tag = document.createElement('script');
         tag.src = 'https://www.youtube.com/iframe_api';
         const firstScriptTag = document.getElementsByTagName('script')[0];
         firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+
+        window.onYouTubeIframeAPIReady = function () {
+            // Play the video only once the API is ready.
+            playVideo(button);
+        };
+    }
+
+    function isScriptLoaded() {
+        return document.querySelector(
+            'script[src="https://www.youtube.com/iframe_api"]'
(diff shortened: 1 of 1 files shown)


stevejalim (commented):
Thanks! This technically works, but I _don't_ think we should ship this extracted-from-Bedrock implementation as-is, rather than the current live implementation.

1) This implementation ends up loading data from youtube.com, not youtube-nocookie.com, which has a privacy aspect that we should discuss more. The current live implementation uses youtube-nocookie.com and doesn't set cookies. We need to ensure we always put privacy first, so I'm prepared to block on this point until we discuss. It might be that the deliberate action of triggering a video is enough before loading cookies, but I don't want to assume in a rush

2) With this implemenation, the post-play behaviour is to show a suggested next watch, which is not a great look - especially when the suggestion is not from our own channel. At the least we should scope the suggestion to our own channel via the rel=0 param on the embed URL (https://developers.google.com/youtube/player_parameters#rel). Note that the current live implemenation does not appear to show next–watch suggestions

3) It sounds like we're not getting analytics from other things on Springfield, which suggests misconfig outside of this codebase

4) Based on what I've gleaned from Slack, we have a workaround to get the engagement data we need

PR #853, opened 2025-12-10
WT-508 CMS image optimization

maribedran:
## One-line summary

This PR fixes the rendered image tags on all CMS blocks, making sure that the height attribute is set, and changes the default image rendition format to WEBP.

- Pass `height` attr to all `srcset_image()` calls on CMS templates
- Add the `WAGTAILIMAGES_FORMAT_CONVERSIONS` setting to convert all image renditions to WEBP

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-508

## Testing

Run `./manage.py wagtail_update_image_renditions` and restart the server.

- Check that all CMS pages have images using the WEBP format
- Check that the image tags have the `width` and `height` attributes 

image

### DIFF
--- a/springfield/settings/base.py
+++ b/springfield/settings/base.py
@@ -1407,6 +1407,12 @@ WAGTAILIMAGES_EXTENSIONS = [
     "svg",
 ]
 
+WAGTAILIMAGES_FORMAT_CONVERSIONS = {
+    "png": "webp",
+    "jpg": "webp",
+    "jpeg": "webp",
+    "webp": "webp",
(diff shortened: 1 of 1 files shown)


maribedran:
@stevejalim When I ran the management command locally, I had to restart the server for the pages to load the images. We'll need to test this carefully if we want to regenerate the renditions in production. If we don't, it's also not a big deal: the existing images will remain PNGs, and only the new ones will be served as WEBP.

PR #856, opened 2025-12-11
2026 - 5 - Carousel progress indicator CSS

kkellydesign:
## One-line summary
Carousel: progress indicators added. Colors revised.

### DIFF
--- a/media/css/cms/flare26-kit-banner.css
+++ b/media/css/cms/flare26-kit-banner.css
@@ -51,7 +51,7 @@
 }
 
 .fl-banner-kit .fl-banner-content .fl-subheading {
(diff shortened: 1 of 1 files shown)


localjo:
I added `embla-carousel` to Springfield and updated the carousel component to use it. The carousel now auto-plays on a 4-second timer, animates the label ring via `--carousel-progress`, and fades smoothly between slides. I also accounted for `prefers-reduced-motion` by disabling autoplay/progress for users who prefer less animation, while still allowing slide changes via label clicks. Finally, I fixed an existing issue where the label numbers were being incremented twice, so they now count up correctly by one.

PR #869, opened 2025-12-15
2026 - 0.1 - Templates organization

maribedran:
## One-line summary

This PR changes the base templates for some pages to make sure that they only display fully implemented components.

- Free Form and What's New pages will always use the 2025 Flare template
  - They should get a feature flag once the components are all migrated to the 2026 design and the pages can safely use the new styles
- Article pages should get the feature flag once that branch is merged
- The Homepage already uses the 2026 template, so no change
- The pattern library will use either the Flare 2025 or 2026 template, depending on the component
  - Each component implemented on the 2026 design should get an entry on the pattern library with "2026" on its name

### DIFF
--- a/media/css/cms/flare-defaults.css
+++ b/media/css/cms/flare-defaults.css
@@ -25,7 +25,6 @@ body {
     color: var(--fl-theme-color-text-primary);
     background: var(--fl-theme-surface-page);
     width: 100%;
-    max-width: var(--token-widths-viewport-wrapper-screen-2xl);
 }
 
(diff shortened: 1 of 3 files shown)


kkellydesign (approved):
This looks good. The aim here is to show both sets of CSS in the pattern library (for as long as there are two sets, temporarily. We'll refactor this as we go.)

Screenshot 2025-12-18 at 3 38 07 PM
Screenshot 2025-12-18 at 3 37 58 PM

Issue #846, opened 2025-12-09
Ensure YouTube-hosted videos send expected GTM events


maureenlholland:
## Description

Reported from slack: https://mozilla.slack.com/archives/C019N6J8FGA/p1765215322919649 

The Bedrock functionality was ported to Springfield here: https://github.com/mozmeao/springfield/pull/674
And there was a corresponding update to `www.firefox.com` tag manager configuration

For correct implementation, see events fired here*:
https://www.mozilla.org/en-US/firefox/136.0/whatsnew/

*You will need to be in a non-EU geo, no DNT or GPC preference, no decline cookies pref

In console, you can check `dataLayer`, but it is best to log into Tag Manager account and use [Tag Assistant](https://chromewebstore.google.com/detail/tag-assistant/kejbdjndbnbjgmefkgdddjlbokphdefk?hl=en) extension (Chrome only) 

Use demo to test CMS (outlined banner with video): www-demo1.springfield.moz.works/en-US/tests-index-page/test-banner-page/

---

#### Success Criteria
- [ ] Static page YouTube videos send events, i.e. https://www.firefox.com/en-US/features/picture-in-picture/
- [ ] CMS YouTube videos send events

maureenlholland:
Confirmed events fire as expected locally on Springfield, and events continue to fire with `nocookie` domain on Bedrock

Need to figure out why the events fire on Springfield local but not on Springfield prod

PR #873, opened 2025-12-16
Set YouTube to nocookie host

maureenlholland:
## One-line summary

Update video inline embed to use `no-cookie` URL like CMS embed does

NOTE: this could be closed instead if it no longer useful!

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/issues/846

## Testing
http://localhost:8000/sco/features/picture-in-picture/ (non-CMS video page)

Screenshot 2025-12-16 at 3 51 56 PM

You can compare with events that fire in production for the same video on `www.mozilla.org`: https://www.mozilla.org/fr/firefox/136.0/whatsnew/

### DIFF
--- a/media/js/base/video-inline-embed.es6.js
+++ b/media/js/base/video-inline-embed.es6.js
@@ -35,6 +35,7 @@ function playVideo() {
     }
 
     new window.YT.Player(videoLink, {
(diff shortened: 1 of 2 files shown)


maureenlholland:
This could be closed as outdated since CMS development has moved on

PR #897, opened 2026-01-08
Add Linux ARM64 button

maureenlholland:
## One-line summary
Update Linux download button group to show ARM64 option

Moved some standalone buttons to use the linux macro helper
- any reason we should revert to the standalones?

Updated `download_button` logic to show both linux options when OS is linux
- this only really makes sense to use with direct downloads, as it creates an extra click to go from a specific download button to /thanks page with the same specific download buttons
- updated release note buttons to use direct downloads

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/issues/648

## Testing
I've used `general.useragent.override` set to `Mozilla/5.0 (X11; Linux x86_64; rv:10.0) Gecko/20100101 Firefox/10.0` in about:config settings (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox#linux)

http://localhost:8000/en-US/channel/desktop/
http://localhost:8000/en-US/thanks/
http://localhost:8000/en-US/thanks/?xv=basic

bottom section CTAs
release (/thanks link): http://localhost:8000/en-US/firefox/148.0beta/releasenotes/
beta (direct download links): http://localhost:8000/en-US/firefox/147.0/releasenotes/
esr (3 buttons, direct download): http://localhost:8000/en-US/firefox/128.12.0/releasenotes/

Available for testing on Demo 5: https://www-demo5.springfield.moz.works/en-US/browsers/desktop/linux/

### DIFF
--- a/springfield/firefox/templates/firefox/releases/release-notes.html
+++ b/springfield/firefox/templates/firefox/releases/release-notes.html
@@ -49,21 +49,11 @@
 {% elif release.product == 'Firefox' %}
   {% set product_name = 'Firefox' %}
   {% set primary_heading = 'See what’s new in Firefox!' if release.is_latest else 'Firefox Release Notes' %}
-  <!-- TODO: remove condition once we refactor /all to simplify Linux download path -->
-  {% if release.channel == 'Release'%}
-    {% set download_button_primary %}
-      {{ download_firefox_thanks(dom_id='download-release-primary') }}
-    {% endset %}
-    {% set download_button_secondary %}
-      {{ download_firefox_thanks(dom_id='download-release-secondary') }}
-    {% endset %}
-  {% else %}
-    {% set download_button_primary %}
-      {{ download_firefox(dom_id='download-release-primary') }}
-    {% endset %}
-    {% set download_button_secondary %}
-      {{ download_firefox(dom_id='download-release-secondary') }}
-    {% endset %}
-  {% endif %}
+  {% set download_button_primary %}
+    {{ download_firefox(dom_id='download-release-primary') }}
+  {% endset %}
+  {% set download_button_secondary %}
+    {{ download_firefox(dom_id='download-release-secondary') }}
+  {% endset %}
   {% set download_all_link = firefox_url('desktop', 'all', channel='release') %}
 {% endif %}


springfield/firefox/templates/firefox/releases/release-notes.html
```diff
@@ -50,10 +50,10 @@
   {% set product_name = 'Firefox' %}
   {% set primary_heading = 'See what’s new in Firefox!' if release.is_latest else 'Firefox Release Notes' %}
   {% set download_button_primary %}
-    {{ download_firefox(dom_id='download-release-primary') }}
+    {{ download_firefox(dom_id='download-release-primary', force_direct=True) }}
```
maureenlholland: should I restrict this behaviour to only force direct if os is linux?

janbrasna:
> "this only really makes sense to use with direct downloads, as it creates an extra click to go from a specific download button to /thanks page with the same specific download buttons"

This is fun. How come it never occurred to me before when testing the flows:/

Question is, the choice needs to stay in one place. What if that place is the /thanks scene only? So all the Linux buttons around the site, if they point to release, not other channels, probably should not break down the platforms or link directly — only /thanks should. (The /linux platform page is the only exception where the direct arch links are hard-coded I guess.)

The arch choices in the download macro only ever make sense for non-release channels, and their e.g. release notes CTAs — which are the ones that do not go through the /thanks page.

springfield/firefox/templates/firefox/releases/release-notes.html
janbrasna: This currently links /thanks for release, and direct for other channels.

While on Linux the label says "for Linux 64-bit" the reality currently is that one still points to /thanks listing all the options anyways — so anything is an improvement over the current state, yeah. (I'd say for Release Notes stable channel, the text of the CTA ought to change, not its href destination — mainly for not showing the APT/RPM instructions there, only at the /thanks page…)

Rendering the repository note with the combo CTAs on the actual page could also be an option, but, TBH, I'm not sure if that only applies to release, or all channels in general? 🤷 

(I don't know enough about attribution, but expecting this to not have an impact, esp. on other platforms, where I believe all these CTAs get the hash added to the bouncer link as well, so the utility of the /thanks scene used to be mainly for Windows as it had some more logic that I believe is no longer there [pinning to taskbar, setting as default etc.] so for now I think there's just the tendency to get rid of any custom logic and route things trough /thanks just for consistency mainly?)

maureenlholland:
> ESR Release Notes page:

oof, yeah, that needs to be handled. will apply alignment fix for this case, as suggested, thanks @janbrasna

maureenlholland:
TODO:
- fix alignment on 3 buttons for rare ESR case
- update direct download vs thanks flow as listed in https://github.com/mozmeao/springfield/pull/897#issuecomment-3744778716 OR direct to thanks if special cases become too complicated

stephaniehobson:
Follow up questions (I'm sorry I'm having trouble figuring out the desired state by reading the thread):

- ~Should the 32bit version be an option for release (download/linux and /thanks)?~
- Should we hide 32bit for non release channels? Or is that only showing because of the lagging product details?
- ~The ARM button isn't translated, I assume that's a lagging translation because I'm seeing English not a string ID but just double checking.~ _Edit: confirmed by looking at code._
- Can we do anything about the wrapping download button?

The real solution here is to redesign /all and then link to a pre-filled Linux download page allowing the user to pick exactly what they want including APT instructions. Until then... Do we add a 4th button with the APT/SUMO link 😵‍💫?

springfield/firefox/templates/firefox/releases/release-notes.html
```diff
@@ -49,11 +49,20 @@
 {% elif release.product == 'Firefox' %}
   {% set product_name = 'Firefox' %}
   {% set primary_heading = 'See what’s new in Firefox!' if release.is_latest else 'Firefox Release Notes' %}
-  {% set download_button_primary %}
-    {{ download_firefox(dom_id='download-release-primary') }}
-  {% endset %}
-  {% set download_button_secondary %}
-    {{ download_firefox(dom_id='download-release-secondary') }}
-  {% endset %}
+  {% if release.channel == 'Release'%}
```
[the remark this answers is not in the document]
stephaniehobson: Maybe add a comment about why it is this way?

maureenlholland:
@stephaniehobson yeah, it's a rabbithole

> Should we hide 32bit for non release channels? Or is that only showing because of the lagging product details?

For demo, I believe this is only due to lagging product details. Locally, it should not show the three buttons [edit: Two are expected everywhere with the exception of ESR, which still supports 32-bit]

> Can we do anything about the wrapping download button?

Yep, that's a pretty straightforward style fix I think it's safe to bundle into this PR

> Until then... Do we add a 4th button with the APT/SUMO link 😵‍💫?

I'd say no. I think in terms of iterative improvements, we can continue to leave anywhere that does not currently have an APT/SUMO link without an APT/SUMO link. It feels like an improvement to have the ARM64 option added for now.

PR #908, opened 2026-01-19
Add noindex to more pages - WT116 (#370)

wen-2018:
## One-line summary

This PR add `noindex` to more pages/templates.

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/370
[WT-116](https://mozilla-hub.atlassian.net/browse/WT-116?atlOrigin=eyJpIjoiOTY4OWYyNGFlOWI1NDNlMmE3N2UzYWM0NWIwMzQyZGQiLCJwIjoiaiJ9)

### DIFF
--- a/springfield/base/templates/404.html
+++ b/springfield/base/templates/404.html
@@ -14,9 +14,6 @@
   {{ css_bundle('page_not_found') }}
 {% endblock %}
 
-{# 404 pages must not include canonical and hreflang x-default: https://github.com/mozilla/bedrock/issues/5895 #}
-{% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
-
 {% block content %}
   <main class="mzp-l-content message-section">
     <h1 class="error-title">{{ ftl('not-found-page-sorry-we-cant-find-that-page') }}</h1>


springfield/base/templates/base-error.html
```diff
@@ -5,6 +5,7 @@
 #}
 
 {% extends 'base-protocol.html' %}
+{% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
```
janbrasna: if moved here, would it make sense for maintainability to remove the same from the child 404 https://github.com/mozmeao/springfield/pull/771#pullrequestreview-3452889319 that inherits it?

(It doesn't usually make sense to noindex error pages as they can't get indexed per se, it's usually only taken as a fallback for cases when such template could erroneously be served with wrong 2xx/3xx code, not 4xx/5xx, so it's kind of a safety net really … there's no harm adding it, canonicals are correctly empty in both cases etc. … the only concern in adding these is getting out of sync w/ bedrock — maybe porting it there as well later for consistency?)

wen-2018: Updated.

PR #952, opened 2026-02-03
Programmatically Create Firefox Features Pages

dchukhin:
## One-line summary
This pull request creates the firefox feature pages (https://www.firefox.com/en-US/features/) in Wagtail, with translations in each of the languages from the [Top Firefox Website Languages document](https://docs.google.com/spreadsheets/d/1ym4DHb8SlcGhx8040srMiqyi7-kuZA3xoyRufCxbEoo/edit?gid=1793559213#gid=1793559213).

This pull request:
 - adds an `index_page_heading` field to the `ArticleDetailPage`, since the index page does not always show the `ArticleDetailPage`'s title
 - adds a `VideoBlock` to the `ArticleDetailPage.content` field, since 1 feature page has a video block
 - adds feature pages to the fixtures directory, and loads them when loading fixtures
 - adds a migration (`springfield/cms/migrations/0038_create_feature_pages.py`) that programmatically creates each of the firefox features pages with content that matches the currently-live content.
 - adds some FTL parsing utility functions

## Testing
1. run the migrations from this pull request with `ALL_LOCALES=True` (`ALL_LOCALES=True python manage.py migrate cms`)
2. navigate to the feature index page in Wagtail, and its children, in each of the expected languages
3. compare the pages created in this pull request to the live pages (https://www.firefox.com/en-US/features/)
4. note that all of these feature pages should be in draft state. http://localhost:8000/en-US/features/ should show the old (non-Wagtail) pages

### Note the following shortcomings:
 - we should improve the styling of the video block (see the "Get more done with pop-out videos " article detail page: http://localhost:8000/en-US/features/picture-in-picture/)
 - es-MX (listed in https://docs.google.com/spreadsheets/d/1ym4DHb8SlcGhx8040srMiqyi7-kuZA3xoyRufCxbEoo/edit?gid=0#gid=0 as a primary (fully-supported) language) doesn't have enough translations right now. I'll leave it as is for now, but without a better fallback system implemeted some sections will appear in English. 
 - The "Edit PDFs for free with Firefox PDF Editor" page doesn't have some translations in es-ES. This is also true on the live page (https://www.firefox.com/es-ES/features/pdf-editor/). See the following sections:
   - "Add drawings"
   - "Create a signature"

### DIFF
--- a/springfield/cms/migrations/0039_create_feature_pages.py
+++ b/springfield/cms/migrations/0039_create_feature_pages.py
@@ -101,6 +101,29 @@ def create_translation_sources(apps, schema_editor):
             source.refresh_segments()
             print(f"  Refreshed segments for: {slug}")
 
+    # Fix schema_version on all TranslationSource records we just created.
+    # During this migration, get_schema_version("cms") returns the previous
+    # migration because the current one hasn't been recorded as applied yet.
+    # We determine the correct value from the migration files on disk so it
+    # matches whatever get_schema_version("cms") will return after all
+    # migrations have been applied.
+    from pathlib import Path
+
+    migrations_dir = Path(__file__).parent
+    migration_names = sorted(f.stem for f in migrations_dir.glob("0*.py"))
+    latest_schema_version = migration_names[-1] if migration_names else ""
+
+    TranslationSource.objects.filter(
+        object_id__in=[index_page.translation_key]
+        + list(
+            ArticleDetailPage.objects.filter(
+                slug__in=PAGE_FTL_MAPPING.keys(),
+                locale=source_locale,
+                path__contains=index_page.path,
+            ).values_list("translation_key", flat=True)
+        ),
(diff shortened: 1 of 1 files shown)


kkellydesign (changes requested):
> Currently, the pages get published when the migration runs. Should we leave them in draft instead?

Yes, draft please. We'll need to test these before they go live.

I'll note the other shortcomings when I pass off the URLs, but as long as they get created as drafts, none of these are blockers.

I'll also fix the video styles before this gets merged.
[1:36 PM]Also, please merge in main and fix migrations. I've been busy on main today

stevejalim (approved):
Eyeballs review: r+

I appreciate you testing this @kkellydesign 

I'll get it onto main now and we'll see all languages being loaded on www-dev because DEBUG=False there (as well as on Stage and Prod). Will run DB backups before that, too

Issue #959, opened 2026-02-08
Missing attribution in the last /download/all step if loaded async


janbrasna:
## Description

/download/all final download button is missing the expected attribution hash if not accessed directly by its permalink, but using the JS step partial logic from the initial page.

### Steps to reproduce

You need to access from a region that doesn't disable tracking stuff, have no DNT et al., also might wanna come from search or a redirect to have some tracking params to start with. (Win & Mac only)

If accessed directly, the final step e.g.: https://www.firefox.com/en-US/download/all/desktop-release/win/br/ should decorate the CTA button with attribution hash.

### Expected result

Now reaching the same page via in-page JS partial logic starting from https://www.firefox.com/en-US/download/all/ (Release, Win32, Breton) should give you attribution on the same button.

### Actual result

Given the hashing runs on domready, it doesn't "watch" for newly added elements or dynamic content, so when the button appears via XHR partials later, the attribution is not added to it at that point:

https://github.com/user-attachments/assets/fdbe4745-9f75-4ed8-bcd0-3048584be78f

Video shows the link destination in status bar upon hover.
1. no attribution when clicking through the steps
2. comparing with another direct download CTA to verify attribution is getting added
3. using the last step permalink to open directly, demonstrating how it should have been added in step 1. too…

### Environment

_(Using remote NA Windows 10 & Edge 138 to be able to reproduce from within EEA.)_

PR #961, opened 2026-02-10
/all fixes (fixes #946, #959)

### LINKED ISSUE #946: Download all scrolled too far in the last step
## Description

Language pack downloads further down in the alphabet do not have their download button appear in the view, showing footer instead.

### Steps to reproduce

Go to https://www.firefox.com/download/all/desktop-release/osx/ and select a language "below the fold", so that you have to scroll before clicking it, e.g. Portuguese or Swedish.

### Expected result

The focus for the last download step should enforce either the heading or the CTA in the view.

### Actual result

The scroll position stays, while the new content gets shorter, effectively making the scroll position point at the footer location:

https://github.com/user-attachments/assets/69404a85-0580-4ce2-8d80-0b8262e35a42

### Environment

Gecko, WebKit

maureenlholland:
## One-line summary

Fixes to scroll focus and partial HTML stub attribution

[This flow is scheduled for a refactor](https://github.com/mozmeao/springfield/issues/258), so the goal is to fix the issues and make clear what cases we are covering

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/946
https://github.com/mozmeao/springfield/issues/959

## Testing

Demo: https://www-demo3.springfield.moz.works/en-US/download/all/

For scroll focus
Go to http://localhost:8000/en-US/download/all/desktop-beta/win64/
Scroll down on the language options and select language.
This will leave you on the footer in prod. PR should move up to download section. 
- NOTE: it seems the lesser of two evils to keep the scroll even if it does cause some slight jumping on the earlier selections. It was not too obnoxious to me, but could use more opinions: https://github.com/mozmeao/springfield/pull/774/changes#r2756640243

For partial HTML stub attribution 
Ensure you are on [a browser that allows cookies](https://github.com/mozmeao/springfield/pull/774/changes#r2756640243) (stub attribution requires cookies)
Start on http://localhost:8000/en-US/download/all and go through the desktop flow to download links. You should see stub attribution params on the download link in the last step (or in dev tools > storage/application > cookies)

### DIFF
--- a/media/js/firefox/all/all-init.es6.js
+++ b/media/js/firefox/all/all-init.es6.js
@@ -79,9 +79,11 @@ import MzpModal from '@mozilla-protocol/core/protocol/js/modal';
 
         // init stub attribution & event tracking for GA4
         if (downloadButtons && downloadButtons.length > 0) {
-            // We cannot rely on DOM ready as this section is partially fetched
-            // This flow is scheduled for refactoring: https://github.com/mozmeao/springfield/issues/258
-            StubAttributionConsent.init();
+            if (StubAttributionConsent) {
+                // We cannot rely on DOM ready as this section is partially fetched
+                // This flow is scheduled for refactoring: https://github.com/mozmeao/springfield/issues/258
+                StubAttributionConsent.init();
+            }
 
             for (let i = 0; i < downloadButtons.length; ++i) {
                 const downloadButton = downloadButtons[i];


maureenlholland:
pushing to Demo 3

maureenlholland:
> I'd imagined { block: "center" } would be most useful but the results are equally underwhelming.

@janbrasna thanks for the note on scrollIntoView. `{ block: 'center' }` is looking good to me across screen sizes and modern browsers. Is there a particular combination that made it underwhelming?

maureenlholland:
@janbrasna awesome! Thanks again for sharing the option!

PR #963, opened 2026-02-10
Add default browser checkbox to download button block

maribedran:
## One-line summary

Add a new option to the Download Firefox Button block to display a "Set as default browser" checkbox.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-643

## Testing

- Check http://localhost:8000/en-US/tests-index-page/test-buttons-page/ for the download buttons
  - The checkbox should only show on Windows
  - The page's preview should display the checkbox with the preview indicator
- Check http://localhost:8000/en-US/tests-index-page-2026/test-home-page/ for the same conditions
- Open the pages in a Windows environment
  - If the button isn't configured to display the checkbox nothing should change with the default stub attribution
  - If the checkbox is set, the `?utm_campaign=SET_DEFAULT_BROWSER` query should be added to the URL image
  - Check the cookies and look for `moz-stub-attribution-code`
  - Decode the base64 value from the cookie. It should look like: `source=(not set)&medium=(direct)&campaign=SET_DEFAULT_BROWSER&content=(not set)&experiment=(not set)&variation=(not set)&ua=chrome&client_id_ga4=491504120.1770734695&session_id=3307072354&dlsource=fxdotcom` (note the `SET_DEFAULT_BROWSER` campaign) image
  - Click the download button and check the "Try downloading again" link for the same attribution values image
  - Uncheck the checkbox and verify that the stub attribution value has changed to the default properties. Something like:  `source=(not set)&medium=(direct)&campaign=(not set)&content=(not set)&experiment=(not set)&variation=(not set)&ua=chrome&client_id_ga4=142934101.1770905600&session_id=8364667907&dlsource=fxdotcom` (no SET_DEFAULT_BROWSER` campaign)


maureenlholland:
confirm Try Again is good on /thanks page

maureenlholland (changes requested):
one blocker around JS running when Wagtail checkbox setting is false

I guess the other difficult thing is this will affect all download buttons on the page, regardless of whether they have a checkbox. Maybe another point in help text?

the UTM param is added to URL and stub attribution cookie is updated for SET_DEFAULT_BROWSER, regardless of which button I press
Image

PR #990, opened 2026-02-17
Remove smoke tests related to modals ahead of flare26's launch 

lucianoratamero:
## Testing

Run `pytest --base-url http://localhost:8000 --driver Firefox --html tests/functional/results.html tests/functional/firefox` locally and verify the tests pass.


janbrasna (approved):
Yup [this is green](https://github.com/mozmeao/springfield/actions/runs/22118539737/job/63932791617) ✅ 

I'd say it's alright at this point to remove this very specific check. New layout will linearize the content similarly to current nonJS fashion.

PR #1001, opened 2026-02-18
Add CMS refresh redirects for enterprise, developer, and mobile pages

stevejalim:
These are gated with the other new redirects behind the `ENABLE_CMS_REFRESH_REDIRECTS` env var, which we're lined up to flip when we need it next week


janbrasna (commented):
You probably want to zap the vanity redirects that are currently sitting on the paths in advance as well:

PR #1008, opened 2026-02-19
Put back all nav icons

janbrasna:
## One-line summary

Partly reverts #999 until the assets are removed from their classes where defined.

Unused ones need to be removed from the icon style sheet first; there might be others still used somewhere, that I'm reluctant to wipe, so I'm putting all back instead — only to trim that properly later when the timing's better.

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/actions/runs/22167269255

## Testing

Trying my CI update to check _collectstatic_ right on PR: [`@janbrasna/springfield`/actions/runs/22175014224](https://github.com/janbrasna/springfield/actions/runs/22175014224/job/64121367636#step:3:7673) ✅


stevejalim (approved):
:batman: r+

PR #1012, opened 2026-02-19
Add parallel PR CI including collectstatic check

janbrasna:
## One-line summary

To evaluate a more comprehensive CI _make_ target I figured adding that in parallel to the current pipeline could help assess if the extra 40–60sec spent on it is worth it.

This will now run both `make test-image` and `make test-ci` side by side. It's a waste of CPU cycles, but is actually great to see the difference over the span of a week or two to see the range of durations this tends to have (between 6–8mins usually), when triggered on the same event at the same time.

(The difference between the two make targets is covered in the linked tickets. TL;DR is: the added strict _collectstatic_ check is implicitly achieved by running a target including `release` service in its build.)

The idea is to move the original `make test-image` job to use `make test-ci` eventually if this proves useful.

(NB: For preview purposes, this duplicate pipeline is only run on PRs, not again on main, as it's just waste of cycles at this point, the same will be done in CD pipeline.)

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/16495
https://github.com/mozilla/bedrock/issues/16918

## Testing

> CI

This actual PR already shows the modified CI run:
[/actions/runs/22188873820](https://github.com/mozmeao/springfield/actions/runs/22188873820) ✅✅✅

> checks

(NB: These start at the same time, so the result doesn't come 8mins later, but just less than a minute later — it only adds 40–60s to the overall CI span running in parallel, not its whole run time.)

Example of missing asset caught:
[`@janbrasna/springfield`/actions/runs/22175415059](https://github.com/janbrasna/springfield/actions/runs/22175415059/job/64122745301#step:3:7714) ❌ 

> fail

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -52,3 +52,13 @@ jobs:
           files: python_coverage/coverage.xml
           fail_ci_if_error: false  # optional (default = false)
           verbose: true  # optional (default = false)
+
+  test-python-more:  # This is a preview `test-ci` to replace the previous `test-image` over time
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request' }}  # Only run the extra release checks on PRs for now
+    steps:
+      - uses: actions/checkout@v6
+      - name: "Run Python tests (on Docker) with extra release service build"
+        run: |
+            make clean test-ci
+        timeout-minutes: 60


stevejalim (approved):

.github/workflows/unit_tests.yml
```diff
@@ -52,3 +52,13 @@ jobs:
...
+
+  test-python-more:  # This is a preview `test-ci` to replace the previous `test-image` over time
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request' }}  # Only run the extra release checks on PRs for now
+    steps:
+      - uses: actions/checkout@v6
+      - name: "Run Python tests (on Docker) with extra release service build"
+        run: |
+            make clean test-ci
+        timeout-minutes: 60
```
janbrasna, line 64: Hmh that wasn't really the smartest name/ID (:) but it was meant to be just temporary to try it out, hence why slightly obnoxious to jump out at people in the checks, to eventually evaluate moving to it completely, or not.

How's its reliability over time?

janbrasna: When I was comparing it, we were in the 6min vs. 8min range… not sure what all was added in the tests over time, but since it's now 20min+, the difference doesn't seem that big. Funnily the random samples I checked showed the longer CI actually taking less time? ;D

| Screenshot 2026-07-27 at 22 46 24 | Screenshot 2026-07-27 at 23 56 36 | Screenshot 2026-07-27 at 22 46 34 |
|--------|--------|--------|

PR #1053, opened 2026-02-24
Ie js support

kkellydesign:
## One-line summary
This fixes JS issues in IE11. This should also fix Selinium tests, but we'll save that for a different PR. 

To test, IE and webpack don't get along. So to run browserstack in local mode: 

* stop the dev server
* run npm run build
* run ./manage.py collectstatic --noinput
* run ./manage.py runserver

* Try the menu, and see that it opens. 
* Open the 'get firefox' link and see that the download starts 
Screenshot 2026-02-24 at 6 26 09 PM

### DIFF
--- a/media/css/cms/flare26-fallback-nav.css
+++ b/media/css/cms/flare26-fallback-nav.css
@@ -77,6 +77,7 @@
     background-color: #fff;
     border: 2px solid #eee;
     display: flex;
+    left: 0;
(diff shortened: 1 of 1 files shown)


stevejalim:
I don't have Browserstack access so pushing to `run-integration-tests` even though it sounds like, from the descripton, that that won't work

PR #1055, opened 2026-02-25
Update Windows support link in download template to avoid redirecting

denyshon:


janbrasna:
Thanks Denys. Would you mind also adding the other instance for consistency?

https://github.com/mozmeao/springfield/blob/1e708caff1ec74144305b1af44f7eea9fff16b82/springfield/firefox/templates/firefox/download/desktop/thanks.html#L74

Appreciate it.

PR #1079, opened 2026-02-26
Hide UI Tour buttons until UITour responds

slightlyoffbeat:
UI Tour buttons were visible but non-functional when UITour was present but not responding (e.g. permissions revoked by clearing site settings). Buttons are now hidden by default via CSS and only revealed after a successful Mozilla.UITour.ping() response. Fade-in transition respects prefers-reduced-motion.

## Testing
Stephen on the websites team is affected and could test.

### DIFF
--- a/media/css/cms/flare26-button.css
+++ b/media/css/cms/flare26-button.css
@@ -241,3 +241,9 @@ I propose we move to "Filled (Primary)", "Outline (Secondary)", and
 .fl-tiny-icon-button:focus {
     color: var(--fl-theme-color-link-hover);
 }
+
+/* UI Tour buttons are hidden by default and revealed via JS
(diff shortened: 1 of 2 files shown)


media/js/cms/ui-tour-buttons.js
```diff
@@ -158,20 +188,10 @@ function init() {
...
-}
-
 if (
     typeof window.Mozilla.Client !== 'undefined' &&
     typeof window.Mozilla.UITour !== 'undefined' &&
     window.Mozilla.Client.isFirefoxDesktop
 ) {
     init();
-} else {
-    hideUITourElements();
```
janbrasna: Yes! this needs to be happening inside init() as that's where the ping() can fail.

PR #1086, opened 2026-02-26
WT-822 Link Pages With Translations After DB Export

dchukhin:
## One-line summary
Create a management command to rebuild the translated page connections.

The db export does not include wagtail-localize data, since we do not want to leak non-published/draft data, and this means that locally, translated pages appear to not be connected (they don't look like they're translations of each other).
This pull request adds a management command to rebuild the wagtail-localize database tables for the exported data, so they appear as translations of each other.
Also, we add a comment to the db export scripts, so that users have a quick reference for which command to run to rebuild the page connections.

## Issue / Bugzilla link
[WT-822](https://mozilla-hub.atlassian.net/browse/WT-822)

## Testing
1. run `./bin/run-db-download.py`
2. run `python manage.py link_translations_after_export`
3. Go into Wagtail, and observe that pages have translations


springfield/cms/management/commands/link_translations_after_export.py
```diff
@@ -0,0 +1,324 @@
...
+    PreFooterCTASnippet,
+    Tag,
+)
+
+# Suppress spurious BeautifulSoup warning that fires when wagtail-localize
+# processes URL strings (e.g. custom_url in link blocks). The round-trip is
+# correct; the warning is purely cosmetic.
+warnings.filterwarnings("ignore", category=MarkupResemblesLocatorWarning)
+
+TRANSLATABLE_SNIPPET_MODELS = [
```
maribedran: I think we can get a list of all registered snippets from Wagtail and check if they inherit from translation mixin.

stevejalim (approved):
LGTM - interesting that we couldn't just bring the relevant tables, but if this works that's wonderful

One usability suggestion: what do you think about calling the new management command from `make preflight` ([source](https://github.com/mozmeao/springfield/blob/main/Makefile#L161-L165)) if the `--retain-db` flag is not passed ?

PR #1101, opened 2026-03-02
pass github parameter in env variables

fkiriakos07:
## One-line summary
pass github parameter in env variables


fkiriakos07:
Hello @stevejalim , can you please take a quick look? Thanks.

PR #1104, opened 2026-03-02
disabled buttons revised in dark mode

kkellydesign:
## One-line summary

On light mode, the primary button disabled styles are reverting to default on light mode, even with the `.force-dark` class applied. 
Screenshot 2026-03-02 at 12 00 33 PM

This is fixed: 
Screenshot 2026-03-02 at 12 01 52 PM

Test here: http://localhost:8000/en-US/features/control/


maureenlholland (approved):
r+

PR #1152, opened 2026-03-16
WT-841 Show correct Windows download link

bluewave41:
## One-line summary
Windows users are shown the 32 bit download link on https://www.mozilla.org/firefox/installer-help/?channel=release regardless of their current architecture.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1564333

## Testing

Before each step remember to change the HTML class OS to "windows"
The x64 doesn't matter as it doesn't affect the shown download button. Only "win" is ever displayed

- [x] Checkout this PR
- [x] Visit http://localhost:8000/en-US/download/installer-help/?channel=release
- [x] Remove the x64 class from HTML if it exists
- [x] The download link should be `https://download.mozilla.org/?product=firefox-latest-ssl&os=win&lang=en-US`
- [x] Add the x64 class back to HTML
- [x] The download link should still be `https://download.mozilla.org/?product=firefox-latest-ssl&os=win&lang=en-US`
- [x] Visit http://localhost:8000/en-US/download/installer-help/?channel=release&installer_arch=1
- [x] The download link should be `https://download.mozilla.org/?product=firefox-latest-ssl&os=win&lang=en-US`
- [x] Visit http://localhost:8000/en-US/download/installer-help/?channel=release&installer_arch=2
- [x] The download link should be `https://download.mozilla.org/?product=firefox-latest-ssl&os=win64&lang=en-US`
- [x] Visit http://localhost:8000/en-US/download/installer-help/?channel=release&installer_arch=3
- [x] The download link should be `https://download.mozilla.org/?product=firefox-latest-ssl&os=win64-aarch64&lang=en-US`

### DIFF
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -114,6 +114,7 @@ def download_firefox(
     button_class="mzp-t-xl",
     locale_in_transition=False,
     download_location=None,
+    force_arch=None,
 ):
     """Output a "download firefox" button.
 
@@ -171,6 +172,7 @@ def download_firefox(
         "button_class": button_class,
         "download_location": download_location,
         "fluent_l10n": ctx["fluent_l10n"],
+        "force_arch": force_arch,
     }
 
     html = render_to_string("firefox/includes/download-button.html", data, request=ctx["request"])
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -41,16 +41,6 @@ def desktop_builds(
     for plat_os, plat_os_pretty in firefox_desktop.platforms(channel, classified):
         os_pretty = plat_os_pretty
 
-        # Firefox Nightly: The Windows stub installer is now universal,
-        # automatically detecting a 32-bit and 64-bit desktop, so the
-        # win64-specific entry can be skipped.
-        if channel == "nightly":
-            if plat_os == "win":
-                continue
-            if plat_os == "win64":
-                plat_os = "win"
-                os_pretty = "Windows 32/64-bit"
-
         # And generate all the info
         download_link = firefox_desktop.get_download_url(
             channel,
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -162,7 +162,7 @@ def download_firefox(
         "button_class": button_class,
         "download_location": download_location,
         "fluent_l10n": ctx["fluent_l10n"],
-        "force_arch": force_arch
+        "force_arch": force_arch,
     }
 
     html = render_to_string("firefox/includes/download-button.html", data, request=ctx["request"])
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -175,6 +175,9 @@ def download_firefox(
         "force_arch": force_arch,
     }
 
+    if force_arch is not None:
+        data["force_arch"] = force_arch
+
     html = render_to_string("firefox/includes/download-button.html", data, request=ctx["request"])
     return Markup(html)
 
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -172,7 +172,6 @@ def download_firefox(
         "button_class": button_class,
         "download_location": download_location,
         "fluent_l10n": ctx["fluent_l10n"],
-        "force_arch": force_arch,
     }
 
     if force_arch is not None:
--- a/springfield/firefox/templatetags/helpers.py
+++ b/springfield/firefox/templatetags/helpers.py
@@ -172,10 +172,8 @@ def download_firefox(
         "button_class": button_class,
         "download_location": download_location,
         "fluent_l10n": ctx["fluent_l10n"],
+        "force_arch": force_arch
     }
-
-    if force_arch is not None:
-        data["force_arch"] = force_arch
 
     html = render_to_string("firefox/includes/download-button.html", data, request=ctx["request"])
     return Markup(html)
--- a/tests/playwright/specs/products/firefox/firefox-installer-help.spec.js
+++ b/tests/playwright/specs/products/firefox/firefox-installer-help.spec.js
@@ -160,5 +160,68 @@ test.describe(
             await expect(downloadButtonBeta).not.toBeVisible();
             await expect(downloadButtonDeveloper).not.toBeVisible();
         });
+
+        test('Download Firefox (Windows) forced win architecture', async ({
+            page,
+            browserName
+        }) => {
+            const downloadButtonRelease = page.getByTestId(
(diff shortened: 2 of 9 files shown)


l-hedgehog:
> 
> 
> 
> That was my first thought but it seemed like I was touching a larger scope than I wanted to for something that will only be used on a single page.
> 
> 
> 
I think you're now offering win64 installers not only for `installer_arch=2`, but also when `.windows.x64` is applied by `site.js`. I believe the latter was deliberately not the case and should stay that way.

maureenlholland (changes requested):
the installer arch work looks good to me 🎉 

one blocking issue on Nightly download button for http://localhost:8000/en-US/channel/desktop/

We should confirm this PR passes existing integration tests before merge: https://mozmeao.github.io/platform-docs/operations/pipeline/#pushing-to-the-integration-tests-branch
(^ this may require some permissions you don't yet have, if you are blocked here, ask in www slack for permissions)

With any new download functionality, we should also add new tests 
- jasmine: https://mozmeao.github.io/platform-docs/testing/jasmine/
- playwright: https://mozmeao.github.io/platform-docs/testing/playwright/

Jasmine is probably the best for the new logic in site.js, but we might need to adjust or update Playwright to confirm the channel page download functionality too

janbrasna:
Right I knew I wanted to add more notes to the above, incl. specifically mentioning how I'm forcing the CTAs to show, sorry forgot — basically that should be the underlying nonJS version, or, shown in cases no supported UA is detected. I usually use Focus iOS for that, which has a weird enough UA;D — but the easiest is "non matching" products, e.g. opening [Nightly](https://www.firefox.com/en-US/firefox/150.0a1/releasenotes/) or [ESR](https://www.firefox.com/en-US/firefox/128.12.0/releasenotes/) release from iOS, as these do not have a product for iOS — so all other options get listed instead. Responsive mode with iPad UA should do the trick for you: _(reload after changing the device/UA to force new parsing)_

> responsive

(I also wanted to understand the linked tickets with reasoning when the special casing was added back then, still catching up on code archaeology, will post what I learn.)

tests/playwright/specs/products/firefox/firefox-installer-help.spec.js
```diff
@@ -160,5 +160,68 @@ test.describe(
...
+
+            await openPage(
+                url + '?channel=release&installer_arch=1',
+                page,
+                browserName
+            );
+            await expect(downloadButtonRelease).toBeVisible();
+        });
+
+        test('Download Firefox (Windows) forced win64 architecture', async ({
```
maureenlholland, line 185: **non-blocking** is it also worth checking the unexpected win downloads are not visible?

maureenlholland (approved):
This seems a well-scoped update for the installer-arch pages

- new force arch param on download helper defaults to None
- only the installer arch page calls the helper with force arch argument
- "force-" css classes will only take effect when force arch is not None

I don't have any context to add around the Nightly special casing. It seems like a docs update might be helpful here if we need to keep it.

r+ but will call in another dev to check on the Nightly case

springfield/firefox/templatetags/helpers.py
```diff
@@ -41,16 +41,6 @@ def desktop_builds(
     for plat_os, plat_os_pretty in firefox_desktop.platforms(channel, classified):
         os_pretty = plat_os_pretty
 
-        # Firefox Nightly: The Windows stub installer is now universal,
-        # automatically detecting a 32-bit and 64-bit desktop, so the
-        # win64-specific entry can be skipped.
-        if channel == "nightly":
```
maureenlholland, line 47: @stevejalim do you know if this special casing is still necessary (related tests here: https://github.com/mozmeao/springfield/pull/1152/changes#diff-787fdc6db4abb92a972f332cdf45bf50fbe8d61618b74f507a937285e8ff6139L173)

More detail in comments: https://github.com/mozmeao/springfield/pull/1152#issuecomment-4179203683

bluewave41: Keeping this behavior makes it confusing and would make the CSS look worse

Release
installer_arch=1? show `.win`
installer_arch=2? show `.win64`
installer_arch=3? show `.win64-aarch64`

Nightly
installer_arch=1? show `.win`
installer_arch=2? show nothing. win64 sets the platform to win but arch 2 shows win64 which doesn't exist
installer_arch=3? show `.win64-aarch64`

So now we need to specifically CSS nightly and show `.win` for forced arch 1 and 2. This also means you can't get a 64 bit installer for nightly from this page despite one existing. The stub installers are universal but these full installers are different

https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central/firefox-151.0a1.en-US.win32.installer.exe
https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central/firefox-151.0a1.en-US.win64.installer.exe

```
matthew:~/Downloads$ md5sum '/home/matthew/Downloads/firefox-151.0a1.en-US.win64.installer.exe' 
a662d6d7a3d2f19f4d638bf57a5b90bd  /home/matthew/Downloads/firefox-151.0a1.en-US.win64.installer.exe
matthew:~/Downloads$ md5sum '/home/matthew/Downloads/firefox-151.0a1.en-US.win32.installer.exe' 
0140dd3abcbb070f8889d4d8b83fc653  /home/matthew/Downloads/firefox-151.0a1.en-US.win32.installer.exe
```

maureenlholland:
on Demo 3

installer help
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=release
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=release&installer_arch=1
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=release&installer_arch=2
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=release&installer_arch=3

https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=nightly
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=nightly&installer_arch=1
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=nightly&installer_arch=2
https://www-demo3.springfield.moz.works/en-US/download/installer-help/?channel=nightly&installer_arch=3

channels
https://www-demo3.springfield.moz.works/en-US/channel/desktop/

release notes
https://www-demo3.springfield.moz.works/en-US/firefox/150.0a1/releasenotes/ (need to be on non-Fx browser to see Nightly download link)

stephaniehobson:
@maureenlholland @bluewave41 Is this blocked on something?

stevejalim:
Note: the demo push failed because somehow the original branch name went up, not fxc-demo-3.
I'll try re-pushing this now by checking out the branch and doing `git push -f HEAD:fxc-demo-3`

maureenlholland:
Confirmed we want to go ahead with removing the Nightly special case and using separate 32/64bit buttons on the long download lists

Issue #1149, opened 2026-03-16
Nav hover underline includes whitespace


janbrasna:
## Description

I know this has been subject of QA before so expecting this time it's also not the ideal result. 

### Steps to reproduce

Hover navigation items with symbols. Compare footer (external link symbol) and dropdown menu items (arrows) with main menu top level items (hover affordance chevrons) and check how long the underline is for these.

### Expected result

No underline for the whitespace, like in the dropdown panels or footer:

| <img width="128" height="102" alt="footer" src="https://github.com/user-attachments/assets/f1e81b45-2599-4595-b8ae-dc8e424cf4e1" /> | <img width="253" height="41" alt="dropdown" src="https://github.com/user-attachments/assets/35da2a84-abe0-4ad7-8ebb-111f9b3db449" /> |
|--------|--------|

### Actual result

Underline on hover state on top level nav items runs into the whitespace:

> <img width="611" height="85" alt="nav" src="https://github.com/user-attachments/assets/c9edb0d8-0be4-4fa2-b934-323ebddd2a51" />

PR #1154, opened 2026-03-16
Remove whitespace underline in navigation

bluewave41:
## One-line summary
The navigation elements (Browser, Features, Resources) underline a whitespace which doesn't match the rest of the page

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/issues/1149

## Testing
- [x] Checkout this PR
- [x] Hover over Browser, Features and Resources
- [x] There should no longer be an underline under the white space between the end of the text and the icon

### DIFF
--- a/media/css/cms/flare26-navigation.css
+++ b/media/css/cms/flare26-navigation.css
@@ -32,7 +32,7 @@
     text-decoration: none;
 }
 
-.fl-header a:hover .fl-menu-heading {
+.fl-header a:hover {
     text-decoration: underline;
 }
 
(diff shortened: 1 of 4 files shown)


media/css/cms/flare26-navigation.css
```diff
@@ -32,7 +32,7 @@
     text-decoration: none;
 }
 
-.fl-header a:hover {
+.fl-header a:hover .fl-menu-heading {
```
maureenlholland: **non-blocking, maintenance**: Ideally, we would not have to add another class to an already-heavy selector.
Image

I think the only other approach we could take is refactoring `.fl-menu-title` to use a flex layout so the space comes through `gap` and not whitespace. But that doesn't seem worth the extra time here.

PR #1169, opened 2026-03-19
Wt-931 Add Notification and Banner blocks to 2026 Free Form and What's New pages

maribedran:
## One-line summary

This PR adds notification and banner blocks to free form and WNP pages.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-931

## Testing

Run `./manage.py load_page_fixtures` to create the sample pages locally, then check the http://localhost:8000/en-US/tests-index-page-2026/test-banner-2026-page/ and http://localhost:8000/en-US/tests-index-page-2026/test-notifications-page/ pages.

### DIFF
--- a/media/css/cms/flare26-banner.css
+++ b/media/css/cms/flare26-banner.css
@@ -32,12 +32,12 @@
 .fl-banner .fl-heading-group {
     display: flex;
     flex-direction: column;
-    gap: var(--token-spacing-3xl);
     margin-block-end: 0;
 }
 
 .fl-banner .fl-heading,
-.fl-banner .fl-subheading {
+.fl-banner .fl-subheading,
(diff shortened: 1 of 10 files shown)


kkellydesign (changes requested):
Re: notifications: 

In the test page, I'm seeing an empty notification heading that is messing with the height. And/or, the `remove_p_tag` filter isn't being honored, which is screwing up the HTML. 

Looking at these in Figma, they should receive the heading style if there is no body content, and related, to make sure this aligns with what Tess is expecting, lets add a field for content in addition to the headline. Figma ref: 
https://www.figma.com/design/5ebjB97defiRGZ7C5N0Hoi/Firefox.com-Design-System?node-id=111126-106639&m=dev


Image

PR #1182, opened 2026-03-23
[WT-947] Enable User Privacy page on the CMS

maribedran:
## One-line summary

Update the Free Form Page 2026 with the blocks necessary to build the existing User Privacy page.

- Created a new Topics List block
- Added the option to hide extra links from the download button block

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-947

## Testing

Create a new Free Form 2026 page with the same content from https://www.firefox.com/en-US/user-privacy/
- Intro
- Topic List
- Banner (purple)
- Banner (default). Add 3 buttons:
  - Download Firefox button (uncheck the option to display extra links to hide the privacy notice link)
  - UI Tour Button with the privacy preferences option
  - Regular button with the link to the privacy notice
 
image

image

### DIFF
--- a/springfield/cms/templates/cms/blocks/topic.html
+++ b/springfield/cms/templates/cms/blocks/topic.html
@@ -27,6 +27,7 @@
   <content:buttons>
     {% for button in value.buttons %}
       {% set block_position = block_position ~ ".button-" ~ loop.index %}
+      {% set block_text = value.heading.heading_text|richtext|remove_tags %}
       {% include_block button %}
     {% endfor %}
   </content:buttons>


kkellydesign (changes requested):
Topic list: 
* lock the font size on the headlines down to `fl-heading-md` regardless of heading level
* add a max-width of 90px to the images
* make the gap on the buttons 16px `var(--token-spacing-lg)`
* The line length is too long on the paragarphs. Add `max-width: 80ch;`
* Which then means `.fl-topic-list` needs `justify-content: start` to fix the layout. 

Banner: 
* On the last banner, I think we should find a way to keep the larger headline, since this is one page that Tess did spec out in detail. What do you think of this: 
    * On the 'slim layout' checkbox, revise the help text to say "Use a more compact layout with reduced spacing and a smaller headline". Then, when that's checked, use `fl-heading-md`, otherwise by default, use `fl-heading-lg`. That lines up with this design anyway, because the banner above it should use `-md` (It does already. Also, I realize the current bottom banner uses `-xl`, but I don't think it should. I need to pin down these rules with Tess.)
* The max width of `.fl-banner-content` is currently locked down at `840px`. Change that to `var(--token-width-desktop-wide-banner)`

PR #1195, opened 2026-03-26
updating /landing/set-as-default/thanks/ for new look

slightlyoffbeat:
## One-line summary

We will eventually have a component that will remove the need for this page, but in the meantime we still want to send traffic here so we are updating the look/feel of this page to match flare26

http://localhost:8000/en-US/landing/set-as-default/thanks/

1. The download_firefox() macro was removed in favor of <include:download-firefox-button>. 
2. The extra links section was intentionally removed (mobile download, Mozilla account, help). The "Having trouble?" link remains in the desktop instructions state.
3. The JS file was not modified — all UITour logic, GA4 events, and polling are untouched.

### DIFF
--- a/media/js/cms/new.es6.js
+++ b/media/js/cms/new.es6.js
@@ -5,6 +5,7 @@
  */
 
 import VideoEngagement from '../base/datalayer-videoengagement.es6';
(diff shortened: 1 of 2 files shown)


slightlyoffbeat:
Before:
image

After:
image

PR #1207, opened 2026-03-30
WT-962: Set as default page in Wagtail

lucianoratamero:
## One-line summary

This PR adds a Set as Default button block and its snippet counterpart, to enable setting your browser as default from anywhere in the app.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-962

## Testing

We added a fixture that creates a freeform page with the set as default button, and adds a test snippet to the db.
Run `./manage.py load_page_fixtures`, then navigate to the page (it's inside the Tests Index Page 2026 page), and check that the contents are loaded correctly for every scenario.

We recommend[ enabling the UI tour locally](https://mozmeao.github.io/platform-docs/integrations/uitour/?h=ui#javascript-api) and using [an user agent browser extension](https://addons.mozilla.org/en-US/firefox/addon/user-agent-string-switcher/) to simulate different OSs.


kkellydesign (dismissed):
I built a snippet in the CMS and then added a button to a freeform page (within a card) and received this message. 

I like Mariana's idea of using a snippetBlockChooser, so let's go ahead and refactor to use that approach

Image

kkellydesign (changes requested):
* Remove the empty heading P if there's no content. 
* left-align text if there's an icon present. 

Image

kkellydesign:
Screenshot 2026-04-09 at 10 58 08 AM

There's a rogue `<p></p>` too. We probably another conditional somewhere.

kkellydesign:
One brief tweak: let's make the "success" message inline, so if the user is already set to default on page load, they don't have to click on a modal to see it. Thoughts?

Screenshot 2026-04-09 at 12 16 45 PM

tesseheinricks:
Taking a look now! Excellent work! Recording notes below:

1. Are we able to change functionality so that the thanks modal closes with the X button and clicking outside of the modal (anywhere on the overlay)?
2. Any ways to smooth the transitions slightly?
3. Can we utilize the purple modal styling as opposed to the green? 

I'm also only seeing the modal upon completion, where / when am I supposed to see the success toast. 

**If we have to choose between experiences, I would recommend:**
- The purple success notification shows for a set duration at top of page after setting as default. No pop-up modal
- What becomes of the primary CTA button? After a user has changes to Firefox as default this button should disappear or transition to support related content in case of error

Let me know if you have any questions and thank you!

lucianoratamero:
I fixed the text align on notifications, thanks for catching that @kkellydesign !

answering @kkellydesign : 

about the modal: currently, both the dialog open and the set as default function triggers are tied to the user clicking the button. we could change the behavior so that the only trigger is the set as default function; and if it detects that we're on FF Desktop (it only works on that combination), we could just not open the modal at all, leaving the modal open conditional to anything that's not FF Desktop.

we would lose an important bit of info though: it might be the case that the set as default function fails on FF Desktop (I could 100% see this happening on linux), and since the message of "if something went wrong, click here for the docs" is in the modal, the user would never see it. there are possible workarounds, and we can discuss this better on slack, but for this reason alone, I think keeping everything in one spot is better, even though I agree that a modal is not ideal.

answering @tesseheinricks :

besides the topic above, I'll answer by topic:

1. yes! the code currently doesn't do that, but it's totally doable. I think it could even be the default, as well as pressing Esc to dismiss. what do you think?
2. we can, but what are you thinking for the transition? maybe a 300ms fade in?
3. I'm not sure what you mean by purple modal styling. the modal is white on light mode and purple on dark. are you referring to the notification message?

one more thing to everyone:

I'm not fond of the notification being rendered inside of the modal. maybe the answer is to either do a notification pop-up outside of the modal and close it on success, or just show a success message inline, inside of the modal.
thoughts?

(also, sorry for the long text :B)

media/css/cms/flare26-utilities.css
```diff
@@ -263,6 +273,7 @@ style rules should have any fallbacks needed for IE8+ and Safari 12.1+
     display: revert;
 }
 
+html.firefox-is-default .conditional-display.condition-unknown:not(.is-preview),
```
maribedran: @lucianoratamero I didn't understand this one. The `condition-unkown` refers to the platform. I don't get the relation between this and the default state.

lucianoratamero: tbh, I never fully understood the conditionals, so I thought this was needed. 😅 
I've removed it, since it made no sense.
thanks for the catch!

kkellydesign:
(duplicate note from Slack)

* All of the text is in a snippet, so it can be changed and translated with Smartling. 
* Take a look in Firefox with it set as your default browser, and you'll see that there is no action to take. 
* Change your default browser to something else (In Chrome, visit this URL chrome://settings/?search=default+browser), then refresh the page in Firefox. You'll see an encapsulated user flow with a modal on the page you're on.
* Look at the page in Chrome. The modal will give you a download button instead of the 'make default' flow. 

This component can be used anywhere on the site that a button can, and eliminates the need for the /set-as-default/thanks/ page.

https://www-demo2.springfield.moz.works/en-US/tests-index-page-2026/freeform-2026-with-set-as-default/ (edited)

kkellydesign (approved):
This all looks good to me, but I want to double-check with @tesseheinricks and @slightlyoffbeat, because we made a handful of UI decisions here.

PR #1222, opened 2026-04-03
Add visual regression testing for flare 26 components

lucianoratamero:
## One-line summary

This PR adds visual regression testing for all flare 26 components that are currently in the pattern library.

If you want more info, I've made a video on it: https://drive.google.com/file/d/12UKtSe9F7IEGVWAjxEfY4YaEUwnw9Fww/view?usp=sharing

Playwright already has tools for visual regression testing built-in. One caveat is that it generates one png per test, per browser, per OS, and that gets out of hand pretty quickly. To mitigate that, we're running the visual regression tests using docker, and only on Chromium, so we have one png per test.

## Testing

Go into the tests/playwright folder and, with the server running, run `npm run visual-regression-tests`. For more information, I added a readme to the playwright folder, explaining the details.


springfield/cms/templates/pattern-library/components/cards-list/cards-list_sticker2026.html
```diff
@@ -4,6 +4,7 @@
  file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #}
 
+<div style="background: var(--token-color-dark-purple);" data-testid="cards-list-sticker">
```
maribedran: Shouldn't we use one of the classes that have a purple background? This would guarantee that CSS variables are working properly.

maribedran (commented):
@lucianoratamero I have one neatpick feedback: there are a few tests called "dark theme", but they actually test the browser's dark mode preference. The name suggests that the tests refer to a dark theme variation of the component. I'd imagine something like http://localhost:8000/en-US/tests-index-page-2026/test-card-gallery-2026/

maribedran (approved):

PR #1235, opened 2026-04-07
Include Locale In 'Get Firefox' and 'Download Firefox' Links

dchukhin:
## One-line summary
The 'Get Firefox' and 'Download Firefox' links should now have the current page's locale.

We now pass `locale_in_transition=True` to `download_firefox_thanks` and `download_firefox_thanks_link` in the following files:
 - `springfield/base/templates/includes/navigation/nav-cta.html`
 - `springfield/cms/templates/components/download-firefox-button.html`

## Issue / Bugzilla link
Previously, links on pages would point the user to the page in the locale of the current page, except for the 'Get Firefox' and 'Download Firefox' links.

## Testing
load the homepage  (or any other page) in a locale different from your browser's locale, and observe that all links take you to a page in the current page's locale. For example (assuming your browser's locale is not es-MX),
1. load http://localhost:8000/es-MX/, and observe that the navbar link "Caracteristicas" goes to  http://localhost:8000/es-MX/features/. 
2. observe that the "Obtener Firefox" button goes to  http://localhost:8000/es-MX/thanks/
3. observe that the "Descarga Firefox" button goes to  http://localhost:8000/es-MX/thanks/

### DIFF
--- a/springfield/base/templates/includes/navigation/nav-cta.html
+++ b/springfield/base/templates/includes/navigation/nav-cta.html
@@ -6,7 +6,7 @@
 
 {% if not custom_nav_cta %}
   <div class="nav-cta-wrap {% if hide_nav_cta %}hide-cta-on-desktop{% endif %} {% if hide_nav_download_button %}hide-nav-download-on-desktop{% endif %}">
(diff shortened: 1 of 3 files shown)


dchukhin:
It looks like the differing behavior for the 'Get Firefox' and 'Download Firefox' links is intentional: the idea is that users will want to see the "thanks" page in their browser's language, and not be limited to the locales that exist in Wagtail.

PR #1241, opened 2026-04-08
Create "Compare" And "More" Pages In Wagtail

dchukhin:
## One-line summary
Create a Page in Wagtail for each of the current static "compare" and "more" pages, in each relevant locale.

Similar to the work done in https://github.com/mozmeao/springfield/pull/952, this pull request creates all of the "compare" and "more" pages as Pages in Wagtail, including their translations.

 - the existing static pages continue to be served for now, but are now wrapped with `prefer_cms()`
 - a new `create_compare_more_pages` management command creates a `Page` in Wagtail for each of the compare and more pages, and their translations
 - the management command is called from a migration, so that it is run by default on deploy, but can be run manually as needed

The following static pages are now these Wagtail Pages:

`/compare/` - `ArticleIndexPage`
`/compare/brave/` - `ArticleDetailPage`
`/compare/chrome/` - `ArticleDetailPage`
`/compare/edge/` - `ArticleDetailPage`
`/compare/opera/` - `ArticleDetailPage`
`/compare/safari/` - `ArticleDetailPage`
`/more/` - `ArticleIndexPage`
`/more/best-browser/` - `ArticleDetailPage`
`/more/browser-history/` - `ArticleDetailPage`
`/more/faq/` | `ArticleDetailPage`
`/more/incognito-browser/` - `ArticleDetailPage`
`/more/update-your-browser/` - `ArticleDetailPage`
`/more/what-is-a-browser/` - `ArticleDetailPage`
`/more/windows-64-bit/` - `ArticleDetailPage`

## Issue / Bugzilla link
[WT-961](https://mozilla-hub.atlassian.net/browse/WT-961)

## Testing

1. Observe the static compare pages at http://localhost:8000/compare/ and the static more pages at http://localhost:8000/more/
2. create the (draft) pages in Wagtail: `ALL_LOCALES=1 python manage.py create_compare_more_pages`
3. Observe that the new pages are translated (some translations are missing, just like the current static pages)
4. publish one of the new compare pages (for example, the compare brave page), then go its URL (http://localhost:8000/en-US/compare/brave/), and observe that the Wagtail page is being served, and also that it shows up on the Wagtail compare list page
5. publish one of the new more pages (for example, the more faq page), then go its URL (http://localhost:8000/en-US/more/faq/), and observe that the Wagtail page is being served, and also that it shows up on the more list page with the expected sticker

Notes:
 - some pages (see https://www.firefox.com/en-US/more/incognito-browser/ and https://www.firefox.com/en-US/more/update-your-browser/) have no translations on the live site, so they do not currently get Wagtail Pages for non-en-US locales.
 - some of these static pages have CTAs (see the top & bottom of https://www.firefox.com/en-US/more/windows-64-bit/ , see the top of https://www.firefox.com/ru/more/faq/ ), but the CMS pages don't get this CTA, because we don't have a CTA field on the article detail page.


springfield/cms/management/commands/create_compare_more_pages.py
```diff
@@ -0,0 +1,238 @@
...
+    "ja",  # Japanese
+    "id",  # Indonesian
+    "nl",  # Dutch
+    "tr",  # Turkish
+]
+
+
+def _get_locale_codes(stdout):
+    """Return locale codes to translate, respecting DEBUG / ALL_LOCALES env var."""
+    from django.conf import settings
```
maribedran: Move import to the top

kkellydesign (changes requested):
@dchukhin Couple of revisions here:

* The index page designs aren't quite working here. Add a "Index Card type" dropdown on the "settings" area, and add "outline card" "illustration card", and "sticker card" options. "Sticker card" should be the default, because in use on prod currently. Since you're building these programatically anyway, make `/compare` use the outline card (no image) and `/more` use the illustration card (bigger image). 
* the images themselves didn't copy over on demo1. There are only a handful, so it's not a big deal to do it manually if we have to, but this seems like a place where it might work fine on demo but get tripped up on the more strict dev/prod server settings. If you think it might be a problem, it's ok to copy these manually once the DB is populated. 

Image

https://www-demo1.springfield.moz.works/en-US/more/
https://www-demo1.springfield.moz.works/en-US/compare/

kkellydesign (changes requested):
One more: we do need to add CTA buttons to the article template, because the windows-64-bit page gets quite a lot of traffic. This is a tangent, so please make a new PR for the following request, and we'll make sure to merge that one before we return to this one.

* The "Download Firefox" button needs a dropdown in the settings area called "specific version" so we can incorporate all of the right analytics tags, etc. for this button. Looking at the code, the `macros.html` page that contains `firefox_download_desktop_button_windows` has some other versions, so it's probably a good idea to go ahead and include them. Use your best judgement though, I'm not sure what sort of lift I'm asking for here.
* In the article template, add a component called "button row" that can be added in the content area. (In this case, we'd add it at the top and bottom). It'd be a div with `class="fl-buttons is-horizontal is-center extra-padding"`, and can take the same button types as other places we allow buttons.  

I'm going to create a new Jira ticket for this one, but lets just leave this note here as unresolved until the task is done. 

Image

kkellydesign (approved):

PR #1244, opened 2026-04-09
WT-823 hide current platform as a download option

wen-2018:
## One-line summary

This PR removes the current platform as a dropdown option.

Dropdown list on download/{platform}

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-823

## Testing

http://localhost:8000/en-US/download/windows/
http://localhost:8000/en-US/download/mac/
http://localhost:8000/en-US/download/linux/
https://www.firefox.com/en-US/download/android/
https://www.firefox.com/en-US/download/ios/
https://www.firefox.com/en-US/download/chromebook/

### DIFF
--- a/springfield/cms/templates/cms/download_page.html
+++ b/springfield/cms/templates/cms/download_page.html
@@ -37,11 +37,13 @@
         </button>
         <ul class="fl-platform-dropdown-list" aria-role="listbox">
           {% for key, platform in platforms.items() %}
-            <li class="fl-dropdown-item" data-platform="{{ key }}" {% if key == page.platform %}aria-selected="true"{% endif %}>
-              <a class="fl-heading-md fl-heading-condensed-on-sm-up" href="{{ platform_links[key] }}">
-                {{ platform }}
-              </a>
-            </li>
+            {% if key != page.platform %}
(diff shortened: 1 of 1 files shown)


springfield/cms/templates/cms/download_page.html
```diff
@@ -37,7 +37,7 @@ <h1 class="fl-heading fl-heading-size-1">
         </button>
         <ul class="fl-platform-dropdown-list" aria-role="listbox">
           {% for key, platform in platforms.items() %}
-            <li class="fl-dropdown-item" data-platform="{{ key }}" {% if key == page.platform %}aria-selected="true"{% endif %}>
+            <li class="fl-dropdown-item {% if key == page.platform %}hidden{% endif %}" data-platform="{{ key }}" {% if key == page.platform %}aria-selected="true"{% endif %}>
```
janriokrause: Is there any specific reason why you are hiding the element with CSS instead of not rendering at all?

    {% if key != page.platform %}<li class="fl-dropdown-item" ...>...</li>{% endif %}

wen-2018: This is better. Updated!

janriokrause (approved):
r+ Great. Thanks, @wen-2018

PR #1245, opened 2026-04-09
Make Sure Page References In Blocks Match Locale Of Page

dchukhin:
## One-line summary
When a page is served in a fallback locale, any references to pages in the page's content should now be in the same locale as the page content.

 - the most significant change is that `AbstractSpringfieldCMSPage.localized` now extends `Page.localized` to make sure that we are returning the appropriate page in fallback scenarios
 - `BaseArticleValue.get_article()` now calls `.specific` before `.localized`, so that `AbstractSpringfieldCMSPage.localized` gets called. Without the `.specific`, `.localized` calls `Page.localized` (not our `AbstractSpringfieldCMSPage.localized`)
 - `Locale.get_active()` also properly handles the scenario where the alias locale does not exist in the database

## Issue / Bugzilla link
Previously, when the requested page didn't exist in an alias locale, and the content was served in the fallback locale at the alias URL , if the fallback content included a foreign key to a Page, that Page reference shown to the end user may
have been in the default (en-US) locale. Now, the Page reference should match the locale that the content is served in.
Example: because there is no features protection page in the pt-PT locale, https://firefox.com/pt-PT/features/protection/ serves pt-BR content. However, "Learn more" links show up on the page, because the article detail pages being linked to are being returned in the default (en-US) locale.

Examples from janbrasna:
Screenshot1
Screenshot2

## Testing
- load the problematic pages locally, and observe that the content is fully translated:
  - http://localhost:8000/pt-PT/features/protection/ should show the same content as http://localhost:8000/pt-BR/features/protection/  "Learn more" should not show up; it should be "Saiba mas"
  - http://localhost:8000/pt-PT/features/focus/ should show the same content as http://localhost:8000/pt-BR/features/focus/       "Learn more" should not show up; it should be "Saiba mas"
  - http://localhost:8000/es-AR/features/protection/ should show the same content as http://localhost:8000/es-MX/features/protection/            "Learn more" should not show up; it should be "Leer más"

### DIFF
--- a/springfield/cms/models/base.py
+++ b/springfield/cms/models/base.py
@@ -154,6 +154,27 @@ class AbstractSpringfieldCMSPage(WagtailBasePage):
 
         return localized
 
+    def get_fallback_url(self, request=None):
+        """
+        Replace the URLs locale with the active locale if the page is a fallback
+        so that the user doesn't navigate away from it's preferred language.
+
+        If the active locale is an alias (e.g. pt-PT → pt-BR) and the page is in the
+        fallback locale (e.g. pt-BR), return a URL with the alias locale (e.g. pt-PT).
+        host/pt-BR/page/ → host/pt-PT/page/
+        """
+        url = super().get_url(request)
+
+        active_language = normalize_language(translation.get_language())
+        fallback_locales = getattr(settings, "FALLBACK_LOCALES", {})
+
+        if active_language in fallback_locales:
+            fallback_code = fallback_locales[active_language]
+            if self.locale.language_code == fallback_code:
+                url = url.replace(f"/{fallback_code}/", f"/{active_language}/", 1)
+
+        return url
+
     @property
     def og_title(self):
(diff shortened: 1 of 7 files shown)


maribedran (approved):

springfield/cms/models/base.py
```diff
@@ -122,6 +124,57 @@ def serve_preview(self, request, *args, **kwargs):
...
+                if localized.locale_id != fallback_locale.id:
+                    fallback_page = self.get_translation_or_none(fallback_locale)
+                    if fallback_page:
+                        return fallback_page
+            except Locale.DoesNotExist:
+                pass
+
+        return localized
+
+    def get_fallback_url(self, request=None):
```
dchukhin: This looks like it works, but the name `get_fallback_url` initially made me think that it's going to return the URL for the fallback locale, which isn't right.
Maybe something like `get_url_in_active_locale` would be better?

PR #1247, opened 2026-04-10
[WT-996] Smart Window

maribedran:

Existing blocks changes:
- Media + Content now uses a stream field for the content, and the new Smart Window instructions can be used in it (added data migration)
- Media + Content has a new `narrow` layout option
- Media + Content can be added outside a section (*Note*: this change required refactoring on how the `block_level` variable is incremented in block templates. The variable is used to determine heading levels. Tests should reflect this change, and the Wagtail user bar can identify if headings are being skipped.)
- Illustration cards now use the Media block instead of the Image block (added data migration)
- The Cards List block now has a scroll option that places all cards in a row and adds a scroll animation with Swiper
- New UITour Button option to enable Smart Window

New blocks:
- Icon Card (the pattern was already used on the Article Index page)
- Line Cards (the pattern was already used on the Article Index page)
- Testimonial Card (used on the new page with the scroll option on the Cards List)
- Sliding Carousel


media/js/cms/new.es6.js
```diff
@@ -560,6 +594,7 @@ if (typeof window.cms === 'undefined') {
         initDownloadDropdown();
```
slightlyoffbeat: Bug: `initAnimationPauseButtons()` is called in the `DOMContentLoaded` branch (line 580) but is missing from the `else` branch here. If the script loads after the DOM is already ready, pause buttons won't be initialized.

springfield/cms/models/pages.py
```diff
@@ -885,3 +897,152 @@ def get_utm_campaign(self):
...
+        related_name="+",
+        help_text="Optional dark mode mobile variant of the image.",
+    )
+
+    content = StreamField(
+        FREEFORM_PAGE_BLOCKS_2026,
+        use_json_field=True,
+    )
+
+    show_smart_window_button = models.CharField(
```
slightlyoffbeat, line 947: Nit: `show_smart_window_button` is the only CharField on this model without `max_length` — the others (`animation_alt`, `waitlist_button_label`, `form_submit_label`) all specify `max_length=255`. Django 5.2 allows this, but since this PR also adds the model to `bin/export-db-to-sqlite.sh`, it's worth being consistent. The migration at 0064:86-92 would need updating too.

media/js/cms/flare26.es6.js
```diff
@@ -58,6 +58,167 @@ function initFlare26Carousel(rootEl) {
...
+    function step() {
+        if (!paused) {
+            position += SPEED;
+            const loopWidth = getLoopWidth();
+            if (position >= loopWidth) {
+                position -= loopWidth;
+            }
+            el.scrollLeft = Math.floor(position);
+        }
+        requestAnimationFrame(step);
```
slightlyoffbeat: `initScrollingCardGrid` starts a `requestAnimationFrame` loop that runs for the lifetime of the page — it pauses on hover/focus but never calls `cancelAnimationFrame()`. This means continuous rAF callbacks even when the element is scrolled well off-screen. An `IntersectionObserver` to pause when not visible would be a low-effort improvement.

springfield/cms/templates/cms/smart_window_page.html
```diff
@@ -0,0 +1,208 @@
...
+              <input type="hidden" name="lang" id="id_lang" value="en">
+              <div class="fl-newsletterform-field">
+                <label class="fl-newsletterform-label visually-hidden" for="newsletter-email">{{ ftl('newsletter-form-your-email-address') }}</label>
+                <input id="newsletter-email" class="fl-newsletterform-input" type="email" name="email" placeholder="{{ ftl('newsletter-form-yournameexamplecom', fallback='newsletter-form-your-email-here') }}" autocomplete="email" data-testid="newsletter-email-input">
+              </div>
+              <div id="newsletter-details" class="fl-newsletterform-details fl-newsletterform-row-hidden">
+                <div class="fl-newsletterform-consent">
+                  <label for="newsletter-privacy" class="fl-body fl-newsletterform-checkbox-label">
+                    <input type="checkbox" id="newsletter-privacy" name="privacy" required aria-required="true" data-testid="newsletter-privacy-checkbox">
+                    {% if LANG == 'de' %}
```
slightlyoffbeat: This page is English-only, but there are hardcoded DE/FR branches for the legal links here (lines 96-102) that would never execute. These could be removed to simplify. Also, the model defines a `privacy_notice` RichTextField (line 958 in `pages.py`) that isn't used in this template — is that field intended for future use, or could one of these two approaches be cleaned up?

springfield/cms/templates/cms/smart_window_page.html
```diff
@@ -0,0 +1,208 @@
...
+          block_position="nav"
+          exclude_unsupported_content="True"
+        />
+      </include:conditional-display>
+      {% if show_try_smart_window %}
+        <include:conditional-display firefox_condition="is-firefox">
+         <include:uitour_button
+          theme_class="{{ ui_tour_class }}"
+          label="{{ page.waitlist_button_label }}"
+          analytics_text="Try Smart Window"
```
slightlyoffbeat: `analytics_text="Try Smart Window"` is hardcoded here and on line 72, while the displayed button label uses `page.waitlist_button_label`. If an editor changes the label in the CMS, the analytics text would diverge from what the user sees. Consider using the page field for both.

slightlyoffbeat:
**What's working well**

- The heading hierarchy fixes across 8+ templates are a significant accessibility improvement — especially the operator precedence fix in `showcase.html`
- ARIA implementation on the sliding carousel is solid (`aria-current`, `aria-hidden`, `aria-live`, keyboard handling)
- `prefers-reduced-motion` is consistently respected throughout
- Data migrations properly handle both current pages and all revisions
- Semantic HTML throughout (blockquote/cite for testimonials, proper hgroup)
- CSS logical properties for RTL support
- SVG icon cleanup is clean — no dangling references

---

**Follow-up items**

There are items we'll need to follow up on, but they can be left out of this PR and addressed separately:

1. **Firefox version detection** — The page currently shows the UITour button to all Firefox users in supported geos, but we need to distinguish Firefox 150+ (enable) from ≤149 (show update messaging + link to update instructions). `Client._getFirefoxMajorVersion()` in `mozilla-client.js` is available for this.
2. **`?view=` URL params for analytics** (new requirement) — We'd like to append `?view=enable`, `?view=update`, `?view=waitlist`, or `?view=download` depending on user state. `?view=waitlist` should also work as an override to force the waitlist view.
3. **"Copy URL to clipboard"** (new requirement) — For non-Firefox users and Firefox users who need to update, we'd like a button to copy the page URL alongside the download/update messaging.
4. **English-only locale enforcement** (new requirement) — If a user visits from another language, the page should default to English. Future localization date is TBD.
5. **Stub attribution campaign tag** — The page needs `{% block html_attrs %}data-stub-attribution-campaign="smart_window"{% endblock %}` in its template.

media/js/cms/new.es6.js
slightlyoffbeat:
```suggestion
        initAnimations();
        initAnimationPauseButtons();
```

springfield/cms/models/pages.py
slightlyoffbeat:
```suggestion
    show_smart_window_button = models.CharField(
        max_length=10,
```

springfield/cms/templates/cms/smart_window_page.html
slightlyoffbeat:
```suggestion
          analytics_text="{{ page.waitlist_button_label }}"
```

Same change needed on line 72.

springfield/cms/models/pages.py
stevejalim: FWIW, sqlite is more likely to tolerate over-long CharFields than postgres, but - yep - good to have `max_length` added

springfield/cms/models/pages.py
```diff
@@ -885,3 +897,152 @@ def get_utm_campaign(self):
...
+        choices=(
+            ("all", "Show to all users"),
+            ("us_ca", "US and Canada only"),
```
stevejalim: Suggestion, nonblocking:

We move `us_ca` to a constant `ALLOWED_TERRITORIES` and the same for the label, so that it's easier to update later, and we can always use `ALLOWED_TERRITORIES` in other code, making things like `if self.show_smart_window_button == "us_ca":` later easier to understand

```suggestion
            (ALLOWED_TERRITORIES, ALLOWED_TERRITORIES_LABEL),
```

springfield/cms/models/pages.py
```diff
@@ -885,3 +897,152 @@ def get_utm_campaign(self):
...
+        if self.show_smart_window_button == "us_ca":
+            add_never_cache_headers(response)
+        return response
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        country = get_country_from_request(request)
+        context["ui_tour_class"] = UI_TOUR_CLASSES[UITOUR_BUTTON_SMART_WINDOW]
+        context["show_try_smart_window"] = self.show_smart_window_button == "all" or (
+            self.show_smart_window_button == "us_ca" and country in {"US", "CA"}
```
stevejalim: See earlier suggestion about moving from strings to constants/vars

stevejalim (approved):
r+wc - Django/Wagtail side looks good. Added a suggestion for easier maintenance and readabitly, nonblocking

media/js/cms/flare26.es6.js
kkellydesign: We're going to use the swiper carousel library here instead of the custom implementation, so I'm marking this as resolved/outdated for now. This will go away in the next PR.

PR #1249, opened 2026-04-13
fix: filter draft snippets from DownloadFirefoxCallToActionSnippet lookup

stevejalim:
This changeset adds .live() to the get_download_firefox_cta_snippet() queryset so that draft-only (live=False) snippets are never returned to public article pages. 

Without this filter, a CMS editor could save an unpublished draft for a locale and have it rendered on production pages, bypassing the draft/publish workflow.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=2030826


maureenlholland (approved):
r+

PR #1254, opened 2026-04-14
Fix checkbox styles and download button alignment

slightlyoffbeat:
- Fix undefined --checkbox-hover-color variable causing checkmark to disappear on hover
- Use white checkmark SVG for hover+checked state to match non-hover state
- Add border-color rule for checked state to support dark/light mode
- Add transition animation for background and border color changes
- Update all theme contexts with correct design system colors using tokens
- Fix dark mode (prefers-color-scheme and class-based) using wrong light theme colors for checkbox hover
- Left-align download button contents in split layouts with media
- Size down checkbox and label text in download button context

## Testing
Can create a test landing page, and in upper content, create an intro with media on left, media on right, media center, and add a download button wtih the 'download as default'. Should be able to see everything visually to approve changes.

### DIFF
--- a/media/css/cms/flare26-defaults.css
+++ b/media/css/cms/flare26-defaults.css
@@ -211,6 +211,9 @@ input[type='checkbox'] {
     inline-size: 24px;
     margin-block-end: 0;
     padding: 4px;
+    transition:
+        background-color var(--token-transition-base),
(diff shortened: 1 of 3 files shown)


wen-2018 (approved):
r+

PR #1256, opened 2026-04-14
Improved handling of excluding fallback locales from smartling

slightlyoffbeat:
## One-line summary
Pulls list of locales to exclude from smartling from `list(FALLBACK_LOCALES.keys()),` instead of adding manually.


stevejalim (approved):
LGTM

PR #1257, opened 2026-04-15
Update /landing/get to serve Flare26 template for all locales

slightlyoffbeat:
Stop the download-privacy split test and serve the new Flare26 template (get-new.html) for en-US, matching all other locales. The old templates (get.html, get-treatment.html) and download_privacy waffle switch are kept for future reuse.

Also addresses leadership request to make the nav permanently sticky on this page so the download button is always visible on scroll.


maureenlholland (approved):
r+

PR #1267, opened 2026-04-17
Add four view states to Smart Window page with analytics URL params

slightlyoffbeat:
Add four view states to Smart Window page with analytics URL params

- Enable (?view=enable): Firefox 150+ in supported geo — shows UITour button
- Update (?view=update): Firefox <150 in supported geo — shows update messaging
- Download (?view=download): non-Firefox in supported geo — shows update link
- Waitlist (?view=waitlist): not in supported geo, or ?view=waitlist override

Server-side: get_context() renamed show_try_smart_window to in_supported_geo, added ?view=waitlist param override to force waitlist regardless of geo.

Client-side: inline script adds is-firefox-150-plus class to <html> for CSS-driven enable/update visibility. Second inline script appends ?view= URL param via history.replaceState() for analytics tracking.

Nav CTAs (download + UITour button) now geo-gated — not-in-geo users see no nav CTA since their action (waitlist) is in the main content area.

Waitlist form moved outside conditional-display wrapper when not in geo so all browsers (not just Firefox) can see it.

## Testing

Enable state (Firefox 150+, supported geo): http://localhost:8000/en-US/smart-window/?geo=US
- Should see "Get early access" UITour button
- URL should update to include ?view=enable

Update state (Firefox <150, supported geo):
- You're probably on a recent Firefox, so to simulate old Firefox you'd need to use a different browser (Chrome/Safari) — but that would trigger the download state instead. The update state is hard to test locally unless you have an old Firefox build. You can verify the HTML is present by inspecting the page source and confirming .sw-view-update div exists (just hidden by CSS).

Download state (non-Firefox, supported geo): Open in Chrome/Safari:  http://localhost:8000/en-US/smart-window/?geo=US
 - Should see "How to update Firefox" button + messaging
- URL should update to ?view=download

Waitlist state (not in supported geo): http://localhost:8000/en-US/smart-window/?geo=DE
- Should see email form + "Join the Waitlist" in any browser
- URL should update to ?view=waitlist

Waitlist override: http://localhost:8000/en-US/smart-window/?geo=US&view=waitlist
- Even though geo=US, should see the waitlist form
- URL should show ?view=waitlist

No-JS fallback:
- Open any URL above, disable JS in DevTools, reload
- Verify something reasonable renders (currently the page will show the heading/subheading but the CTA area will be empty for the in-geo case since conditional-display hides by default — this is the existing behavior, not a regression)

## QUESTIONS

Should some of this be handled in CMS vs template level?

### DIFF
--- a/media/css/cms/pages/flare26-smart-window.css
+++ b/media/css/cms/pages/flare26-smart-window.css
@@ -121,6 +121,20 @@
     max-inline-size: 100%;
 }
 
+/* Smart Window view states: version-based visibility */
+
+.sw-view-enable {
+    display: none;
+}
+
+.is-firefox-150-plus .sw-view-enable {
+    display: block;
+}
+
+.is-firefox-150-plus .sw-view-update {
+    display: none;
+}
+
 /* Play/pause buttons */
 
 .fl-smart-window-page .fl-video-play-icon,
--- a/springfield/cms/models/pages.py
+++ b/springfield/cms/models/pages.py
@@ -1025,12 +1025,17 @@ class SmartWindowPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
 
     def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
-        country = get_country_from_request(request)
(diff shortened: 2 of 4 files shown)


slightlyoffbeat:
duplicate. closed.

PR #1271, opened 2026-04-17
Fix Smart Window preview to support view states and geo detection

slightlyoffbeat:
Extract Smart Window context logic into _get_smart_window_context() and call it from both get_context() (live) and get_preview_context( (preview). Previously, preview skipped get_context() entirely, so show_try_smart_window, ui_tour_class, redirect_url, and override_view were undefined — causing preview to always show the waitlist state.


slightlyoffbeat:
This approach is wrong, and based off of incorrect information. Closing.

PR #1287, opened 2026-04-20
Wt 935 review: fixes

slightlyoffbeat:
## Summary

Refinements on top of @bluewave41's floating QR code work. The core feature: the snippet model, per-page overrides, resolution logic, templates, and tests,  was solid. These changes address issues caught during review.

### Use DB queries instead of fixture helpers in production code
`get_context()` on each page was importing `get_floating_qr_code_snippet` from `fixtures/snippet_fixtures.py`. Fixture helpers use `update_or_create`, which means a write query on every page load in production. The fix adds a `get_live_floating_snippet()` helper in `qr.py` that does a read-only query by locale, matching the pattern the existing `get_qr_code_snippet` template tag uses.

### Add missing `get_context()` to WhatsNewPage
The other three page types had `get_context()` overrides to build the `qr` context dict, but WhatsNewPage was missing it. The template needs `qr` to render, so enabling the feature on a WhatsNewPage would crash with `UndefinedError`. Tests now include WhatsNewPage.

### Remove dead `get_context()` from FreeFormPage and WhatsNewIndexPage
FreeFormPage doesn't have QR code fields and its template doesn't render the snippet. WhatsNewIndexPage only redirects via `serve()` — it never renders a template. Both had unnecessary `get_context()` overrides.

### Guard template against None
`qr-code-floating-snippet.html` used `qr.open`, `qr.type`, `qr.value` without checking if `qr` is None. Added `{% if qr %}` wrapper. Also fixed `{%if` → `{% if` spacing and simplified the alt text.

### Scope JS selectors
`document.querySelector('.fl-icon-add')` searches the entire page. Changed to `qrCodeSnippetEl.querySelector(...)` so icon toggles are scoped to the snippet element.

### Fix cookie dismiss for floating snippet
The dismiss cookie was only checked when the old snippet was present. For the floating snippet, dismissing had no effect on the next page load. Now the cookie check collapses the floating snippet on load while still allowing the user to re-open it.

### Add `get_preview_context()` to snippet model
The Wagtail admin preview crashed because the template needs the `qr` context dict. The floating snippet's template uses both `value` and `qr`, so the preview needs to build `qr` from the snippet's own fields.

### Squash migrations
The iterative development produced 4 migrations, including a create-then-drop of `qr_code_floating_button` FK. Squashed into one clean migration that only adds fields and creates the model — no field drops during deploy.

### CSS polish
- `overflow: hidden` on the container — fixes a white line when collapsed and content overlap during animation
- Flexbox layout for vertical centering of QR code and text
- `cursor: pointer` on the header when collapsed and on the close button
- QR code image size increased from 160px to 200px

### Help text on fields
Added `help_text` to override fields on page models and to `url`/`image` on the snippet so editors understand the precedence: an uploaded image overrides the generated QR code from the URL.

## Test plan
- [ ] 70 existing tests pass
- [ ] `makemigrations --check` reports no pending migrations
- [ ] Create a QRCodeFloatingSnippet in admin — preview works
- [ ] Enable on ThanksPage, FreeFormPage2026, WhatsNewPage, WhatsNewPage2026 — renders on all
- [ ] Dismiss → reload → starts collapsed, can re-open
- [ ] Override URL and override image work correctly


slightlyoffbeat:
@stevejalim you can confirm, but I squashed all migrations so that it is only adding what is needed. No removals. This should be risk free.

springfield/cms/migrations/0073_merge_20260420_1609.py
```diff
@@ -0,0 +1,16 @@
...
+
+# Generated by Django 5.2.12 on 2026-04-20 23:09
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0072_freeformpage2026_override_default_open_and_more"),
+        ("cms", "0072_smartwindowpage_smart_window_button_label_and_more"),
```
stevejalim: This will need amending/re-doing - 0073 is already taken on main

springfield/cms/models/snippets.py
```diff
@@ -274,3 +275,60 @@ def get_preview_template(self, request, mode_name):
...
+    panels = [FieldPanel("heading"), FieldPanel("content"), FieldPanel("url"), FieldPanel("image"), FieldPanel("default_open")]
+
+    override_translatable_fields = [SynchronizedField("url"), SynchronizedField("image")]
+
+    class Meta(BaseDraftTranslatableSnippetMixin.Meta):
+        verbose_name = "QR Code Floating Snippet"
+        verbose_name_plural = "QR Code Floating Snippets"
+
+    def __str__(self):
+        from springfield.cms.templatetags.cms_tags import remove_tags
```
stevejalim, line 325: Is this inline import needed here? Claude has a habit of inlining imports that may not need to be put this far down

springfield/cms/models/snippets.py
```diff
@@ -274,3 +275,60 @@ def get_preview_template(self, request, mode_name):
...
+
+        return f"{remove_tags(richtext(self.heading))} – {self.locale}"
```
stevejalim, line 327: This feels a little heavy to me - I wonder if there's a lighter approach

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+#}
+
+{% if qr %}
+<aside class="fl-qr-code-floating-snippet fl-qr-code-snippet-closable {% if qr.open and not qr.hide %}is-open{% endif %}">
+  <div class="fl-qr-code-floating-header">
+    <div class="fl-qr-code-prompt">
+      <span class="fl-icon fl-icon-device-mobile"></span>
+      {{ value.heading|striptags }}
+    </div>
+    <div class="fl-qr-code-close fl-qr-code-snippet-close">
```
maureenlholland:
```suggestion
    <button class="fl-qr-code-close fl-qr-code-snippet-close">
```
this may need a bit of styling too, to override any unwanted default button styles, but it's an interactive element and must be keyboard accessible

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+      {% else %}
+        <span class="fl-icon fl-icon-add"></span>
+      {% endif %}
+    </div>
+  </div>
+
+  <div class="fl-qr-code-floating-content">
+    <div class="fl-qr-code-image">
+      {% if qr.type == "image" %}
+        <img src="{{ qr.value }}" class="fl-qr-code-image" alt="QR code">
```
maureenlholland: **non-blocking, fast-follow**: alt should be a translated string, not hardcoded EN

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+      {% else %}
+        {{ qrcode(qr.value, 12) }}
```
maureenlholland: **non-blocking**: the QR code SVG should generate a title for an accessible name (if that's not possible with the helper, we can add an aria-label on the containing div to act as accessible name for image or SVG)

springfield/cms/models/snippets.py
```diff
@@ -274,3 +275,60 @@ def get_preview_template(self, request, mode_name):
...
+
+    heading = RichTextField(
+        features=HEADING_TEXT_FEATURES,
+        blank=True,
+    )
+    content = RichTextField(
+        features=EXPANDED_TEXT_FEATURES,
+        blank=True,
+    )
+    url = models.CharField(blank=True, help_text="A QR code will be generated from this URL. Not used if an image is uploaded.")
```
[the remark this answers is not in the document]
stevejalim: How did we get passing unit tests then @copilot?

springfield/cms/migrations/0059_freeformpage2026_override_default_open_and_more.py
stevejalim, line 1: It's strange/unexpected (therefore worrying) that this is counted as a deletion compared to `main` because it's a file that doesn't exist on `main`.


I'm going to see if i can clean up this squashing further

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
+{#
+ This Source Code Form is subject to the terms of the Mozilla Public
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+{% if qr %}
+<aside class="fl-qr-code-floating-snippet fl-qr-code-snippet-closable {% if qr.open and not qr.hide %}is-open{% endif %}">
```
maureenlholland: **non-blocking, fast-follow**: this block blends into a light background, likely needs a border or shadow style

Image

Image

springfield/cms/qr.py
```diff
@@ -0,0 +1,27 @@
...
+def resolve_qr_source(page, snippet):
+    resolved_image = getattr(page, "override_image", None) or getattr(snippet, "image", None)
+    resolved_url = getattr(page, "override_url", None) or getattr(snippet, "url", None)
+
+    override_default_open = getattr(page, "override_default_open", None)
+
+    resolved_default_open = override_default_open if override_default_open is not None else getattr(snippet, "default_open", None)
+
+    if resolved_image:
+        return {"type": "image", "value": resolved_image.file.url, "open": resolved_default_open}
```
stevejalim: `resolved_image.file.url` will get the URL of the source image, not any auto-generated rendition of it, which is probably fine for a QR code but we should be aware not to upload a massive one

media/css/cms/flare26-floating-qr-overlay.css
```diff
@@ -0,0 +1,147 @@
...
+    box-shadow: var(--fl-theme-shadow-card);
+    color: black;
+    display: flex;
+    flex-direction: column;
+    inline-size: 22.5rem;
+    inset: auto var(--token-spacing-lg) var(--token-spacing-lg) auto;
+    overflow: hidden;
+    position: fixed;
+    transition:
+        block-size 300ms ease,
```
maureenlholland, line 23: **non-blocking**: until we have support for interpolate-size and keyword sizing, I think the best approach is this grid transition: https://css-tricks.com/css-grid-can-do-auto-height-transitions/

springfield/cms/templatetags/cms_tags.py
```diff
@@ -370,3 +370,31 @@ def get_qr_code_snippet(context):
...
+    Returns the first live snippet for the locale, or None if not found.
+
+    Usage in templates:
+        {% set qr_code_snippet = get_floating_qr_code_snippet() %}
+        {% if qr_code_snippet %}
+            {% set value = qr_code_snippet %}
+            {% include "cms/snippets/qr-code-floating-snippet.html" %}
+        {% endif %}
+    """
+    from springfield.cms.models.snippets import QRCodeFloatingSnippet
```
stevejalim, line 391: Does this import need to be inline? I don't think so, but i can see similar helper tags also do the same so right now am assuming there is a reason

media/css/cms/flare26-floating-qr-overlay.css
```diff
@@ -0,0 +1,147 @@
...
+        inline-size 300ms ease;
+    z-index: 5;
+}
+
+.fl-qr-code-floating-snippet.is-open {
+    block-size: 19.188rem;
+    inline-size: 22.5rem;
+}
+
+@starting-style {
```
maureenlholland, line 33: **non-blocking, maintenance**: what is this used for? I don't see what effect it's having

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
maureenlholland: also needs accessible names for "show" and "hide" states

springfield/cms/models/snippets.py
```diff
@@ -274,3 +275,60 @@ def get_preview_template(self, request, mode_name):
...
+        return f"{remove_tags(richtext(self.heading))} – {self.locale}"
+
+    def get_preview_context(self, request, mode_name):
+        context = super().get_preview_context(request, mode_name)
+        if self.image:
+            context["qr"] = {"type": "image", "value": self.image.file.url, "open": self.default_open}
+        elif self.url:
+            context["qr"] = {"type": "qr", "value": self.url, "open": self.default_open}
+        else:
+            context["qr"] = {"type": "qr", "value": "", "open": self.default_open}
```
stevejalim: What gets rendered as the QR code in this instance? Looks like the template doesn't render anything so why have a key-value pairs in in the dict?

maureenlholland (dismissed):
FE review

The only definite blocker for me is having a keyboard-accessible button with accessible names for the show/hide icons

The need for styling on light backgrounds is potentially blocking

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+      <span class="fl-icon fl-icon-device-mobile"></span>
+      {{ value.heading|striptags }}
+    </div>
+    <div class="fl-qr-code-close fl-qr-code-snippet-close">
+      {% if qr.open %}
+        <span class="fl-icon fl-icon-subtract"></span>
+      {% else %}
+        <span class="fl-icon fl-icon-add"></span>
+      {% endif %}
+    </div>
```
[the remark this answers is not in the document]
maureenlholland: redundant: https://github.com/mozmeao/springfield/pull/1287#discussion_r3116833861

media/css/cms/flare26-floating-qr-overlay.css
```diff
@@ -0,0 +1,147 @@
...
+    block-size: 180px;
+    inline-size: 180px;
+    transform: translateX(-8px);
+}
+
+.fl-qr-code-floating-snippet-kit svg rect {
+    fill: transparent;
+}
+
+@media (--viewport-below-md) {
```
maureenlholland: **blocking**: We don't hide the closed version of the snippet, so clicking this just makes the element disappear. We should hide both open and closed versions.

Image

springfield/cms/models/pages.py
```diff
@@ -381,16 +383,64 @@ class ThanksPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
         help_text="If true, a floating QR code snippet will be displayed on the page.",
     )
 
+    show_floating_qr_code_snippet = models.BooleanField(
+        default=False,
+        help_text="If true, an updated floating QR code snippet will be displayed on the page.",
+    )
+
+    override_url = models.CharField(
```
stevejalim: blocking: This should be a URLField so we can have URL validation (and it automatically adds a max_length attr)

springfield/cms/models/pages.py
```diff
@@ -818,13 +868,50 @@ class FreeFormPage2026(UTMParamsMixin, AbstractSpringfieldCMSPage):
         default=False,
         help_text="If true, a floating QR code snippet will be displayed on the page.",
     )
+    show_floating_qr_code_snippet = models.BooleanField(
+        default=False,
+        help_text="If true, an updated floating QR code snippet will be displayed on the page.",
+    )
+
+    override_url = models.CharField(
```
stevejalim: Again should be `URLField`, I think

Given that we're repeating ourselves here, I wonder if a mixin class would be cleaner here

springfield/cms/models/pages.py
```diff
@@ -891,12 +991,49 @@ class WhatsNewPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
```
stevejalim: Let's use a `URLField` + another vote for using a mixing

springfield/cms/models/pages.py
```diff
@@ -944,13 +1094,50 @@ class WhatsNewPage2026(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
```
stevejalim: Let's use a `URLField` + another vote for using a mixin

stevejalim (changes requested):
There's still some work to do here, in addiiton to the copilot comments. I'm strongly for a mixin to avoid needing to repeat very specific code additions, plus we must watch out for all CharFields lacking max_length

I'm happy to take it on, but not going to rush it out of the door.

maureenlholland:
The token color on our theme shadow var appears to be missing. I don't want to touch anything global right now, so I've hardcoded the Figma soft purple shadow on this QR component for now
Screenshot 2026-04-21 at 1 15 36 PM

maureenlholland:
> 16.36 ?: (staticfiles.W004) The directory '/app/assets' in the STATICFILES_DIRS setting does not exist.
16.74 CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0073_merge_20260420_1609, 0074_alter_smartwindowpage_mobile_message in cms).
16.74 To fix them run 'python manage.py makemigrations --merge'

Tests appear to be failing due to migration issues

media/js/cms/new.es6.js
```diff
@@ -420,7 +418,11 @@ if (typeof window.cms === 'undefined') {
 
     function initQRCodeSnippet() {
         const COOKIE_ID = 'moz-qr-snippet-dismissed';
-        const qrCodeSnippetEl = document.querySelector('.fl-qr-code-snippet');
+        const oldSnippet = document.querySelector('.fl-qr-code-snippet');
```
maureenlholland: **non-blocking, maintenance**: any classes used as JS selectors should be clearly marked with `js-` prefix: https://www-demo1.springfield.moz.works/pattern-library/#class-prefixes

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+#}
+
+{% if qr %}
+<aside class="fl-qr-code-floating-snippet fl-qr-code-snippet-closable {% if qr.open and not qr.hide %}is-open{% endif %}">
+  <div class="fl-qr-code-floating-header">
+    <div class="fl-qr-code-prompt">
+      <span class="fl-icon fl-icon-device-mobile"></span>
+      {{ value.heading|striptags }}
+    </div>
+    <button class="fl-qr-code-snippet-close fl-qr-code-floating-button">
```
maureenlholland, line 14: **non-blocking, maintenance**: we have a mix of CSS classes from the old QR code & new QR code styles in use here, some are also in use for JS. We should follow-up to cleanly separate old from new (new styles should only rely on classes from the new CSS file)

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
maureenlholland: captured here: https://mozilla-hub.atlassian.net/browse/WT-1041?focusedCommentId=1384285

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
maureenlholland: captured here: https://mozilla-hub.atlassian.net/browse/WT-1041?focusedCommentId=1384285

media/css/cms/flare26-floating-qr-overlay.css
maureenlholland: captured here: https://mozilla-hub.atlassian.net/browse/WT-1041?focusedCommentId=1384285

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
maureenlholland: captured here: https://mozilla-hub.atlassian.net/browse/WT-1041?focusedCommentId=1384285

media/js/cms/new.es6.js
maureenlholland: captured here: https://mozilla-hub.atlassian.net/browse/WT-1041?focusedCommentId=1384285

springfield/cms/models/snippets.py
bluewave41: This was me. Complaints of circular redundancies I didn't have time to dig deeper into.

springfield/cms/models/pages.py
```diff
@@ -381,16 +383,64 @@ class ThanksPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
         help_text="If true, a floating QR code snippet will be displayed on the page.",
     )
 
+    show_floating_qr_code_snippet = models.BooleanField(
+        default=False,
+        help_text="If true, an updated floating QR code snippet will be displayed on the page.",
+    )
+
+    override_url = models.CharField(
+        blank=True, help_text="Override the snippet URL. A QR code will be generated from this. Not used if an override image is set."
```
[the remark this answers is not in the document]
bluewave41: There's no way this is true? None of my manual tests would've worked in this case. There are other cases where CharFields don't have max_lengths and nothing broke for them.

Either way these are being changed to URLField

springfield/cms/models/snippets.py
bluewave41: Pretty sure Claude added this. I'm fairly sure I was just returning None as an else case.

springfield/cms/models/pages.py
```diff
@@ -381,16 +383,64 @@ class ThanksPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
+            heading="QR Code Floating Button",
+            classname="collapsed",
+        ),
+    ]
+
+    override_translatable_fields = [
+        SynchronizedField("override_url"),
+        SynchronizedField("override_image"),
+        SynchronizedField("override_default_open"),
     ]
```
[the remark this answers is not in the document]
bluewave41: Don't think this is applicable. The old QRCodeSnippet didn't bother with this. See snippets.py:260

springfield/cms/models/snippets.py
maribedran: It's a circular import issue. The template tags module imports models.

springfield/cms/models/snippets.py
maribedran: If we don't remove the tags, you'd see the full HTML on the listing page.

media/css/cms/flare26-floating-qr-overlay.css
bluewave41: Probably nothing. I started with a copy paste of the old qr code styling and adapted from there.

Removing it still works the same. Will remove.

springfield/cms/models/pages.py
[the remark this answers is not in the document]
stevejalim: This all surprised me, so I dug into it a bit. TIL time!

Looking at the source, `max_length` is only needed for DB backends that can't handle unlimited-length charfields - see https://github.com/django/django/blob/8ee73415270a1a54daaec9bb529ad82c6f7a6d4c/django/db/models/fields/__init__.py#L1226-L1241

Handily for us, both sqlite (for local dev and for springfield-test) and postgres (for production and demos) _can_ handle it: 
* https://github.com/django/django/blob/a77be30a4346d111beaf366dcee4934734458a48/django/db/backends/postgresql/features.py#L187
* https://github.com/django/django/blob/a77be30a4346d111beaf366dcee4934734458a48/django/db/backends/sqlite3/features.py#L58

Seems this support landed four (!) years ago https://github.com/django/django/commit/7eee1dca420ee7c4451d24cd32fa08aed0dcbb36 - so Copilot must have been looking at a stale textbook, metaphorically.

cc @maribedran

springfield/cms/models/pages.py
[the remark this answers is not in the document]
stevejalim: The difference here might be that a snippet is closer to a Django model than a Page, so has a different superclass than `SpringfieldAbstractCMSPage`. I'm happy to make this tweak, as it makes sense to me - otherwise we might end up with needing to manually copyover the `override_*` values to all 14 translations of a page

springfield/cms/models/pages.py
```diff
@@ -944,13 +1094,50 @@ class WhatsNewPage2026(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
```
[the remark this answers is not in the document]
stevejalim: Again - probably worth doing. Will add it in

springfield/cms/models/pages.py
```diff
@@ -818,13 +868,50 @@ class FreeFormPage2026(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
```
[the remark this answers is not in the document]
stevejalim: Again - likely worth doing

springfield/cms/models/pages.py
```diff
@@ -891,12 +991,49 @@ class WhatsNewPage(UTMParamsMixin, AbstractSpringfieldCMSPage):
...
```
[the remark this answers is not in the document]
stevejalim: Again - likely worth doing

stevejalim:
@maribedran @bluewave41 @maureenlholland FYI I am rebasing this on latest `main` now before doing any other tweaks we need

springfield/cms/models/pages.py
```diff
@@ -191,6 +192,75 @@ def get_context(self, request, *args, **kwargs):
         return context
 
 
+class QRCodeFloatingSnippetMixin(models.Model):
```
stevejalim: Wondering if this should be extending `AbstractPage` so we can extend `override_translatable_fields` to include values from other bases too

stevejalim: yeah, the MRO for pages like `ThanksPage(UTMParamsMixin, QRCodeFloatingSnippetMixin, AbstractSpringfieldCMSPage)` means Python finds `QRCodeFloatingSnippetMixin.override_translatable_fields` first and stops, so `AbstractSpringfieldCMSPage`'s `SynchronizedField("slug")` is silently dropped

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -0,0 +1,34 @@
...
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+{% if value and value.qr %}
+<aside class="fl-qr-code-floating-snippet fl-qr-code-snippet-closable {% if value.qr.open %}is-open{% endif %}">
+  <div class="fl-qr-code-floating-header">
+    <div class="fl-qr-code-prompt">
+      <span class="fl-icon fl-icon-device-mobile"></span>
+      {{ value.heading|striptags }}
```
maribedran: The `striptags` will remove all HTML tags, which defeats the purpose of having a rich text field. The correct filter here is `|richtext`

springfield/cms/tests/test_qr_code_floating_button.py
```diff
@@ -0,0 +1,394 @@
...
+    soup = BeautifulSoup(_serve_page(page, rf).content, "html.parser")
+    aside = _get_floating_qr_aside(soup)
+    assert aside, "Floating QR <aside> should render when show_floating_qr_code_snippet=True"
+    assert "Get Firefox on your phone" in aside.get_text()
+
+
+@pytest.mark.parametrize(
+    "get_page_fn",
+    PAGES_WITH_FLOATING_QR + [get_whats_new_page_with_qr_snippet],
+)
```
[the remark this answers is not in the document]
stevejalim: Fixed

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
stevejalim: Fixed and pushed

stevejalim (approved):

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
kkellydesign: I didn't go nuts prefixing all JS-used classes, but I did add classes for the two parent selectors.

kkellydesign:
> The token color on our theme shadow var appears to be missing. I don't want to touch anything global right now, so I've hardcoded the Figma soft purple shadow on this QR component for now Screenshot 2026-04-21 at 1 15 36 PM

I used another box shadow variable that does the job already: 
```box-shadow: var(--notification-box-shadow);```

kkellydesign (approved):
I'm pencils-down here.

PR #1290, opened 2026-04-21
Prefers reduced motion fixes

kkellydesign:
## One-line summary
Prefers Reduced Motion improvements

* All `autoplay` videos are paused on load when reduced motion is preferred.
* The testimonial carousel design looks better when the animation is disabled
* (unrelated), I added `unsupported` to the display of the mobile conditional text. 

Before: 
image

After: 
Screenshot 2026-04-21 at 10 19 32 AM

## Testing
Screenshot 2026-04-21 at 10 20 03 AM

### DIFF
--- a/springfield/cms/templates/cms/smart_window_page.html
+++ b/springfield/cms/templates/cms/smart_window_page.html
@@ -185,7 +185,7 @@
                 </div>
               </include:conditional-display>
               {# Show mobile message for mobile platforms #}
-              <include:conditional-display platform_conditions="{{ ['android', 'ios'] }}">
+              <include:conditional-display platform_conditions="{{ ['android', 'ios', 'unsupported'] }}">
                   {{ page.mobile_message|richtext }}
               </include:conditional-display>
             </include:conditional-display>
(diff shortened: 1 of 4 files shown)


springfield/cms/templates/cms/smart_window_page.html
```diff
@@ -185,7 +185,7 @@ <h1 class="fl-heading fl-smart-window-heading-success">{{ page.thank_you_heading
                 </div>
               </include:conditional-display>
               {# Show mobile message for mobile platforms #}
-              <include:conditional-display platform_conditions="{{ ['android', 'ios'] }}">
+              <include:conditional-display platform_conditions="{{ ['android', 'ios', 'unsupported'] }}">
```
stevejalim, line 188: Does this reveal an extra section that will need QA before we ship it to prod?

stevejalim (approved):
r+wc

springfield/cms/templates/cms/smart_window_page.html
kkellydesign: It's a fallback message for edge-cases where we can't sniff out the platform. It will display the same message as the mobile devices: 

"This experience is only available on desktop. Please open this page on your computer."

It's not a perfect message for that case, because we don't know that they're _not_ on desktop. But linux, windows, and mac are covered already.

PR #1292, opened 2026-04-21
Fix Some Fallback Locale URL Issues

dchukhin:
## One-line summary
This pull request fixes an inconsistency in the way that fallback locales were serving pages.

This pull request fixes a few issues with fallback pages:
1.  when a URL has the `prefer_cms` decorator, the user requests an alias locale, the alias locale root page exists but the requested page does not exist in the alias locale, and the requested page exists for a fallback locale and also exists as a static page, make sure we serve the fallback page content, not the static page. An example of this bug is the current behavior of https://www.firefox.com/es-AR/features/bookmarks/ - this serves the static view, not the es-MX Page.
2. for extra clarity, we make sure that a `<link rel="canonical">` tag is rendered even for pages with a `noindex` value of `True`. Since `noindex` and `canonical` provide different information to crawlers, we now provide both pieces of information (previously `noindex` pages omitted both). This is probably not hugely significant, outside of 2 noindex pages having identical content, but it does better fit the original spec: "Every page that returns a 200 response includes rel="canonical"."

## Issue / Bugzilla link
[WT-854](https://mozilla-hub.atlassian.net/browse/WT-854)

## Testing
1. load http://localhost:8000/es-AR/features/bookmarks/, and make sure it serves the http://localhost:8000/es-MX/features/bookmarks/ content
2. find a non-indexable page (for example, http://localhost:8000/pt-PT/thanks/), and make sure that:
  a. it is not indexable (`<meta name="robots" content="noindex,follow">`)
  b. it has a canonical tag (`<link rel="canonical" href="https://www.firefox.com/pt-BR/thanks/">`)


kkellydesign (approved):
I can confirm that this PR does what it says it does. I'd like to request an additional review from @stevejalim on this one though, as he's more familiar with the potential 'gotchas' regarding this redirect code than I am. 

So, a cautious approval from me.

stevejalim (approved):
Also LGTM and the testing checks out.

PR #1298, opened 2026-04-23
mark instructions field as unrequired

kkellydesign:


stevejalim (approved):

stevejalim:
IIRC Django prefers us not to set `null=True` on TextField (which is the base for RichTextField)

https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.Field.null

so made a tweak before merging

PR #1299, opened 2026-04-23
fix: disable no-longer-needed data migration that now breaks DB export

stevejalim:
Should stop https://mozilla.sentry.io/issues/7421933611/events/latest/?project=4508842391633920&query=is%3Aunresolved&referrer=latest-event happening during a DB export run


stevejalim:
No longer needed

PR #1305, opened 2026-04-28
[WT-1030] Notification spacing

maribedran:
## One-line summary

CSS improvements for notification and banner

- Reduce top padding when a page starts with a notification block
- Add missing spacing to banner block on mobile

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1021


media/css/cms/flare26-notification.css
maureenlholland: Is there a pre-existing regression here? comment indicates it used to be negative margin

edit: I'd like to remove code from this component instead: https://github.com/mozmeao/springfield/pull/1305/changes#r3159707095

media/css/cms/flare26-notification.css
```diff
@@ -15,6 +15,13 @@
     z-index: 10;
 }
 
+.fl-split-page-upper:has(.fl-notification-wrapper:first-child),
```
maureenlholland: I think this is overcomplicating things. The restriction to split-page-upper does not address spacing issues on http://localhost:8000/en-US/whatsnew/146/

Image

Could we take a simplifying approach?
- remove margin from notification and padding and margin-block-end from notification-wrapper, use margin-inline to provide default inline buffer on smaller screens
   - It's hard to reason about compounding spacing like this, unclear what selector is responsible. If there is spacing required, I would argue it belongs on the wrapper only (the component itself shouldn't have any outside layout concerns)

Image

Image

   - we can use gap on notification wrapper to conditionally add spacing between notifications if needed

Some very light testing on the no-block-spacing approach:

Image

Image

media/css/cms/flare26-notification.css
```diff
@@ -15,6 +15,13 @@
...
+.fl-split-page-upper:has(.conditional-display:first-child):has(
+        .fl-notification-wrapper
+    ) {
+    padding-block-start: 0;
+}
+
 /* When a notification immediately precedes an intro or home intro component, tighten up the space between them */
```
maureenlholland: This appears to already be tightening spacing for certain combinations, it might be a better place to add the conditional display selectors if needed

media/css/cms/flare26-notification.css
maureenlholland, line 44: **non-blocking, maintenance:** we have position relative here but z-index on a different selector, is this dead code that can be removed?

media/css/cms/flare26-banner.css
```diff
@@ -219,6 +219,7 @@
 .fl-banner-layout {
     align-items: center;
     display: grid;
+    gap: var(--token-layout-xs);
```
maureenlholland, line 222: resolves this issue ✅ 
Image

maureenlholland (dismissed):
r+ on the banner layout fix

I have thoughts on the notification styling - primarily, I'd like this to be simpler to reason about. I was looking for use cases in the pattern library but are there better places to look to understand all possible effects the code-removing approach has?

media/css/cms/flare26-notification.css
maribedran: This used to handle the spacing between the notification and the next block on the 2025 version. I think the blocks there had a more uniform spacing.

maribedran: Removed the z-index above. It didn't seem to be doing anything.

maribedran:
@maureenlholland I simplified the spacing logic to:
- remove start padding from the split page upper when the first element is a notification
- reduce start padding on the intro block when preceded by a notification

media/css/cms/flare26-notification.css
kkellydesign: ok, I've implemented the following: 
* No more margin on notification or wrappers. 
* horizontal padding has been added to the wrapper to allow for correct mobile gutters. 
* For the rare cases when we'd have two stacked notifications, It's an unnecessary complexity to have an added 'wrapper' component at the CMS level, so the component itself contains the wrapper div. So I did this, to account for those cases: 

```
.fl-notification-wrapper + .fl-notification-wrapper {
    margin-block-start: var(--token-spacing-lg);
}
```
* The "intro" component currently has larger vertical padding than most sections, which is why Mari had a special case there. Rather than add this complexity to the notification CSS, I adjusted the intro to use the normal section padding _by default_, and then overrode it with the larger padding only when it is the first component on the page. This lets it shrink to a reasonable size when there's a notification, or anything else, above it. (I'd argue that an "intro" used elsewhere is an abuse of the pattern, but now it's slightly more flexible without harming anything.)
```
.fl-intro {
    padding-block: var(--fl-section-v-padding);
}

.fl-intro:first-child {
    padding-block: var(--token-layout-md);
}

```

Good suggestions, thanks @maureenlholland!

kkellydesign: related: I left a todo to get rid of the top padding on `.fl-split-page-upper` globally. We're overriding that in several places.

kkellydesign (approved):
works for me 🚢

PR #1308, opened 2026-04-29
fix: avoid blow-up when DB is in maintenance mode

stevejalim:
In response to https://mozilla.sentry.io/issues/7449029422/

When the DB is unavailable, we should survive without it wherever possible

### DIFF
--- a/lib/l10n_utils/__init__.py
+++ b/lib/l10n_utils/__init__.py
@@ -147,7 +147,7 @@ def render(request, template, context=None, ftl_files=None, activation_files=Non
     # if `active_locales` is given use it as the full list of active translations
     translations = []
 
-    if is_cms_page and request._locales_available_via_cms:
+    if is_cms_page and hasattr(request, "_locales_available_via_cms") and request._locales_available_via_cms:
         translations = request._locales_available_via_cms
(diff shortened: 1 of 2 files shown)


stevejalim (commented):
Added a regression test in lib/l10n_utils/tests/test_base.py (test_cms_page_without_locales_available_does_not_raise) covering the case where request.is_cms_page = True but _locales_available_via_cms is never set. Verifies render() does not raise and calls through to django_render.

lib/l10n_utils/__init__.py
```diff
@@ -147,7 +147,7 @@ def render(request, template, context=None, ftl_files=None, activation_files=Non
     # if `active_locales` is given use it as the full list of active translations
     translations = []
 
-    if is_cms_page and request._locales_available_via_cms:
+    if is_cms_page and hasattr(request, "_locales_available_via_cms") and request._locales_available_via_cms:
```
dchukhin, line 153: It may be helpful to log something in the case that `is_cms_page` is True and `request` doesn't have a `_locales_available_via_cms` attribute, so we have more information about something not working properly.

dchukhin (approved):
👍🏻

PR #1310, opened 2026-04-30
Prevent GitHub Actions on forks (WT-1124)

janriokrause:
# Summary

Add `if: github.repository == 'mozmeao/springfield'` guards to GitHub Actions workflows that were missing them, so none of them run on forks.

# Background

We observed noisy emails from CDN tests on forks. The `cdn-tests` job is correctly guarded with `if: github.repository == 'mozmeao/springfield'`, but `notify-of-test-run-completion` uses `if: always()` without the same guard. That completion job still executes on forks (including `workflow_dispatch`), so the workflow appears to run on forks even though the main `cdn-tests` job is skipped. Adding a repository condition to the completion job prevents any fork execution.

# Problem

Several GitHub Actions workflows were triggering on forks, either on a schedule, on push, or when manually dispatched by a fork owner. This caused two classes of issues:

* Failure emails: jobs that use secrets (Slack, SauceLabs, Heroku) fail silently on forks because secrets are not passed, generating noise for maintainers.
* Unintended side-effects: `routine_l10n_obsolete.yml` uses `GITHUB_TOKEN` which is available on forks, meaning it would actually create l10n cleanup issues on fork repositories.

# Changes

* `cdn_tests.yml`: The `notify-of-test-run-completion` job had `if: always()` without a repo check. Because `if: always()` overrides skipped upstream dependencies, this job ran on forks even when the `cdn-tests` job was correctly skipped, causing failure emails from missing secrets.
* `routine_l10n_obsolete.yml`: This is a scheduled workflow that uses `GITHUB_TOKEN`, which is automatically available on forks. Without the guard it would open GitHub issues on fork repositories on a bimonthly schedule.
* `demo_deploy.yml`: Triggered on pushes to `demo/*` branches. Any fork with such a branch would trigger a Heroku deploy attempt, fail on the missing `HEROKU_API_KEY` secret, and send failure email notifications.
* `integration_tests.yml`: No job had a repo check. Being `workflow_dispatch`-only lowers the risk, but running it on a fork would fail due to missing secrets and generate failure emails. Added guards to all four jobs, using `if: always() && ...` where `always()` was already present.
* `post_deploy_asset_check.yml`: Same situation as `integration_tests.yml`: `workflow_dispatch`-only, no repo checks, missing secrets would cause failures and emails.
* `db_sync.yml`: Same situation: `workflow_dispatch`-only with no repo check.
* `unit_tests.yml` and `fluent_linter.yml`: These don't use secrets and wouldn't fail on forks, but they would run unnecessarily on fork pushes to `main`. Added guards for consistency.

# Not changed

`a11y_tests.yml`, `download_tests.yml`, `build-and-push.yml`, `generate-gh-release-note.yml`, and `send_firefox_fluent_strings_to_l10n_org.yml` were already correctly guarded and required no changes.

### DIFF
--- a/.github/workflows/cdn_tests.yml
+++ b/.github/workflows/cdn_tests.yml
@@ -91,7 +91,7 @@ jobs:
           if-no-files-found: ignore  # this avoids a false "Warning" if there were no issues
 
   notify-of-test-run-completion:
-    if: always()
+    if: always() && github.repository == 'mozmeao/springfield'
     needs: cdn-tests
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/db_sync.yml
+++ b/.github/workflows/db_sync.yml
@@ -34,6 +34,7 @@ jobs:
   sync:
     name: "Sync ${{ inputs.source }} → ${{ inputs.target }}"
     runs-on: ubuntu-latest
+    if: github.repository == 'mozmeao/springfield'
     environment: db-sync
     concurrency:
       group: db-sync-${{ inputs.target }}
--- a/.github/workflows/demo_deploy.yml
+++ b/.github/workflows/demo_deploy.yml
@@ -9,6 +9,7 @@ jobs:
   deploy:
     name: Deploy Heroku
     runs-on: ubuntu-latest
+    if: github.repository == 'mozmeao/springfield'
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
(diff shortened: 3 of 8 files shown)


janriokrause:
Similar to mozilla/bedrock#17152.

stevejalim (approved):
LGTM - thanks. I wondered for a moment about the integration-tests lockdown, but in reality no one with a fork will have an equivalent private setup for integration tests

PR #1324, opened 2026-05-06
Fix missing CSS bundles in prod build

maureenlholland:
## One-line summary

Fixes errors in production build, i.e. "Post-processing 'css/cms/pages/flare26-cookie-settings.css' failed!"

NOTE: this is **not** a production bug. It's an issue with our `main` branch that we need to resolve before next prod deploy

As noted in the CSS comments, the flare26-tokens import is needed in standalone bundles for postcss-custom-media to resolve `@custom-media` variables.

This works as expected with CSS file imports when run through webpack, but the files are not typical relative paths, so all imports in the pages folder are failing to be found on staticfile production builds.

Moving to a typical relative file path structure (../flare26-tokens) creates an error with the custom webpack loader.

Adding the required files into the static bundles instead of importing through the CSS file is the simplest CSS fix to unblock deploys. We may want to revisit to work more seamlessly with the flare-import-anywhere webpack loader.

## Issue / Bugzilla link
https://github.com/mozmeao/springfield/actions/runs/25414000246/job/74541466354

## Testing
Run in [Prod Mode](https://mozmeao.github.io/platform-docs/getting-started/running-locally/#prod-mode) and confirm no errors
Smoke check one or two of the affected pages:
- http://localhost:8000/en-US/
- http://127.0.0.1:8000/en-US/firefox/150.0.2/releasenotes/ (confirm larger screens show three columns on pre-footer notes)

Successful integration test run: ❌ ([new playwright test failures after successful build](https://github.com/mozmeao/springfield/actions/runs/25430003245))

### DIFF
--- a/media/css/cms/pages/flare26-channels.css
+++ b/media/css/cms/pages/flare26-channels.css
@@ -11,9 +11,6 @@
     what should be refactored away here.
 */
 
-/* Required for postcss-custom-media to resolve @custom-media variables in standalone bundles */
-@import 'flare26-tokens.css';
-
 @layer page {
     .fl-brand-icon,
     .fl-brand-icon-wrap img {
--- a/media/css/cms/pages/flare26-cookie-settings.css
+++ b/media/css/cms/pages/flare26-cookie-settings.css
@@ -4,9 +4,6 @@
 * file, You can obtain one at https://mozilla.org/MPL/2.0/.
 */
 
-/* Required for postcss-custom-media to resolve @custom-media variables in standalone bundles */
-@import 'flare26-tokens.css';
-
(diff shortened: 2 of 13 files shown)


maureenlholland:
Nope, this approach doesn't work. It fixes the build errors but the custom media queries don't take effect. The import needs to be directly in the CSS file

i.e. dev build is correctly displaying 3 columns on larger screens: https://www-dev.springfield.moz.works/en-US/firefox/150.0.2/releasenotes/ but this PR is not

maureenlholland:
quick fix for build causes more issues with CSS, closing as tried but not viable

PR #1331, opened 2026-05-06
button wrap revisions (1)

kkellydesign:
## One-line summary
ref: https://mozilla-hub.atlassian.net/browse/WT-1135?focusedCommentId=1416410

Intro, Media+Content, and Banner all needed this fix. 

Also note that “tirtiary” buttons don’t actually exist in the ‘26 design system – it’s an artifact from the ‘25 styles. We’ll clean that up when we remove the ‘25 styles. For now, it just uses the default. ‘Secondary’, 'Ghost’ and 'Link” all work as expected. 

Screenshot 2026-05-06 at 3 44 53 PM
Screenshot 2026-05-06 at 3 44 45 PM
Screenshot 2026-05-06 at 3 44 42 PM


media/css/cms/components/flare26-banner.css
```diff
@@ -318,6 +319,7 @@
     }
 
     .fl-banner-content.fl-banner-content-left .fl-buttons {
+        align-items: start;
         justify-content: left;
```
janriokrause:
```suggestion
        justify-content: flex-start;
```

Non-blocking suggestion: I know you didn't touch these lines of code, but it could be useful to use logical/RTL‑safe alignment for the left‑aligned button group in the banner. `justify-content: left` is not logical and can behave oddly in RTL locales. `flex-start` (or `start` if you want logical) is more consistent. See [[…]/templates/pattern-library/docs/css/css_localization_md.md#rtl](https://github.com/mozmeao/springfield/blob/refs/heads/main/springfield/cms/templates/pattern-library/docs/css/css_localization_md.md#rtl)

janriokrause (approved):
r+ Looks good to me! I added a small suggestion 😇

media/css/cms/components/flare26-banner.css
kkellydesign: Good catch, thanks. I'm not even sure `left` is valid code. I opted for `start` instead of `flex-start`, so marking this one as solved.

PR #1332, opened 2026-05-06
Add analytics attributes to all rich text links

maribedran:
## One-line summary

Add `data-cta-uid`, `data-cta-position`, and `data-cta-text` attributes to links in rich text fields and blocks.

- Replace the rich text field and block classes with custom `RichTextBlock` and `RichTextField` classes, which insert a random `uid` attribute in each `< a ' tag in rich text values when absent from the original tags.
- Create Wagtail hooks to process link attributes between the database values and the format needed for the editing UI.
- Insert `data-cta-position` and `data-cta-text` attributes in the custom `richtext` template filter
- Make sure that all block templates use the `|richtext` template filter when rendering rich text blocks

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-752

## Testing

- Edit any CMS page and insert internal and external links in any rich text field
  - Check that both the preview and the published page render the links with the analytics attributes
- Translate the page and check that the `data-cta-uid` attribute is the same on the translated page (the `data-cta-position` may change if the translation changes the order of the words, and the `data-cta-text` will show the translated text)
- Edit an existing page and check that the attributes are added to all links when the page is saved (the http://localhost:8000/en-US/landing/w10-choose/ page has both internal and external links)


springfield/cms/templatetags/cms_tags.py
```diff
@@ -234,7 +239,7 @@ def richtext(context, value: str) -> str:
         }
         optional_attributes = {
             "data-cta-uid": uid,
-            "data-cta-position": "-".join([context.get("block_position", ""), "fxa-link"]),
+            "data-cta-position": ".".join([context.get("block_position", ""), "fxa-link"]),
```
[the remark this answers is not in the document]
stevejalim: Also, will changing from `-` to `.` break anything (eg downstream analytics - Claude suggests the `-` format is used in FXA uids

springfield/cms/templates/pattern-library/components/intro/intro.yaml
```diff
@@ -24,7 +24,6 @@ context:
           theme_class: 'button-secondary'
       orientation: horizontal
       center_class: is-center
-    fine_print: Fine print
```
stevejalim, line 27: Why is `fine_print` being dropped here and in a few other places? If it's not strictly relevant to this PR, I think it's better to land it as its own PR.

springfield/cms/wagtail_hooks.py
```diff
@@ -262,3 +267,113 @@ def register_firefox_logo_feature(features):
...
+      get a fresh UUID on first save.
+    """
+    # Render-time: add data-uid to <a> tags in the output HTML.
+    features.register_link_type(ExternalLinkHandler)
+    features.register_link_type(UIDPageLinkHandler)
+
+    # Editor round-trip: override the "link" contentstate converter so that
+    # uid survives the DB → Draft.js → DB cycle. Our hook runs after
+    # wagtail.admin's register_core_features (springfield.cms is later in
+    # INSTALLED_APPS), so this registration replaces the built-in one.
```
stevejalim, line 421: This dependency on ordering is a bit fragile - it'd be good to at least comment in `settings.base` about the ordering expectations of INSTALLED_APPS and to point to this hook

springfield/cms/rich_text.py
```diff
@@ -0,0 +1,38 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import uuid
+
+from bs4 import BeautifulSoup
+from wagtail.blocks import RichTextBlock as WagtailRichTextBlock
+from wagtail.fields import RichTextField as WagtailRichTextField
+
+
+def inject_link_uids(html):
```
stevejalim, line 12: Could we have unit tests for `inject_link_uids` at the least - ideally also for our custom block and field in this module, please?

springfield/cms/wagtail_hooks.py
```diff
@@ -262,3 +267,113 @@ def register_firefox_logo_feature(features):
     # Add the feature to the default features list to make it available
     # on rich text fields that do not specify an explicit 'features' list
     features.default_features.append(feature_name)
+
+
+class ExternalLinkHandler(LinkHandler):
```
stevejalim, line 274: I think we need unit tests for all of these methods, ideally

springfield/cms/templates/cms/blocks/media-content.html
```diff
@@ -31,7 +32,12 @@
       {# In case a block with a heading is included, increment the block level #}
       {# Add 1 because this block already has a heading #}
       {% with block_level = block_level + 1 if block_level else 3 %}
-        {% include_block block %}
+        {% if block.block_type == "rich_text" %}
+          {# Make sure to use our custom richtext filter to add UTM parameters and other attributes #}
+          {{ block.value|richtext }}
```
stevejalim: I get that it's less likely to lead to oversight to replace the default `richtext` filter with out UID-setting one, but also it does possibly make it harder for future generations to know that what looks like the regular filter actually does extra stuff. Rather than switch to `|springfield_richtext` or similar, let's add some docs to the github.com/mozmeao/platform-docs, I think

springfield/cms/tests/test_templatetags.py
```diff
@@ -146,7 +146,7 @@ def test_richtext_parses_fxa_tag():
     expected_html = f"""<p>Paragraph with <a
         class="js-fxa-cta-link js-fxa-product-button fxa-link"
         data-action="{settings.FXA_ENDPOINT}"
-        data-cta-position="block-1-fxa-link"
+        data-cta-position="block-1.fxa-link"
```
stevejalim: So yeah, here's the FXA-related use of the old format - will this change break analytics or anyting else?

stevejalim (approved):
Thanks! r+wc -- have left some comments to be addressed and the robot ones appear useful, too

(I don't think there's an XSS angle being added here, though 🎉 )

springfield/cms/templatetags/cms_tags.py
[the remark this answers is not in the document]
maribedran: The important information for GA is the UID; the position and text are auxiliary to make it easier to locate the links on the page.

springfield/cms/templates/pattern-library/components/intro/intro.yaml
maribedran: It's not used by the block template and has no associated styles. It could be easily replaced with a `<small` tag if anyone wanted to use it.

springfield/cms/wagtail_hooks.py
```diff
@@ -262,3 +267,113 @@ def register_firefox_logo_feature(features):
...
+    # INSTALLED_APPS), so this registration replaces the built-in one.
+    features.register_converter_rule(
+        "contentstate",
+        "link",
+        {
+            "from_database_format": {
+                "a[href]": UIDExternalLinkElementHandler("LINK"),
+                'a[linktype="page"]': UIDPageLinkElementHandler("LINK"),
+            },
+            "to_database_format": {"entity_decorators": {"LINK": uid_link_entity}},
```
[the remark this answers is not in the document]
maribedran: The document link handler is registered separately by Wagtail and isn't affected by this. The other link types (phone, mail, and anchor) all use `href`, so they're all covered.

springfield/cms/templates/cms/blocks/media-content.html
maribedran: I added a template to the rich text block to make sure that the filter is always called

maribedran:
@stevejalim I double checked and the implementation didn't affect any existing link behavior. I addressed the comments and added the same logic to Wagtail document links. The feature wasn't even registered. Now we can add document links inline.

PR #1334, opened 2026-05-07
feat: Extend WNP evergreen redirect to support Nightly, Developer, and Beta

stevejalim:
This changeset adds support for ensuring users of Nightly and Developer will be shown a channel-specific, CMS-backed 'evergreen' WNP on www.firefox.com, if one exists in their locale. It also supports this behaviour for Beta, but we may not use it, because currently we don't offer a custom WNP for Beta, we just use a generic fallback static WNP.

## Points to note

`_get_general_wnp_redirect` (now `_get_evergreen_wnp_redirect`) previously always redirected to `/whatsnew/general/`. It now detects the channel from the version string and redirects to the appropriate CMS evergreen WNP:

(using 152 as an example version number, but valid for all versions)

* `/<locale>/whatsnew/152.0a1` --> `/<locale>/whatsnew/nightly/`
* `/<locale>/whatsnew/152.0a2` --> `/<locale>/whatsnew/developer/`
* `/<locale>/whatsnew/152.0beta` → `/<locale>/whatsnew/beta/`
* `/<locale>/whatsnew/152/` --> `/<locale>/whatsnew/general/` (unchanged behaviour)

If no channel-specific CMS page exists for the locale, the view falls back to the existing static Django/Fluent evergreen templates.

Tests have been updated to cover all three new channels, their static fallbacks, and the  no-redirect-loop guarantee for the new slugs.

## Issue / Bugzilla link

Jira: https://mozilla-hub.atlassian.net/browse/WT-1121

### Setup

1. Run the dev server
2. Log into the Wagtail admin at /cms/
3. Navigate via Pages to the What's New section (the `WhatsNewIndexPage` at `/whatsnew/`)

### Create the test CMS pages

Add four child pages under the What's New index. For each, choose What's New 2026 Page as the page type, add any placeholder content (minimal is fine), then publish.

- Slug `general`, title e.g. "WNP General" — may already exist, skip if so
- Slug `nightly`, title e.g. "WNP Nightly"
- Slug `developer`, title e.g. "WNP Developer"
- Slug `beta`, title e.g. "WNP Beta"

### Test 1 — Channel-specific redirects (CMS pages present)

With all four pages published, confirm each URL returns a 302 to the expected destination with the version preserved in `?version=`:

- [x] http://localhost:8000/en-US/whatsnew/152.0a1/ → 302 to /en-US/whatsnew/nightly/?version=152.0a1
- [x] http://localhost:8000/en-US/whatsnew/152.0a2/ → 302 to /en-US/whatsnew/developer/?version=152.0a2
- [x] http://localhost:8000/en-US/whatsnew/152.0beta/ → 302 to /en-US/whatsnew/beta/?version=152.0beta
- [x] http://localhost:8000/en-US/whatsnew/152/ → 302 to /en-US/whatsnew/general/?version=152

Confirm each destination page loads as 200.

### Test 2 — Fallback to static pages (CMS pages absent)

Unpublish (or delete) the nightly, developer, and beta pages, leaving only the general page. Confirm the channel URLs now return 200 directly (static evergreen, no redirect):

- [x] http://localhost:8000/en-US/whatsnew/152.0a1/ → 200, static nightly evergreen
- [x] http://localhost:8000/en-US/whatsnew/152.0a2/ → 200, static developer evergreen
- [x] http://localhost:8000/en-US/whatsnew/152.0beta/ → 200, static release evergreen
- [x] http://localhost:8000/en-US/whatsnew/152/ → 302 to /en-US/whatsnew/general/ (general still live)

### Test 3 — Non-CMS locale always gets static page

With all WNP pages republished, confirm a locale that isn't CMS-backed (e.g. cy) never redirects:

- [x] http://localhost:8000/cy/whatsnew/152.0a1/ → 200, static nightly evergreen
- [x] http://localhost:8000/cy/whatsnew/152/ → 200, static release evergreen

### Test 4 — No redirect loop

Confirm the channel slug URLs are served directly by Wagtail and don't loop:

- [x] http://localhost:8000/en-US/whatsnew/nightly/ → 200, CMS page content
- [x] http://localhost:8000/en-US/whatsnew/developer/ → 200, CMS page content
- [x] http://localhost:8000/en-US/whatsnew/beta/ → 200, CMS page content
- [x] http://localhost:8000/en-US/whatsnew/general/ → 200, CMS page content

### DIFF
--- a/springfield/cms/tests/test_wnp_general_redirects.py
+++ b/springfield/cms/tests/test_wnp_general_redirects.py
@@ -7,10 +7,10 @@ Tests for the channel-aware WNP redirect behaviour.
 
 When a user requests /LOCALE/whatsnew/VERSION/ and there is no version-specific
 CMS WNP, we check for a channel-specific CMS WNP and 302 to it:
-  - Nightly (a1 suffix)  → /LOCALE/whatsnew/nightly/?version=VERSION
+  - Nightly (a1 suffix)   → /LOCALE/whatsnew/nightly/?version=VERSION
   - Developer (a2 suffix) → /LOCALE/whatsnew/developer/?version=VERSION
-  - Beta (b/beta suffix) → /LOCALE/whatsnew/beta/?version=VERSION
-  - Release              → /LOCALE/whatsnew/general/?version=VERSION
+  - Beta (beta suffix)    → /LOCALE/whatsnew/beta/?version=VERSION
+  - Release               → /LOCALE/whatsnew/general/?version=VERSION
 
 If no channel-specific CMS WNP exists for the locale (or its CMS fallback
 locale), the static evergreen page is rendered instead.
--- a/springfield/cms/tests/test_wnp_general_redirects.py
+++ b/springfield/cms/tests/test_wnp_general_redirects.py
@@ -3,11 +3,17 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 """
-Tests for the General WNP redirect behaviour (WT-1038).
+Tests for the channel-aware WNP redirect behaviour.
 
 When a user requests /LOCALE/whatsnew/VERSION/ and there is no version-specific
-CMS WNP but a General WNP (slug='general') exists for the locale (or its CMS
-fallback locale), we 302 to /LOCALE/whatsnew/general/?version=VERSION.
+CMS WNP, we check for a channel-specific CMS WNP and 302 to it:
+  - Nightly (a1 suffix)  → /LOCALE/whatsnew/nightly/?version=VERSION
+  - Developer (a2 suffix) → /LOCALE/whatsnew/developer/?version=VERSION
+  - Beta (b/beta suffix) → /LOCALE/whatsnew/beta/?version=VERSION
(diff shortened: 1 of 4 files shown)


springfield/cms/tests/test_wnp_general_redirects.py
```diff
@@ -180,3 +213,106 @@ def test_alias_locale_redirects_to_alias_url_when_fallback_has_general_wnp(
...
     assert "version=151" in location
+
+
+# ---------------------------------------------------------------------------
+# Channel-specific redirects (nightly, developer, beta)
+# ---------------------------------------------------------------------------
+
+
+def test_nightly_redirects_to_nightly_wnp(nightly_wnp, client):
+    """302 to nightly WNP when version ends with a1 and a nightly CMS page exists."""
```
maureenlholland, line 224: Is it worth testing the other en- locales redirect and use en-US CMS page? (This is happening as expected in browser testing, en-CA does not have a published CMS page but still redirects )

I'm assuming that would be redundant (because it's tested elsewhere) but wanted to check.

maureenlholland (approved):
r+ 🌃 

non-blocking test coverage question

PR #1338, opened 2026-05-08
WT874,875 - fix a11y violations (fix #1330)

wen-2018:

- refactors templates to include `pre-footer` in `main` to avoid elements outside landmarks
- avoid using aside inside landmark
- remove headings from navs

## Issue / Bugzilla link

https://github.com/mozmeao/springfield/issues/1330
https://mozilla-hub.atlassian.net/browse/WT-874
https://mozilla-hub.atlassian.net/browse/WT-875

## Testing

1. `cd tests/playwright`
2. `npm run a11y-tests` all tests should pass
3. Check pages using updated templates
4. Check `partial-target` AJAX swap still works: id="partial-target" and the conditional data-current="..." moved to `main_attrs` on the wrapper.


springfield/firefox/templates/firefox/all/includes/main-flare26.html
```diff
@@ -7,7 +7,6 @@
 {% set campaign = "firefox-all" %}
 {% set params = '?utm_source=www.firefox.com&utm_medium=referral&utm_campaign=' + campaign %}
 
-<main id="partial-target" class="fl-download-all fl-main" {% if product %} data-current="{{ product.slug }}"{% endif %}>
```
maureenlholland, line 10: **blocking:** `all` JS is trying to fully replace the `main` content section on every selection (browser, platform, language). Because we've moved `main` to the cms base, it only runs only the initial page load and when content is replaced, the `partial-target` attribute no longer exists.

Image

**suggested fix** The selector is only looking at id, not element type, so we can change the `main` element here to `div`, keep all the classes/attributes in place, and the JS should work as expected again.

maureenlholland (changes requested):
LGTM, thanks! 

One blocking issue on the `all` page, suggested fix in PR here: https://github.com/mozmeao/springfield/pull/1343

springfield/firefox/templates/firefox/all/base-flare26.html
```diff
@@ -40,6 +40,9 @@
   {% set current_step = ftl('firefox-all-choose-browser') %}
 {% endif %}
 
+{% block main_class %}fl-download-all fl-main{% endblock %}
```
maureenlholland: **avoid duplicated classes**: these are not used in JS, but the CSS is doubling padding-block-end

Image

maureenlholland (approved):
a11y tests pass ✅ 
download all partials work ✅ 

r+wc for the duplicated classes, might also be worth a push to integration tests before merge

maribedran:
The `fl-main` class has this big border radius; moving it up on the page hierarchy changes the design. The pre-footer should be outside `fl-main` to blend into the footer.

image
image

kkellydesign (approved):
Looks good to me 👍

PR #1340, opened 2026-05-08
[WT-1140] Kit Banner tweaks

maribedran:


kkellydesign (approved):
made some revisions to make sure these all work with kit banners, and made padding  consistent with mockups

PR #1341, opened 2026-05-08
Add body class field to Free Form 2026 page

maribedran:
## One-line summary

Allow pages to load a page-specific CSS bundle by setting a body class field.

- Add a new `body_class` field to the Free Form 2026 Page
- Load a CSS bundle with the same name as the body class when set

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1155

## Testing

- Create a Free Form 2026 Page and check that no extra class is added to the body element or any new CSS is loaded
- Set the "body class" field and check that the `<body` element gets the class
- Check the network tab for a 404 request trying to fetch the CSS file with that name
- Create a new entry in `media/static-bundles.json` with the name of the body class and check that the page now loads the CSS correctly


kkellydesign (approved):
This works well. *High fives*

PR #1345, opened 2026-05-11
Prevent redirects from caching query strings between requests (WT-1086)

janriokrause:
# Summary

* `redirects.util.redirect()` was mutating the closed-over `query` dict on every request, so query strings accumulated across requests when `merge_query=True`. Switched to a per-request `deepcopy` so each invocation starts clean.
* Added the `{_locale}` placeholder substitution branch for offsite redirects, so we use the already-known locale and avoid a follow-up 302 at the destination.
* Test coverage for: merge_query behaviour (on-site, with and without locale), offsite locale handling, and query string isolation across consecutive requests.

Ported from bedrock PR mozilla/bedrock#16391. Fixes WT-1086.

# Test plan

* [x] `pytest springfield/firefox/tests/test_redirects.py` passes
* [x] `make test` passes in CI

### DIFF
--- a/springfield/firefox/tests/test_redirects.py
+++ b/springfield/firefox/tests/test_redirects.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import importlib
+from functools import partial
 from unittest.mock import patch
 
 from django.http.response import HttpResponse
@@ -13,7 +14,7 @@ import pytest
 import springfield.firefox.redirects as redirects_module
 from springfield.firefox.redirects import mobile_app, validate_param_value
 from springfield.redirects.middleware import RedirectsMiddleware
-from springfield.redirects.util import get_resolver
+from springfield.redirects.util import get_resolver, redirect
 
 
 @pytest.mark.parametrize(
@@ -131,7 +132,7 @@ def test_refresh_redirect_destinations(source, destination, permanent):
         ("/browsers/desktop/mac/?utm_source=foo&utm_medium=bar", "/download/mac/?utm_source=foo&utm_medium=bar"),
     ),
 )
-def test_refresh_redirect_preserves_querystrings(source, destination, permanent):
+def test_refresh_redirect_preserves_query_strings(source, destination, permanent):
     rf = RequestFactory()
     with override_settings(PERMANENT_CMS_REFRESH_REDIRECTS=permanent):
         middleware = _get_refresh_middleware()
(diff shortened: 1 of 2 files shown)


springfield/firefox/tests/test_redirects.py
```diff
@@ -171,10 +172,114 @@ def test_refresh_redirects_not_in_redirectpatterns_when_disabled():
         ("/ai/", "/smart-window/?view=waitlist"),  # no locale; middleware will later add the most appropriate locale
         ("/en-US/ai/", "/en-US/smart-window/?view=waitlist"),  # with locale
         ("/fr/ai/", "/fr/smart-window/?view=waitlist"),  # non-default locale
-        ("/en-GB/ai/?foo=bar", "/en-GB/smart-window/?view=waitlist"),  # incoming querystrings lost until WT-1086 is done
```
janriokrause, line 174: Not sure if the plan was to set `merge_query=True` on the `/ai/` redirect? But in case this redirect will be removed, I wrote a dedicated test for the merge query functionality. So I think this approach is better.

stevejalim: We don't _have_ to set `merge_query=True` on `/ai/` (there was no request to do so) but it's good to know we now can if we need to. I don't mind if you do a small follow-up PR to enable `merge_query=True`

springfield/firefox/tests/test_redirects.py
```diff
@@ -171,10 +172,114 @@ def test_refresh_redirects_not_in_redirectpatterns_when_disabled():
...
+@pytest.mark.parametrize(
+    "path, expected_dest",
+    (
+        ("/en-US/firefox/new/?hello=world", f"{_TEST_EXT_HOST}/en-US/{EXPECTED_REDIRECT_QS}&hello=world"),
+        ("/en-US/firefox/new/", f"{_TEST_EXT_HOST}/en-US/{EXPECTED_REDIRECT_QS}"),
+        (
+            "/en-US/firefox/installer-help/?bar=baz&bam=bam",
+            f"{_TEST_EXT_HOST}/en-US/download/installer-help/{EXPECTED_REDIRECT_QS}&bar=baz&bam=bam",
+        ),
+        ("/en-US/firefox/installer-help/", f"{_TEST_EXT_HOST}/en-US/download/installer-help/{EXPECTED_REDIRECT_QS}"),
```
stevejalim, line 218: I can't spot what the `/en-US/firefox/installer-help/` redirect is testing that's not already covered by `/en-US/firefox/new/` a couple of lines above. What am I missing?

stevejalim: Oh it's that the query isn't being cached between runs, of course 😮‍💨 🤦

stevejalim (approved):
Nice one. I appreciate the improvements to the tests too 💯

PR #1346, opened 2026-05-12
WT-1000: Pattern lib refactor

lucianoratamero:
## One-line summary

This PR reorganizes the pattern library to use Flare26 as default and have leaner templates. It also adds visual regression 

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1000

## Testing

Manually open the pattern library and check the changes. Also, run visual regression tests to verify that they're passing.


maribedran:
@lucianoratamero I spotted a few issues going through the pattern library. Some of them are pre-existing, so you can decide if you want to fix them in this PR or in a follow up.

1. The icons on the icons list are missing the circle background.
image

2. Those two versions of the step cards list look the same. Were they supposed to be different?
image

3. We should probably update the icon variants to include all the icons available. http://localhost:8000/pattern-library/render-pattern/pattern-library/components/flare-26/icon/icon_variants.html
@dchukhin wrote a management command to update the icons' CSS. I suppose it could update the .yaml file too.

4. The tag variants are missing the corner variations: http://localhost:8000/pattern-library/pattern/pattern-library/components/flare-26/tag/tag_variants.html

5. The dark card on the two column cards has a border that shouldn't show when the image sticks to the border.
image

maribedran (approved):

PR #1352, opened 2026-05-14
Enable stub attribution mode on freeform pages

slightlyoffbeat:
Change FreeFormPage2026 to base its promote_panels on UTMParamsMixin.promote_panels instead of AbstractSpringfieldCMSPage.promote_panels. This ensures the UTM parameter promote panels are included for this page type while still appending the enable_marketing_attribution FieldPanel.


stevejalim (approved):
LGTM. r+

`UTMParamsMixin` is a model not a `Page` subclass, but its `promote_panels` already extends `AbstractSpringfieldCMSPage.promote_panels` so we don't "lose" any `Page` panels as a result of this

PR #1356, opened 2026-05-15
Make navigation always sticky, remove headroom.js

slightlyoffbeat:
The nav was using headroom.js to hide on scroll-down and reappear on scroll-up. We're switching to a permanently-pinned nav

The "scrolled past top" state (used for a subtle drop-shadow under the nav) is now detected with a 1px sentinel element + IntersectionObserver instead of headroom's scroll listener (same visual, no library).

Also:
- Add `scroll-padding-block-start` so in-page anchor links account for the sticky nav.
- Disable the shadow transition under `prefers-reduced-motion`.
- Remove now-dead `body:has(.fl-landing-get) .headroom-unpinned` override.
- Drop headroom.js dependency.

Saves about 5kb in bundle

### DIFF
--- a/media/css/cms/components/flare26-navigation.css
+++ b/media/css/cms/components/flare26-navigation.css
@@ -16,6 +16,10 @@
 * and preferred-color-scheme fallback, in case a class isn't provided.
 */
 
+html {
+    scroll-padding-block-start: var(--fl-navigation-menu-height);
+}
+
 .fl-header {
     background-color: var(--fl-theme-background);
     color: var(--fl-theme-color-text);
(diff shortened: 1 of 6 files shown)


janriokrause (approved):
r+ Looks good to me!

Nice touch adding `scroll-padding-*` to avoid anchor jump under the sticky header, and honoring `prefers-reduced-motion` ✨

If we want to support very old browsers, we could consider a lightweight fallback like always add `is-scrolled` or use a simple `scroll` listener when `IntersectionObserver` is unavailable. But I think it's fine.

kkellydesign (approved):
Works for me 👍

PR #1366, opened 2026-05-18
Attribute mobile downloads from /thanks/-bound buttons and CMS buttons

slightlyoffbeat:
## Summary
Closes the mobile attribution gap on landing-page download buttons. Two related fixes:
### Path B — `/thanks/`-bound CTAs (nav, pre-footer, etc.)
Today, mobile users clicking these buttons land on `/thanks/`, which auto-redirects to a **bare** Play Store / App Store URL with no `install_referrer` / `pt`+`ct` params for the originating page's campaign. The install reaches the store but the page-campaign is lost.

A new client-side module (`media/js/base/mobile-attribution.js`):
- Resolves the page's campaign via the same cascade `stub-attribution.js` uses: `force > override > URL utm_campaign > default`.
- On mobile UAs, rewrites matching button HREFs to attributed `play_store_url()` / `app_store_url()`-shaped URLs.
- Attaches a click handler that fires `Mozilla.TrackProductDownload.sendEventFromURL` with the captured `https://` URL (mirroring `thanks-init.js`) and navigates via `window.location.href` so the rewrite works in any browser despite `mozilla-utils.js`'s `market://` scheme mutation.

### Path A — CMS `download_firefox_button` Play/App Store badges

These already have attributed URLs server-side and have always fired the `firefox_mobile_download` GA4 event correctly **on iOS** — the existing [`datalayer-productdownload-init.es6.js`](media/js/base/datalayer-productdownload-init.es6.js) attaches `handleLink` to every `.ga-product-download` element.

But on Android the event has been silently dropping. `mozilla-utils.js` mutates `https://play.google.com/...` to `market://details?...` on DOM ready, and the tracker's `marketURL` regex (`/^market:\/\/play.google.com/`) doesn't match that shape — so `isValidDownloadURL` returns false and the event never fires.

We supplement Android Path A buttons (`.fl-store-button-android`) with a click handler that uses the captured `https://` URL. iOS Path A is untouched (existing handler still works there) — adding ours would double-fire the event.

Scenario Before this PR After this PR
---------

Mobile click on **nav / pre-footer** download CTA (Android or iOS), page has a campaign Click → `/thanks/` → auto-redirect to **bare** store URL (with hardcoded `utm_campaign=download` for Android, or no `pt`/`ct` for iOS). `firefox_mobile_download` event fires correctly from `/thanks/` (it captures the URL synchronously), but the install itself reaches the store with no page-campaign attribution. Click → directly to **attributed** store URL with the page's campaign. `firefox_mobile_download` event fires from the landing page (not `/thanks/`) with the same payload. Same total event volume; campaign now reaches Play Console / App Store Connect.

Mobile click on **nav / pre-footer** CTA, no campaign anywhere on page Same as today (bare store URL via `/thanks/` auto-redirect, event fires from `/thanks/`). Strict no-op (JS exits early). Identical to today.

Click on **CMS `download_firefox_button` Play Store badge (Android)** Event silently dropped on click. The existing `datalayer-productdownload-init.es6.js` handler reads `event.target.href` at click time, which `mozilla-utils.js` has already mutated to `market://details?...`; the tracker's `marketURL` regex (`/^market:\/\/play.google.com/`) doesn't match that shape. No `firefox_mobile_download` event fires. Event now fires correctly with the campaign in the URL. **Expect previously-zero volume on Android Path A clicks to appear in dashboards.**

Click on **CMS `download_firefox_button` App Store badge (iOS)** Event fires correctly (existing handler works — `mozilla-utils.js` doesn't mutate iOS URLs). Unchanged — we deliberately don't double-attach. `/thanks/` page itself (mobile user lands there directly, e.g. from a hand-rolled link or in the absence of a campaign) `thanks-init.js` auto-redirects after 1s and fires the event correctly (captures URL synchronously before mutation). **Unchanged.** Our JS doesn't touch the auto-redirect path. `/thanks/` GTM pageview for mobile users transiting `/thanks/` Fires Skipped for users who would have transited `/thanks/` (because they now go directly to the store). Mobile users who land on `/thanks/` directly still fire it.
**No GTM dashboard changes needed.** The canonical signal for mobile downloads is the `firefox_mobile_download` event itself, which fires under both Path A and Path B. If a dashboard estimates mobile downloads by counting `/thanks/` pageviews specifically, it should switch to the event count.

**Existing latent bug worth a follow-up (separate PR)**: the `marketURL` regex in [`datalayer-productdownload.es6.js:15`](media/js/base/datalayer-productdownload.es6.js) is `/^market:\/\/play.google.com/`, but `mozilla-utils.js`'s `initMobileDownloadLinks` produces `market://details?...` (no `play.google.com` in the path). This affects any click-time-URL-read handler on a `.ga-product-download` Android button. `/thanks/`'s auto-redirect is unaffected because `thanks-init.js` captures the URL synchronously before the mutation. This PR works around the regex bug for our specific buttons via the same synchronous-capture pattern. A separate PR could broaden the regex to `/^market:\/\//` to fix the root cause for any other Path A buttons across the site (e.g. release-notes, landing/gaming, landing/ios-summarizer).

## Files
- `media/js/base/mobile-attribution.js` (new)
- `tests/unit/spec/base/mobile-attribution/mobile-attribution.js` (new)
- `media/static-bundles.json` — new `mobile-attribution` bundle
- `webpack.test.config.js` — `datalayer-productdownload-init.es6.js` and `mobile-attribution.js` added to `test-deps`
- `springfield/base/templates/base-protocol.html`, `springfield/cms/templates/cms/base-flare26.html`, `springfield/cms/templates/cms/base-flare.html` — load the new bundle alongside the existing `stub-attribution` bundle
### Regression checks (must be byte-identical to before)
- [x] **Desktop (no UA spoof), `stub_attr_utm_campaign_value="test-19"` set**: nav button HREF stays `/thanks/`. Click → stub installer flow, `moz-stub-attribution-code` cookie carries `test-19`.
- [x] **Page with no `stub_attr` and no URL `utm_campaign`, mobile UA**: nav HREF stays `/thanks/`. JS exits early. Identical to today.
- [x] **Hand-rolled footer "Download Firefox" link** still routes to `/thanks/` on any UA (intentional — no wrapper class).
- [x] **Mobile user lands directly on `/thanks/`** (e.g. from a hand-rolled link or no-campaign nav): auto-redirect to store fires after 1s with the existing event. Unchanged.
### New behavior
- [x] **Mobile UA Chrome devtools (Android UA), page with `test-19`**: nav and pre-footer button HREFs rewrite to `play.google.com/...&utm_campaign=test-19`. `mozilla-utils.js` further converts to `market://...` on real Android. Click → store install attributed.
- [x] **Mobile UA Chrome devtools (iOS UA), same page**: nav and pre-footer button HREFs rewrite to `apps.apple.com/...&ct=test-19`. Click → App
Store.
- [x] **URL-param attribution**: any page (CMS or not) with `?utm_campaign=test-19`, mobile UA → nav HREF rewrites.
- [x] **Cascade precedence**: `stub_attr_utm_campaign_mode="force"` value `force-val` + URL `?utm_campaign=url-val` → mobile HREF reflects `force-val`. Switch mode to `default` → URL value wins.
- [x] **Path A Android button**: CMS `download_firefox_button` block on a page, Android UA, click the green Play Store badge. In DevTools
Console: `dataLayer.filter(e => e.event === 'firefox_mobile_download')` returns an entry with `platform: 'android'`, `method: 'store'`, `release_channel: 'release'`. (Previously this would have been empty due to the regex mismatch.)
- [x] **Path A iOS button**: same setup, iOS UA, click the black App Store badge. Confirm exactly **one** event fires (not two — we're relying on the existing handler, not adding our own).
### End-to-end (recommended before launch)
- [ ] **Real Android device**: visit a `stub_attr`-configured page on a device without Firefox installed. Tap nav button → Play Store opens → install Firefox → open it once. Verify the install appears in Play Console attributed to the campaign value.
- [ ] **Real iOS device**: same, with the App Store. Verify campaign shows in App Store Connect attribution.

### DIFF
--- a/media/js/base/mobile-attribution.js
+++ b/media/js/base/mobile-attribution.js
@@ -1,0 +1,234 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Mobile attribution: rewrites /thanks/-bound download buttons to attributed
+ * store URLs on mobile UAs so the page-level campaign survives the click.
+ * Desktop is unaffected — those clicks continue through the stub-attribution
+ * cookie/installer flow.
+ *
+ * Campaign resolution mirrors stub-attribution.js exactly:
+ *   force > override > URL utm_campaign > default.
+ */
+
+if (typeof window.Mozilla === 'undefined') {
+    window.Mozilla = {};
+}
+
+(function (Mozilla) {
+    'use strict';
+
+    var MobileAttribution = {};
+
+    var ANDROID_RE = /\bAndroid\b/i;
+    var IOS_RE = /\b(iPhone|iPad|iPod)\b/i;
+
+    /**
+     * Resolve the campaign value using the same cascade as stub-attribution.js:
+     * force > override > URL utm_campaign > default.
+     * @param {HTMLElement} html - typically document.documentElement.
+     * @param {String} search - typically window.location.search.
+     * @returns {String|null}
+     */
+    MobileAttribution.getCampaign = function (html, search) {
+        // IE9-safe URL parse (avoids URLSearchParams).
+        var urlMatch = (search || '').match(/[?&]utm_campaign=([^&]+)/);
+        var urlCampaign = urlMatch ? decodeURIComponent(urlMatch[1]) : null;
+
+        return (
+            html.getAttribute('data-stub-attribution-campaign-force') ||
+            html.getAttribute('data-stub-attribution-campaign-override') ||
+            urlCampaign ||
+            html.getAttribute('data-stub-attribution-campaign') ||
+            null
+        );
+    };
+
+    /**
+     * Build the attributed store URL for the given campaign + platform.
+     * Mirrors springfield/firefox/templatetags/misc.py app_store_url /
+     * play_store_url so the rewritten URL is byte-identical to what the CMS
+     * download_firefox_button block emits server-side.
+     */
+    MobileAttribution.getStoreUrl = function (campaign, isAndroid) {
+        var encoded = encodeURIComponent(campaign);
+        if (isAndroid) {
+            return (
+                'https://play.google.com/store/apps/details?id=org.mozilla.firefox' +
+                '&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral' +
+                '%26utm_campaign%3D' +
+                encoded
+            );
+        }
+        return (
+            'https://apps.apple.com/app/apple-store/id989804926' +
+            '?mz_pr=firefox_mobile&pt=373246&ct=' +
+            encoded +
+            '&mt=8'
+        );
+    };
+
+    /**
+     * Find /thanks/-bound download buttons, rewrite their HREFs to the given
+     * store URL, and attach a click handler that fires the
+     * `firefox_mobile_download` + legacy `product_download` GA4 events that
+     * /thanks/ would have fired today (via thanks-init.js calling
+     * sendEventFromURL after the auto-redirect).
+     *
+     * Selector covers both shapes:
+     *   - `.c-button-download-thanks-link` (set on buttons when
+     *     exclude_unsupported_content is False)
+     *   - `.c-button-download-thanks > .download-link` (the broader pattern,
+     *     covers nav + pre-footer where exclude_unsupported_content is True
+     *     and the `-link` class is omitted)
+     * Path A store buttons (rendered by show_store_button) use
+     * `.fl-store-button-*` classes and are not matched.
+     *
+     * Why sendEventFromURL with a closure-captured URL (vs. attaching
+     * handleLink directly): mozilla-utils.js's initMobileDownloadLinks
+     * mutates Android Play Store hrefs to `market://details?...` on DOM
+     * ready. TrackProductDownload's marketURL regex requires the URL to
+     * start with `market://play.google.com` (which mozilla-utils.js never
+     * produces), so reading event.target.href at click time would not
+     * match. Capturing the original https:// URL bypasses the regex
+     * mismatch entirely.
+     */
+    MobileAttribution.rewriteLinks = function (root, storeUrl) {
+        var buttons = root.querySelectorAll(
+            '.c-button-download-thanks-link, .c-button-download-thanks > .download-link'
+        );
+        for (var i = 0; i < buttons.length; i++) {
+            var href = buttons[i].getAttribute('href') || '';
+            if (href.indexOf('/thanks') === -1) continue;
+
+            buttons[i].setAttribute('href', storeUrl);
+
+            if (
+                window.Mozilla &&
+                window.Mozilla.TrackProductDownload &&
+                typeof window.Mozilla.TrackProductDownload.sendEventFromURL ===
+                    'function'
+            ) {
+                MobileAttribution._attachTrackingClick(buttons[i], storeUrl);
+            }
+        }
+    };
+
+    /**
+     * Attach a click listener that fires the GA4 + legacy download events
+     * AND navigates to the closure-captured https:// store URL.
+     *
+     * Why navigate via JS instead of letting the browser follow the HREF:
+     * mozilla-utils.js mutates Android Play Store hrefs to `market://...`
+     * on DOM ready. `market://` resolves correctly on real Android (the
+     * Play Store app is registered for the scheme) but fails in desktop
+     * browsers (including Chrome with Android UA emulation), surfacing a
+     * "scheme has no registered handler" error and breaking dev testing.
+     *
+     * Navigating to the https:// URL works in both contexts: on real
+     * Android, Android App Links open the Play Store app from the https
+     * URL automatically; in any browser, it loads play.google.com /
+     * apps.apple.com directly. This mirrors how thanks-init.js navigates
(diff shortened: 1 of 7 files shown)


media/js/base/mobile-attribution.js
```diff
@@ -0,0 +1,234 @@
...
+        );
+    };
+
+    /**
+     * Build the attributed store URL for the given campaign + platform.
+     * Mirrors springfield/firefox/templatetags/misc.py app_store_url /
+     * play_store_url so the rewritten URL is byte-identical to what the CMS
+     * download_firefox_button block emits server-side.
+     */
+    MobileAttribution.getStoreUrl = function (campaign, isAndroid) {
```
stephendherrera, line 52: This function is in line with what we've been advised each store needs for downloads.

stephendherrera (approved):

media/js/base/mobile-attribution.js
```diff
@@ -0,0 +1,234 @@
...
+ */
+
+/*
+ * Mobile attribution: rewrites /thanks/-bound download buttons to attributed
+ * store URLs on mobile UAs so the page-level campaign survives the click.
+ * Desktop is unaffected — those clicks continue through the stub-attribution
+ * cookie/installer flow.
+ *
+ * Campaign resolution mirrors stub-attribution.js exactly:
+ *   force > override > URL utm_campaign > default.
```
maureenlholland: @slightlyoffbeat I'm removing the override logic in refactored download-attribution. We don't have any current use of it and it is not clear why we need this attribute (why would a campaign be important enough to override a UTM campaign, but not important enough to refresh the attribution cookie?)

Data attribute campaigns are easier to reason about without the `override` option:
`data-stub-attribution-campaign-force` => essential campaign that should refresh attribution cookie
`data-stub-attribution-campaign` => default analytics campaign that will apply if no UTM campaign is available

**non-blocking**: remove `data-stub-attribution-campaign-override` logic from new code

It doesn't have to be removed in this PR. I can remove in the download attribution PR after this merges. But calling out for awareness.

media/js/base/mobile-attribution.js
```diff
@@ -0,0 +1,234 @@
...
+     * mozilla-utils.js mutates Android Play Store hrefs to `market://...`
+     * on DOM ready. `market://` resolves correctly on real Android (the
+     * Play Store app is registered for the scheme) but fails in desktop
+     * browsers (including Chrome with Android UA emulation), surfacing a
+     * "scheme has no registered handler" error and breaking dev testing.
+     *
+     * Navigating to the https:// URL works in both contexts: on real
+     * Android, Android App Links open the Play Store app from the https
+     * URL automatically; in any browser, it loads play.google.com /
+     * apps.apple.com directly. This mirrors how thanks-init.js navigates
```
maureenlholland: FYI: part of the download attribution refactor means we are renaming the `thanks` JS files to `auto-download` (this allows us to reuse the logic for RTAMO and have a more descriptive name). 

**non-blocking, documentation**: these comments are very verbose. If we can simplify for clarity, that would be easier to scan.

stevejalim:
@maureenlholland Would you mind taking a look at the failing tests here, please?

slightlyoffbeat:
@maureenlholland @stevejalim tightened up comments and fixed tets

media/js/base/mobile-attribution.js
```diff
@@ -0,0 +1,182 @@
...
+
+            buttons[i].setAttribute('href', storeUrl);
+
+            if (
+                window.Mozilla &&
+                window.Mozilla.TrackProductDownload &&
+                typeof window.Mozilla.TrackProductDownload.sendEventFromURL ===
+                    'function'
+            ) {
+                MobileAttribution._attachTrackingClick(buttons[i], storeUrl);
```
maureenlholland, line 93: explaining out loud to myself - this could be called twice for android, but for different use cases

In init, Android has a special case to **always** run `MobileAttribution.attachAndroidStoreButtonTracking(document)`, which calls `_attachTrackingClick`. This only applies to elements with a `fl-store-button-android` class. 

**If there is a campaign**, both iOS and Android run `MobileAttribution.rewriteLinks(document, storeUrl);`, which calls `_attachTrackingClick` here. This only applies to elements with `c-button-download-thanks-link` or `c-button-download-thanks > .download-link` selectors.

media/js/base/mobile-attribution.js
```diff
@@ -0,0 +1,182 @@
...
+     *      the captured https:// URL. iOS Path A is unaffected and not
+     *      supplemented (would double-fire).
+     *   2. Path B: rewrite /thanks/-bound CTAs to attributed store URLs.
+     *      Gated on a declared campaign (CMS stub_attr or URL utm_campaign).
+     */
+    MobileAttribution.init = function () {
+        var html = document.documentElement;
+        var ua = navigator.userAgent || '';
+        var isAndroid = ANDROID_RE.test(ua);
+        var isIOS = IOS_RE.test(ua);
```
maureenlholland, line 137: **(non-blocking) nit, maintenance** this is slightly different from the way we detect iOS in `site.js`, it might be worth leveraging that existing logic: https://github.com/mozmeao/springfield/blob/f727777ac6b5fc5fd10db545a90a34e6a9089a30/media/js/base/site.js#L27

Non-blocking because I don't see any difference between prod & this PR for platform detection

maureenlholland (approved):
r+ waiting on successful integration test run

maureenlholland:
successful test run: https://github.com/mozmeao/springfield/actions/runs/26957182916

PR #1367, opened 2026-05-19
Create Roadmap page

maribedran:
## One-line summary

Create the Roadmap page and its blocks.

- Create the new `RoadmapPage` page type
- Create the `RoadmapListSectionBlock`, a list of roadmap items with statuses and tags
- Implement filtering by tag for roadmap items

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1168
https://www.figma.com/design/5ebjB97defiRGZ7C5N0Hoi/Firefox.com-Design-System?node-id=116457-192450&m=dev

## Testing

- Check the pattern library: http://localhost:8000/pattern-library/pattern/pattern-library/components/flare-26/roadmap-list-section/roadmap-list-section.html
- Check the test page: http://localhost:8000/en-US/tests-index-page-2026/test-roadmap-list/
  - Visit the page with a filter query and check that the filter is applied: http://localhost:8000/en-US/tests-index-page-2026/test-roadmap-list/?tags=ios


springfield/cms/templates/cms/blocks/sections/section.html
```diff
@@ -31,7 +31,7 @@
     {% if value.cta %}
       <content:cta>
         {% for cta in value.cta %}
-          {% set block_position = block_position ~ ".cta" %}
+          {% set block_position = block_position ~ ".cta-" ~ loop.index %}
           {% include_block cta %}
```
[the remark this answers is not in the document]
maribedran: That's not how template variables actually work

media/css/cms/components/flare26-roadmap-list-section.css
```diff
@@ -0,0 +1,220 @@
...
+
+@media (prefers-reduced-motion: reduce) {
+    .fl-roadmap-item {
+        transition: none;
+    }
+}
+
+.fl-roadmap-item-icon .fl-icon {
+    block-size: 20px;
+    inline-size: 20px;
```
lucianoratamero: just so you know, the fl-icon class already has a block-size and inline-size set to 1em, so I'm changing this to a font-size instead of block/inline-size

media/css/cms/components/flare26-roadmap-list-section.css
```diff
@@ -0,0 +1,220 @@
...
+}
+
+.fl-roadmap-filter-button.is-active .fl-tag::after {
+    background-color: currentColor;
+    block-size: var(--token-spacing-md);
+    content: '';
+    inline-size: var(--token-spacing-md);
+    inset-block-start: 50%;
+    inset-inline-end: var(--token-spacing-md);
+    -webkit-mask: url('/media/img/firefox/flare/2026/icons/desktop-16/close-cancel-dismiss/close-16.svg')
```
lucianoratamero: I'd love to not have this inline here.
we could use the after_icon param from the tag component to load it as a <include:icon> instead, if the tag component was setting the icon as a slot, instead of a prop. 

this is a great opportunity to refactor the tag component to allow that, since we still don't use it too much, if at all.
that would also allow you to add a wrapper to the icon, so you can hide/show it with js.

thoughts?

tests/playwright/specs/visual-regression/roadmap-section.spec.js-snapshots/roadmap-list-section-chromium-linux.png
lucianoratamero, line 1: could we have only one roadmap section on the visual regression test?
each file is almost 200kb just for this test, and I think we should be more selective with what/how we test it.

also, maybe having the full thing here just to test the active state is a bit overkill.
imo, I'd prefer to either have a test for active-inactive separately with just the tags, or not have the active test at all.

lucianoratamero: it also brings the question: maybe the active/inactive states should be something the tags css itself implements, instead of keeping it here.

lucianoratamero (approved):

l10n/en/cms/roadmap.ftl
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# The Roadmap page shows a list of features coming up on Firefox
```
flodolo, line 5: This is stand-alone comment that nobody will see. It would have been useful as a resource comment (three dashes).

l10n/en/cms/roadmap.ftl
```diff
@@ -0,0 +1,22 @@
...
+
+# Label indicating when the list was last updated
+# Variables:
+#    $date (string) Date when the page was last updated
+roadmap-last-updated = Last updated: { $date }
+
+# Pill indicating the status of an item in Firefox's roadmap
```
flodolo, line 12: This should have been a group comment.

l10n/en/cms/roadmap.ftl
```diff
@@ -0,0 +1,22 @@
...
+roadmap-status-exploring = Exploring
+roadmap-status-in-progress = In Progress
+roadmap-status-testing = Testing
+roadmap-status-coming-soon = Coming Soon
+roadmap-status-recently-shipped = Recently Shipped
+
+# Tag indicating where the feature is available
```
flodolo, line 19: Same here. This comment only applies to the next string.

PR #1369, opened 2026-05-20
WT-1217: Drop HTML comments from sanitized output

stevejalim:
Since switching from bleach to justhtml (WT-578), HTML comments in release-notes Markdown have been surfaced on the page as literal `<!-- ... -->` text because the markup was HTML escaped. With the setting on `disallowed_tag_handling="escape"`, justhtml's escape pass wins over its DropComments transform, and using an explicit Sanitize() to force ordering discards our custom policy.

The fix is to strip `<!--...-->` with a regex before handing the HTML to justhtml. This keeps the escape behaviour for unknown tags so example markup still renders as visible source.

## Testing

* `./manage.py update_release_notes`
* on main, visit http://localhost:8000/en-GB/firefox/153.0a1/releasenotes/ scroll down to Web Platform and see a HTML comment (escaped)
* switch to this branch 
* reload the page and you'll see the HTML comment is gone

### DIFF
--- a/springfield/base/sanitization.py
+++ b/springfield/base/sanitization.py
@@ -9,7 +9,14 @@ This module provides functions for sanitizing HTML content, replacing
 the deprecated bleach library.
 """
 
+import re
+
 from justhtml import JustHTML, SanitizationPolicy, UrlPolicy, UrlRule
+
+# Matches HTML comments. justhtml's DropComments transform cannot drop
+# comments when disallowed_tag_handling="escape" is in effect), so
+# we strip comments before handing the input to justhtml.
+_HTML_COMMENT_RE = re.compile(r"<!--.*?-->", re.DOTALL)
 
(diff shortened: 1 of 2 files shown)


springfield/base/sanitization.py
```diff
@@ -9,8 +9,15 @@
 the deprecated bleach library.
 """
 
+import re
+
 from justhtml import JustHTML, SanitizationPolicy, UrlPolicy, UrlRule
 
+# Matches HTML comments. justhtml's DropComments transform cannot drop
+# comments when disallowed_tag_handling="escape" is in effect), so
```
maureenlholland:
```suggestion
# comments when disallowed_tag_handling="escape" is in effect, so
```

maureenlholland (approved):
comments gone 🧹 typo nitpick in comment r+

PR #1370, opened 2026-05-20
WT-1174: Landing page follow up tasks

lucianoratamero:
## One-line summary

This PR adds tweaks necessary for the landing pages. See ticket for more info.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1174


media/css/cms/components/flare26-banner.css
```diff
@@ -103,7 +103,14 @@
 }
 
 .fl-banner.fl-banner-dark-purple-gradient.fl-banner-kit {
-    background-image: var(--kit-img), var(--token-noise-gradient);
+    background-color: var(--token-color-dark-purple);
+    background-image: var(--token-noise-gradient);
+}
+
+/* When a kit banner is the last item on the page, affix it to the bottom */
+.fl-split-page-lower > .fl-banner-container:last-child {
```
maribedran: The banner block has the conditional display option. If there's a condition, the last child selector won't work.

maribedran (approved):

media/css/cms/components/flare26-banner.css
knowler: @maribedran Is this still true if we were to leverage the [`:nth-last-child(<nth> of <complex-selector>)`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:nth-last-child) syntax [which has fairly wide support](https://caniuse.com/css-nth-child-of)? I’m still getting familiar with the various blocks and the markup they produce, so I can’t tell for sure.

I ask because there’s a JS error with the component since the `document.querySelectorAll()` selector uses `:has()` and some older browsers, namely ESR 115, don’t support `:has()` (see: https://mozilla-hub.atlassian.net/browse/WT-1266). That’s easy enough to avoid by feature checking with `CSS.supports("selector(:has(+ *))")`, but when I looked into what the component was doing, I wondered if we could actually accomplish this without any JS.

PR #1373, opened 2026-05-21
Technical SEO - Stop emitting empty meta description tags

slightlyoffbeat:
## Summary

When a CMS page has a blank `search_description`, base-flare26.html was rendering `<meta name="description" content="">` and `<meta property="og:description" content="">` — empty tags. This change captures the rendered block content, strips tags and whitespace, and only emits the surrounding `<meta>` tag when there's actual content to put in it.

## Why

- An **empty** meta description tag explicitly signals "no description" to crawlers and LLM-driven search.
- An **absent** tag lets Google synthesize a snippet from the page content (which it does well).
- Absent is strictly better than empty.

Part of the firefox.com technical SEO workstream (item A3 in the master plan).

  ## What changed

`springfield/cms/templates/cms/base-flare26.html` only — two parallel changes on the `page_desc` and `page_og_desc` blocks. The block definitions are unchanged, so the ~60 child templates that override them (`features/fast.html`, `browsers/desktop/base.html`, etc.) continue to work without modification.

Before:
```jinja
  <meta name="description" content="{% filter striptags %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is
  defined }}{% endblock %}{% endfilter %}">
```

After:
```jinja
{% set page_desc_content %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endset
%}
{% set page_desc_content = page_desc_content|striptags|trim %}
{%- if page_desc_content %}
<meta name="description" content="{{ page_desc_content }}">
{%- endif %}
```

Same pattern for `page_og_desc`.

## Out of scope

The same defect exists in `cms/base-flare.html` and `base/base-protocol.html`. Both are being sunset and intentionally excluded from this change.

  ## Test plan

  - [x] CMS page with blank `search_description` → no `<meta name="description">` and no `<meta property="og:description">` in rendered `<head>`
  - [x] CMS page with `search_description` set → both tags render with content
  - [ ] Hard-coded template overriding `page_desc` (e.g. `firefox/features/fast/`) → meta description tag renders with the overridden content
  - [ ] `firefox/landing/win10-eos/` (explicit empty `{% block page_desc %}{% endblock %}`) → no tag rendered (improvement over current empty tag)
  - [x] `pytest springfield/cms/tests/` passes

### DIFF
--- a/springfield/cms/templates/cms/base-flare26.html
+++ b/springfield/cms/templates/cms/base-flare26.html
@@ -44,14 +44,22 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
     {% block shared_meta %}
     <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{{ (page.og_title or page.title) if page is defined and page.title is defined else 'Firefox' }}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox.com{% endblock %}{% endfilter %}</title>
-    <meta name="description" content="{% filter striptags %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endfilter %}">
+    {% set page_desc_content %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endset %}
+    {% set page_desc_content = page_desc_content|striptags|trim %}
+    {%- if page_desc_content %}
+    <meta name="description" content="{{ page_desc_content }}">
+    {%- endif %}
     <meta property="og:type" content="website">
     <meta property="og:site_name" content="Firefox">
     <meta property="og:locale" content="{{ LANG|replace("-", "_") }}">
     <meta property="og:url" content="{% filter trim|absolute_url %}{% block page_og_url %}{{ settings.CANONICAL_URL + '/' + LANG + canonical_path }}{% endblock %}{% endfilter %}">
     <meta property="og:image" content="{% filter trim|absolute_url %}{% block page_image %}{% if page and page.og_image %}{% set img = image(page.og_image, "width-1200") %}{{ img.url }}{% else %}{{ static('img/firefox/flare/og.png') }}{% endif %}{% endblock %}{% endfilter %}">
     <meta property="og:title" content="{% filter striptags %}{% block page_og_title %}{{ (page.og_title or page.title) if page is defined and page.title is defined else 'Firefox'  }}{% endblock %}{% endfilter %}">
-    <meta property="og:description" content="{% filter striptags %}{% block page_og_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endfilter %}">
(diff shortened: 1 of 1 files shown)


springfield/cms/templates/cms/base-flare26.html
```diff
@@ -44,14 +44,22 @@
 
     {% block shared_meta %}
     <title>{% filter striptags %}{% block page_title_full %}{% block page_title_prefix %}{% endblock %}{% block page_title %}{{ (page.og_title or page.title) if page is defined and page.title is defined else 'Firefox' }}{% endblock %}{% endblock page_title_full %}{% block page_title_suffix %} — Firefox.com{% endblock %}{% endfilter %}</title>
-    <meta name="description" content="{% filter striptags %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endfilter %}">
+    {% set page_desc_content %}{% block page_desc %}{{ page.og_description if page is defined and page.og_description is defined }}{% endblock %}{% endset %}
+    {% set page_desc_content = page_desc_content|striptags|trim %}
```
maureenlholland, line 48: the hardcoded templates use `page_desc`, so this is not using the hardcoded desc as override (there is no desc on features/fast page on PR & it is an empty desc on prod)
https://github.com/mozmeao/springfield/blob/main/springfield/firefox/templates/firefox/features/fast-2024.html#L12

maureenlholland (changes requested):
incomplete review, but the hardcoded override checklist item is not behaving as described

maureenlholland:
NOTE: "search_description" is code-facing, the term you see in CMS dashboard editor is "Meta Description"
Screenshot 2026-06-08 at 5 49 50 PM

maureenlholland:
http://localhost:8000/en-US/features/fast/ not behaving as described
no-tag


Expected
- hardcoded description applied

maureenlholland:
http://localhost:8000/en-US/landing/win-new-features/ not behaving as described
Screenshot 2026-06-08 at 5 53 07 PM

Expected
- no tag

maureenlholland:
will investigate failed criteria tomorrow

maureenlholland:
issue with `firefox/features/fast/` is that it's rendered by CMS
http://localhost:8000/en-US/features/fast/

there is a static, Protocol-based template with `page_desc` but this does not extend the base-flare26 CMS template
http://localhost:8000/sco/features/fast/
Screenshot 2026-06-11 at 3 29 17 PM

No override is possible for this URL. The CMS needs to be populated with existing meta description translations.

If I switch over the Protocol-based template to Flare26 base, the meta desc works as described
Screenshot 2026-06-11 at 3 35 13 PM

maureenlholland:
http://localhost:8000/en-US/landing/win-new-features/ issue:

also extends static Protocol template, not included in these changes

it's noindexed, so the meta desc shouldn't matter anyway

maureenlholland (approved):
From what I can tell, this is working as expected, the test cases are a little misleading

http://localhost:8000/en-US/features/fast/ => rendered by CMS, no CMS meta description means no rendered meta desc tag

Any static page with `base-flare26` will be able to use the `page_desc` var

An empty CMS description will not be able to pull from a static `page_desc`, they are separate templates that do not inherit from each other

PR #1374, opened 2026-05-21
Technical SEO - Add Vary: User-Agent to /download/ platform redirect

slightlyoffbeat:
## Summary

The `/download/` endpoint redirects to `/download/{platform}/` based on the requesting client's `User-Agent`. This change adds `Vary: User-Agent` to that response so caches and crawlers correctly understand the endpoint as content-negotiated.
## Why

- `download_redirect()` reads `User-Agent` and picks a destination — without `Vary: User-Agent`, an intermediate cache could serve one user's resolved destination to another user with a different platform
- `@never_cache` already protects against compliant caches, but `Vary` is the correct semantic declaration and protects against misconfigured intermediaries that ignore `Cache-Control`
- Most important for SEO: Googlebot and AI search crawlers can correctly understand `/download/` as a content-negotiated endpoint rather than treating one observed destination as canonical

Part of the firefox.com technical SEO workstream (item A13 in the master plan; A13a was already correctly setting `Vary: Accept-Language` on the locale redirect).
## What changed
`springfield/firefox/views.py` — `download_redirect()` view. The platform-detection return path now builds the response, sets `Vary: User-Agent`, and returns.

The switch-off fallback path (line 490) is intentionally not modified — it returns a fixed homepage redirect regardless of User-Agent, so declaring variance there would be inaccurate.
## Test plan

- [ ] `curl -I https://www.firefox.com/en-US/download/` (after deploy) shows `Vary: User-Agent` in response headers
- [ ] Hit `/download/` locally with different `User-Agent` strings via curl — confirm correct platform redirect for each, and `Vary: User-Agent` present
- [ ] `pytest springfield/firefox/tests/` passes

### DIFF
--- a/springfield/firefox/views.py
+++ b/springfield/firefox/views.py
@@ -505,7 +505,9 @@ def download_redirect(request):
     if qs:
         redirect_url = f"{redirect_url}?{qs}"
 
-    return HttpResponseRedirect(redirect_url)
+    response = HttpResponseRedirect(redirect_url)
+    response["Vary"] = "User-Agent"
+    return response
 
 
 class DownloadView(L10nTemplateView):


janriokrause:
_🚗 passing through without whole review_

I want to mention, that [Mozmeao Platforms Documentation on Redirects](https://mozmeao.github.io/platform-docs/development/redirects/#utilities) says "Do not set the `Vary: User-Agent` header; this will not work in production.".

Also, but not sure if relevant, we should use `patch_vary_headers()` maybe, because it preserves any existing `Vary` values rather than replacing the whole header.

``` python
from django.utils.cache import patch_vary_headers

response = HttpResponseRedirect(redirect_url)
patch_vary_headers(response, ["User-Agent"])
return response
```

Curious about Steve's opinion 🧐

slightlyoffbeat:
> _🚗 passing through without whole review_
> 
> I want to mention, that [Mozmeao Platforms Documentation on Redirects](https://mozmeao.github.io/platform-docs/development/redirects/#utilities) says "Do not set the `Vary: User-Agent` header; this will not work in production.".
> 
> Also, but not sure if relevant, we should use `patch_vary_headers()` maybe, because it preserves any existing `Vary` values rather than replacing the whole header.
> 
> ```python
> from django.utils.cache import patch_vary_headers
> 
> response = HttpResponseRedirect(redirect_url)
> patch_vary_headers(response, ["User-Agent"])
> return response
> ```
> 
> Curious about Steve's opinion 🧐

Good find, thanks @janriokrause 🙏

  Happy to go either way here pending @stevejalim's input

  - **Close the PR** if the doc warning applies to this
  - **Keep the change with a code comment** documenting the trade-off, if we want to preserve the signal for crawlers. Fastly may ignore `Vary`, but Googlebot and other search crawlers do read it, they may use it to decide to re-crawl with different UAs to discover the platform variants. The header is still semantically accurate even if Fastly doesn't act on it.

stevejalim:
The docs (and @janriokrause) are right. 

When we compute the CDN-level cache key, we fold in a number of things ([private repo link](https://github.com/mozilla/webservices-infra/blob/76916a0e7c8b525c20053ee2353f5f227b8155d4/springfield/tf/modules/fastly/vcl/main.vcl.tftpl#L122-L135)), but `Vary` is not one - so that we maximise cache hits. If we were to Vary on user-agent, the number of variations possible could practically result in the CDN being useless because the cache key is so variable between requests for the same page.

So, I would say close this. We _could_ keep the `vary:user-agent` header for googlebot, but do we actually serve different content for different UAs? I think we don't (we may sometimes conditionally _show_ different content but it's all in the same document, yep?) It's cleaner not to do this at all, unless the upside is significant

slightlyoffbeat:
This PR is specific for `/download/`, which behaves differently and redirects differently depending on UA...but I suppose the pages themselves are the same once a user lands on them?

stevejalim:
Yeah the pages are 99% the same when a user lands on them - all that varies is the platform name (Mac, Windows, Linux) in in some copy and the download link.

So, I think this isn't worth doing: Fastly ignores `Vary`, the destination pages don't really differ by UA, and `@never_cache` should be respected - non-compliant caches are kind of out of our firm control even with `Vary`, I'd argue.

Let's close this?

PR #1378, opened 2026-05-22
Apply `py-svg-hush` sanitization output to uploaded SVG (WT-1153)

janriokrause:
# Summary

* Use `py-svg-hush`'s sanitized output as the source of truth for uploaded SVG bytes, instead of only treating its parse result as a tripwire.
* Layers 1 (regex) and 2 (`defusedxml` walk) continue to hard-reject inputs that clearly look unsafe, so editors still get a clear error.
* Adds regression tests covering URL attributes whose dangerous scheme/MIME is hidden behind entity encoding – exercising the rewrite path the earlier layers cannot catch.

# Test plan

* [x] run `pytest springfield/cms/tests/test_svg_sanitization.py`
* [x] run `make test` for the full suite
* [x] Manually upload a known-clean SVG via the Wagtail image library and confirm it renders unchanged
* [x] Manually upload an SVG with an obvious `<script>` / `onclick` / cleartext `javascript:` payload and confirm the existing validation error still fires

I also tried `j%61vascript:alert(...)` against our SVG sanitizer as a probe. That's `javascript:` with the `a` percent-encoded as `%61`, a variant some older browsers used to decode. `py-svg-hush` stripped it cleanly. Didn't keep the test in the suite though. We can't exhaust every encoding variant a browser might decode, so the policy is to trust `py-svg-hush` for full URL-spec coverage and add regression tests only for specific patterns we've actually seen reported.

# References

* https://mozilla-hub.atlassian.net/browse/WT-1153
* https://bugzilla.mozilla.org/show_bug.cgi?id=2036460

### DIFF
--- a/springfield/cms/fields.py
+++ b/springfield/cms/fields.py
@@ -247,10 +247,12 @@ class SanitizingWagtailImageField(WagtailImageField):
         This uses a three-layer defense:
         1. Fast regex check for obvious dangerous patterns (scripts, event handlers)
         2. XML parsing to detect dangerous elements/attributes more reliably
-        3. py-svg-hush as a final safety net (but we don't reject based on normalization)
-        This approach avoids false positives from py-svg-hush's normalization
-        (attribute reordering, encoding changes) while still catching actual threats.
-        Returns None if file is safe, otherwise returns a ValidationError.
+        3. `py-svg-hush`, whose sanitized output replaces the uploaded bytes.
+        Layers 1 and 2 reject inputs that clearly look unsafe so editors get a
+        useful error. Subtler content is silently normalized by `py-svg-hush`,
+        which interprets SVG and URL syntax through a real parser rather than
+        by substring matching.
+        Returns `None` if file is safe (possibly after rewrite), otherwise a `ValidationError`.
         """
         try:
             # Read original content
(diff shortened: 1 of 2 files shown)


janriokrause:
⚠️ Heads up: `super().to_python(data)` validates upload size against the *original* bytes, before Layer 3 rewrites the file. If `py-svg-hush` ever produces output larger than the input, the stored file *could* exceed `FILE_UPLOAD_MAX_MEMORY_SIZE` without re-validation. In practice sanitization almost always shrinks, but worth either a one-line note in the docstring or a defensive `len(sanitized) <= len(original)` check.

💡 We could also reorder: sanitize first, then validate. Move `_sanitize_svg` before `super().to_python(data)`. Means parent validation runs against the rewritten bytes by default. Downside: we lose the parent's file-type detection that runs first, and we'd be sanitizing files that Wagtail might reject for other reasons (wasted work, and the call order in `to_python` is currently nicely layered: "parent validates basics, we sanitize on top").

I would prefer the defensive size check – or leave it as-is 😅

stevejalim:
> I would prefer the defensive size check – or leave it as-is 😅

Same here - it seems a light check to add, but if not, we can live without it.

stevejalim (approved):
Thanks for this - I appreciate the care here @janriokrause 

Tested locally and all fine.

I wondered whether there would be any problem using `seek`, `truncate` etc against the cloud-storage backend, but that'll be fine.

r+

There are a couple of comments from copilot about expanding tests - handy to do before merge, I think.

PR #1382, opened 2026-05-27
Skip inactive refactor tests

maureenlholland:
These are helpful for local development, but slowing down the integration tests CI too much in production. The switch remains OFF for production, so there is no harm in skipping these tests for now.

We'll need to reduce tests to maintain a reasonable playwright runtime in CI

## Testing
confirm tests are skipped locally: https://mozmeao.github.io/platform-docs/testing/playwright/?h=playw#installing-playwright

### DIFF
--- a/tests/playwright/specs/download-attribution/update.spec.js
+++ b/tests/playwright/specs/download-attribution/update.spec.js
@@ -21,7 +21,7 @@ test.beforeEach(async ({ page }) => {
     await removeDownloadAsDefault(page);
 });
 
-test.describe('analytics download attribution', () => {
+test.describe.skip('analytics download attribution', () => {
     test.describe(
         'essential data added',
         {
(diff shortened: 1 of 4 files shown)


wen-2018 (approved):

tests/playwright/specs/download-attribution/update.spec.js
```diff
@@ -383,7 +383,7 @@ test.describe('essential download attribution', () => {
     );
 });
 
-test.describe('conflict management', () => {
+test.describe.skip('conflict management', () => {
```
janbrasna, line 386: > "These are helpful for local development, but slowing down the integration tests CI too much in production."

Could be tagged e.g. `@local` and added to the grep-invert that selects tests in the manifest? (Or having different entrypoints/commands for CI vs. local with different args passed.)

PR #1386, opened 2026-05-28
Fix comments in roadmap.ftl

flodolo:
See https://github.com/mozmeao/springfield/pull/1367#pullrequestreview-4378948498


maureenlholland (approved):
r+

PR #1391, opened 2026-05-28
Restrict `cms/*` Fluent files in Pontoon to locales overlapping with CMS languages

janbrasna:
## One-line summary

Limits which Pontoon locales will get CMS–specific Fluent files exposed.
(See https://mozmeao.github.io/platform-docs/l10n/configuration/#pipeline-configuration for deets)

Might need maintaining over time to be in sync with languages added or removed in CMS.

Updates only Pontoon config to limit to the few that will get used. Smartling languages can't be limited, so these jobs will always trigger e.g. ar, ms or hi-IN that will end up unused.

## Issue / Bugzilla link

https://github.com/mozilla-l10n/www-firefox-l10n/pull/133#issuecomment-4563066282

## Testing

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -111,6 +111,25 @@ locales = [
 [[paths]]
     reference = "en/cms/*.ftl"
     l10n = "{locale}/cms/*.ftl"
+    # subset to locales set up in CMS
+    locales = [
+        "de",
+        "en-CA",
+        "en-GB",
+        "es-ES",
+        "es-MX",
+        "fr",
+        "id",
+        "it",
+        "ja",
+        "nl",
+        "pl",
+        "pt-BR",
+        "pt-PT",
+        "ru",
+        "tr",
+        "zh-CN",
+    ]
 [[paths]]
     reference = "en/banners/**/*.ftl"
     l10n = "{locale}/banners/**/*.ftl"
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -125,7 +125,6 @@ locales = [
         "nl",
         "pl",
         "pt-BR",
-        "pt-PT",
         "ru",
         "tr",
         "zh-CN",


maribedran (approved):

l10n/configs/pontoon.toml
```diff
@@ -111,6 +111,25 @@ locales = [
...
+        "es-ES",
+        "es-MX",
+        "fr",
+        "id",
+        "it",
+        "ja",
+        "nl",
+        "pl",
+        "pt-BR",
+        "pt-PT",
```
janbrasna: Actually, one less:
```suggestion
        "pt-BR",
```

(aliased)

l10n/configs/pontoon.toml
maribedran, line 1: We noticed that ftl translations don't apply the alias logic. It's probably fine to remove the alias languages from the list here since the fix will land before we start to use the status labels on the roadmap, which originated this thing.

janbrasna: Oooh. So it falls through all the way to `en` if missing?

I hoped this will be caught on the Django/l10n tooling level, that also has similar pair defined to "fill in" the blanks — but given I really don't know at which point this kicks in on the view where this is used… I'm totally deferring to your judgement (either way, whether to add AR+CL & PT here if favorable, or keep them out if the plan is to _not_ rely on any customized strings even for locales aliased "on the top level".)

My take was simply to bother as little teams as possible, if ultimately unused.

maribedran: Those labels won't be used for a while, so I think we'll have enough time to look into the alias for ftl translations. I only noticed the issue when I introduced those strings for the Roadmap page.

janbrasna: (Right, it probably doesn't even make sense to be prioritizing non-aliased real translation for a UI element, if the whole of the content around it was in a different language locale to start with. Point.)

This seems to have worked now for restricting, in case if need be the few extra locales can be added later as a stopgap if no other measure is in place in time to get rid of the English fallbacks. 💪

janbrasna:
This seems to have worked correctly:

> pontoon-nl-vs-fy

PR #1454, opened 2026-06-02
Fallback unsupported alpha WebM to poster (safari)

slightlyoffbeat:
Detect browsers that don't support alpha WebM (Safari / iOS) and replace alpha WebM video elements with their poster images. Adds helper functions isAlphaWebmUnsupported, isAlphaWebm, and swapAlphaWebmForPoster which finds .fl-video containers with a -alpha.webm source, creates an  from the video poster (preserving alt text), removes play buttons and clears the fl-animation-playing class. swapAlphaWebmForPoster() is invoked from setupAnimations to provide a graceful fallback for affected browsers.

## Testing
Added this to assets for testing: https://assets.mozilla.net/wc/kit-balancing-1-alpha.webm

  **Setup**

  1. In Wagtail admin, create or edit a page with an Animation block.
  2. Set Animation URL to: `https://assets.mozilla.net/wc/kit-balancing-1-alpha.webm`
  3. Set a poster image and alt text. Save & publish.

  **Test cases**

  | # | Browser | Expected |
  |---|---|---|
  | 1 | Chrome desktop | Animation plays with full transparency |
  | 2 | Firefox desktop | Animation plays with full transparency |
  | 3 | Safari desktop (macOS) | Static poster image shown — no video, no broken black box |
  | 4 | Mobile Safari (iPhone) | Static poster image shown |
  | 5 | Safari on iPad with "Request Desktop Site" enabled | Static poster image shown |
  | 6 | Chrome on iOS | Static poster image shown |

  **Regression check**

  7. Repeat steps 1–4 with any **non**-alpha webm (a URL that doesn't end in `-alpha.webm`) — animation should play in all browsers including
  Safari, no fallback triggered.

  **DOM verification (optional)**

  In Safari devtools on the affected page:
  - Inspect `.fl-video` — it should contain ``, not a `<video>` element
  - No `.fl-video-play` button inside the container


janriokrause:
Any more details on how to test this?

media/js/cms/components/flare26-animations.es6.js
```diff
@@ -4,6 +4,43 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+function isAlphaWebmUnsupported() {
+    const ua = navigator.userAgent;
+    if (/iPad|iPhone|iPod/.test(ua)) return true;
```
janriokrause, line 9:
```suggestion
    if (/iPad|iPhone|iPod/.test(ua)) return true;
    // iPadOS 13+ may use a desktop-class UA (contains "Macintosh") to request desktop sites.
    // Use touch points to distinguish it from real macOS Safari.
    if (/Macintosh/.test(ua) && navigator.maxTouchPoints > 1) return true;
```

Note that starting around iPadOS 13, Safari on iPad often uses a *desktop-class* User-Agent (to get desktop sites), so it may include `Macintosh` and omit `iPad`. In that case `/iPad|iPhone|iPod/` can miss iPads, so a common mitigation is also treating `Macintosh` + `navigator.maxTouchPoints > 1` as iPadOS.

media/js/cms/components/flare26-animations.es6.js
```diff
@@ -4,6 +4,43 @@
...
+
+function swapAlphaWebmForPoster() {
+    if (!isAlphaWebmUnsupported()) return;
+
+    document.querySelectorAll('.fl-video').forEach(function (container) {
+        const video = container.querySelector('video');
+        if (!video) return;
+
+        const source = video.querySelector('source');
+        if (!source || !isAlphaWebm(source.src)) return;
```
janriokrause:
```suggestion
        if (!source || !isAlphaWebm(source.getAttribute('src') || source.src)) return;
```

Rationale: `getAttribute('src')` reads the literal markup value (good for filename-convention checks like `-alpha.webm`), while `source.src` is the resolved URL and can sometimes be empty/unexpected depending on timing/how the element is populated.

janriokrause (commented):
Looks good to me with some non-blocking comments.

media/js/cms/components/flare26-animations.es6.js
slightlyoffbeat: Good catch - added.

media/js/cms/components/flare26-animations.es6.js
slightlyoffbeat: Applied. Agreed that the literal markup value is the safer read for a filename-convention check - getAttribute('src') first, falling back to .src only if missing.

slightlyoffbeat:
> Any more details on how to test this?

Added to desc

janriokrause (approved):
r+ Thank you! Tested all 6 test cases, the 7th regression test and DOM verification.

PR #1459, opened 2026-06-03
kick page CSS revisions

kkellydesign:
## One-line summary
* Kit animated hero is ready. Upload the video as an animation, and the background image is in the repo
* Font sizes are fixed in cards
* kit banner has the soccer background image


springfield/cms/blocks.py
```diff
@@ -1118,7 +1118,7 @@ def MixedButtonsBlock(
     button_types: list,
     min_num: int,
     max_num: int,
-    themes=BUTTON_THEMES_2025,
+    themes=BUTTON_THEMES_2026,
```
maureenlholland, line 1121: I defer to your judgement here, but is this potentially going to break any uses of this block relying on `BUTTON_TERTIARY`
Image

media/css/cms/pages/flare26-kick-page.css
```diff
@@ -14,3 +14,50 @@
 .fl-pre-footer-cta .fl-pre-footer-cta-button::before {
     content: unset;
 }
+
+.fl-featured-image-section-media .fl-video {
```
maureenlholland, line 18: this is all scoped to the kick page, so safe to update however is needed for that content specifically 👍

maureenlholland (approved):
r+wc

slight concern on swapping out the block button themes, because it is not a completely additive change

media/css/cms/pages/flare26-kick-page.css
kkellydesign: Yeah, I would've used the page-specific body class if this CSS file wasn't isolated to this page. In this case, we're ok 👍

PR #1460, opened 2026-06-03
Wrap home kit banner buttons in .fl-buttons

slightlyoffbeat:
Fixes minor regression on button on homepage (padding/margin).

Wrap the button include loop in a <div class="fl-buttons"> container in home-kit-banner.html so the rendered buttons can be targeted as a group for styling/layout (e.g., flex or spacing). No behavioral changes to the button blocks themselves—just markup grouping and indentation cleanup.

Before:
image

After:
image


janriokrause (approved):
r+

PR #1465, opened 2026-06-03
[WT-1289] Fix main nav menus accessibility

knowler:
## One-line summary

The current nav menu is not fully keyboard accessible and has some issues for screen readers, so this PR fixes those issues without modifying the presentation or interaction pattern (except it removes the redundant close buttons).

- **Since, the mobile and desktop views use the same markup (the major constraint for this approach), we need to apply the appropriate semantics for each view using JavaScript.**
    - On desktop, the category titles should be expand/collapse buttons (i.e. ARIA `button` role with an `aria-expanded` state).
        - `aria-haspopup` should not be used as these nav menus are not ARIA menus. `aria-haspopup=true` is a backwards compat equivalent to `aria-haspopup=menu` which expects that the target an ARIA `role=menu`. Those have particular expectations for focus management and keyboard shortcuts. These are not advisable for a site navigation.
    - On mobile, the category titles can just be text or a link.
        - Ideally, for this view they should be headings, but the entire mobile menu should be a modal `<dialog>`. For now, we can just present them as text/links.
    - It is easier to layer on button semantics on a `<a>` element than link/presention semantics on a `<button>`, so I’m using an `<a>` element for each of the category titles. Only *Features* is actually a link (i.e. it has an `href`). **On desktop, ARIA is used to apply button semantics.**
    - *In the future, it might be worth reconsidering if we want to use the same markup for the mobile and desktop nav, especially considering that we should be using a modal dialog for the mobile menu. There are some tricks using shadow DOM composition that could be leveraged to avoid markup duplication (e.g. re-slotting the light DOM content into a desktop/mobile specific structures).*
- This uses event delegation to avoid needing to add/remove many event listeners, especially as mobile and desktop have a different UX.
- The **close buttons were removed**. These provide little to no benefit for keyboard or assistive tech users as:
    1. The panels auto-close when focus leaves the panel.
        - Even if they didn’t auto close the original trigger can be used to close the panel and other triggers would cause the panels to close.
    2. The panels can be closed with <kbd>Escape</kbd> key — which is a standard user expectation.
- When opened with a keyboard, the “hover-to-open” interaction is ignored. The production site doesn’t do this, but it is quite annoying and unexpected.

## Issue / Bugzilla link

- https://mozilla-hub.atlassian.net/browse/WT-1289

## Testing

Compare against the current production deploy:

- Test keyboard navigation using the tab key and the <kbd>Escape</kbd> key.
- Test with a screen reader such as NVDA/JAWS/Narrator on Windows or VoiceOver on iOS or macOS or TalkBack on Android.

### DIFF
--- a/media/css/cms/components/flare-navigation.css
+++ b/media/css/cms/components/flare-navigation.css
@@ -190,6 +190,13 @@ html {
     text-decoration: none;
 }
 
+.fl-menu-title:where(button) {
+    background: none;
+    border-radius: 0;
+    margin-block-end: 0;
+    padding: 0;
+}
+
 .fl-menu-subtitle {
     font-family: var(--fl-theme-font-family-headline);
 }
--- a/media/css/cms/components/flare-navigation.css
+++ b/media/css/cms/components/flare-navigation.css
@@ -190,11 +190,15 @@ html {
     text-decoration: none;
 }
 
-.fl-menu-title:where(button) {
+.fl-menu-title:where(button, [role='button']) {
     background: none;
     border-radius: 0;
     margin-block-end: 0;
     padding: 0;
+}
+
+.fl-menu-title:is(a:not(:any-link):hover) {
+    text-decoration: none;
 }
 
 .fl-menu-subtitle {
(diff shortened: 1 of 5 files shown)


lucianoratamero (approved):
tested locally both with keyboard and voiceover

PR #1468, opened 2026-06-04
Kick page QA

kkellydesign:
I'm going to keep a running PR here for QA. We can merge this whenever, but no need to deploy 100x times today. 

* Illustration card gap is now consistent between animations and images
Screenshot 2026-06-04 at 9 04 15 AM


maureenlholland (approved):
r+

PR #1469, opened 2026-06-04
WT-856 - Convert Other Buttons To Use Pretranslated Text

dchukhin:
## One-line summary
Similar to #1449, this pull request updates all non-DownloadButton buttons to support pretranslated phrases.

Similar to #1250, this pull request updates all buttons:
 - a management command (called from a migration) moves all current `label`s into the button's `custom_label` field
 - buttons support a foreign key to a `PretranslatedPhrase`, so that Wagtail users can choose text that does not need to go to Smartling for translation

## Issue / Bugzilla link
discussion: https://github.com/mozmeao/springfield/pull/1250#issuecomment-4547875333

## Testing
After running migrations, view any current page with buttons, and observe that they still appear the same as before in each language (no text should be lost in any language).

Note: this pull request assumes that all current non-downloadbutton buttons' text is custom (does not match "Get Firefox" or "Download Firefox"), but in the future, any button can choose from the pretranslated phrases.


maribedran (approved):

stevejalim:
Needs https://github.com/mozmeao/springfield/pull/1250/ to merge first

PR #1473, opened 2026-06-04
WT-1252 expand domains for UTM addition

wen-2018:
## One-line summary

This PR expands domains for UTM addition.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1252

## Testing

1. edit any CMS page in Wagtail admin
2. update a URL to use any of the added domain
3. preview the link on the edited page
4. link should have the page slug as UTM campaign

### DIFF
--- a/springfield/cms/templatetags/cms_tags.py
+++ b/springfield/cms/templatetags/cms_tags.py
@@ -103,7 +103,8 @@ def remove_tags(value: str) -> str:
 def add_utm_parameters(context: dict, value: str) -> str:
     """
     Appends UTM parameters to URLs that point to *.mozilla.org, *.mozillafoundation.org,
-    and *.firefox.com domains, except for www.firefox.com.
+    *.firefox.com, *.mozilla.ai, *.mozilla.vc, *.thunderbird.net, and *.mozilla.com
+    domains, except for www.firefox.com.
     """
 
     utm_parameters = context.get("utm_parameters", {})
@@ -116,7 +117,7 @@ def add_utm_parameters(context: dict, value: str) -> str:
         elif parsed_url.path:
             host = parsed_url.path.split("/")[0]
         pattern = re.compile(
-            r"^(\w+\.)?((mozilla.org)|(mozillafoundation.org)|(firefox.com))",
+            r"^(\w+\.)?((mozilla.org)|(mozillafoundation.org)|(firefox.com)|(mozilla.ai)|(mozilla.vc)|(thunderbird.net)|(mozilla.com))",
             re.IGNORECASE,
         )
         if host and host not in ["www.firefox.com", "firefox.com"] and pattern.match(host):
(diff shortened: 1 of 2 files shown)


springfield/cms/templatetags/cms_tags.py
```diff
@@ -103,7 +103,8 @@ def remove_tags(value: str) -> str:
 def add_utm_parameters(context: dict, value: str) -> str:
     """
     Appends UTM parameters to URLs that point to *.mozilla.org, *.mozillafoundation.org,
-    and *.firefox.com domains, except for www.firefox.com.
+    *.firefox.com, *.mozilla.ai, *.mozilla.vc, *.thunderbird.net, and *.mozilla.com
+    domains, except for www.firefox.com.
```
janriokrause:
```suggestion
    domains, except for www.firefox.com and firefox.com.
```

janriokrause:
The regex is still structurally unsafe, but that was already the case before your PR. A `.` means "any character" in regex. Also, the regex has no end anchor, so it can match the start of a malicious-looking hostname like: `mozilla.com.evil.example`, `example.mozilla.com.evil.example`, `mozillaXcom`. Low impact because the function only appends UTM parameters.

springfield/cms/templatetags/cms_tags.py
```diff
@@ -116,7 +117,7 @@ def add_utm_parameters(context: dict, value: str) -> str:
         elif parsed_url.path:
             host = parsed_url.path.split("/")[0]
         pattern = re.compile(
-            r"^(\w+\.)?((mozilla.org)|(mozillafoundation.org)|(firefox.com))",
+            r"^(\w+\.)?((mozilla.org)|(mozillafoundation.org)|(firefox.com)|(mozilla.ai)|(mozilla.vc)|(thunderbird.net)|(mozilla.com))",
```
janriokrause:
```suggestion
            r"^(\w+\.)?((mozilla\.org)|(mozillafoundation\.org)|(firefox\.com)|(mozilla\.ai)|(mozilla\.vc)|(thunderbird\.net)|(mozilla\.com))\Z",
```

If the list is getting longer, I would rather prefer having a allow/exclude list than a hard to maintain regex.

springfield/cms/templatetags/cms_tags.py
```diff
@@ -116,7 +117,7 @@ def add_utm_parameters(context: dict, value: str) -> str:
...
             re.IGNORECASE,
         )
```
janriokrause, line 122:
```suggestion
        )
        host = host.lower().rstrip(".")
```

It would also be good to normalize `host` before the `firefox.com` exclusion check, since the regex is case-insensitive but the exclusion list is not.

janriokrause (approved):
r+ with my requested changes applied. Waiting for CI, then merge.

PR #1476, opened 2026-06-04
[WT-1294] AI Controls conditional display

maribedran:
## One-line summary

Add AI Controls options to the conditional display component and the Pin to Taskbar option to the UI Tour button.

- Use `Mozilla.UITour.getConfiguration('aiControls')` to determine if the user has AI Controls enabled for a new "AI Controls" option in the conditional display component
- Use `Mozilla.UITour.pinToTaskbar` for a new option in the UI Tour button block
- Fix the Default Browser conditional display (the JS that adds the correct classes to the HTML element was only called when a Set as Default button existed on the page)

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1294
https://mozilla-hub.atlassian.net/browse/WT-1295

## Testing

- Open the http://localhost:8000/en-US/tests-index-page-2026/test-conditional-display/ page
  - On Firefox <152, neither the "AI Controls: available" nor the "AI Controls: unavailable " notifications should be displayed
  - On Firefox >=152:
    - With enabled AI Controls: the "AI Controls: available" notification should be displayed and the "AI Controls: unavailable " shouldn't
    - With disabled AI Controls:  the "AI Controls: available" notification should not be displayed and the "AI Controls: unavailable " should
  - Check for the "Firefox is default browser " or "Firefox is NOT default browser" notifications depending on your default browser settings
- Open http://localhost:8000/en-US/tests-index-page-2026/test-buttons-2026/ on Firefox >= 152 on  Windows or Mac
  - Open `about:config` and make sure that UI Tour is enabled
  - Check that the Pin to Taskbar button works

Available:
image
image

Unavailable
image
image

Firefox is the default browser
image
image

Pin to Taskbar
image
image


media/js/cms/components/flare26-conditional-display.es6.js
```diff
@@ -0,0 +1,103 @@
...
+        document.documentElement.setAttribute('data-firefox-version', version);
+
+        const conditionalEls = document.querySelectorAll(
+            '.condition-fx-version'
+        );
+        conditionalEls.forEach((el) => {
+            const minVersion = el.dataset.minVersion;
+            const maxVersion = el.dataset.maxVersion;
+            let show = false;
+            if (parseFloat(minVersion) && version >= parseFloat(minVersion)) {
```
lucianoratamero: quick question about this: do you always want just the big number part (151, for example), or do you want the whole thing (151.0.3)? because if you want the whole thing, I'm not sure if parseFloat is the best idea here. there might be a scenario in which they use something like "152.beta" or anything that has a word in it, and if that happens, it will strip the word part of it, and we might get "152" instead of "152.beta" when comparing versions.

if that scenario is not real, this is 100% fine 👍 but if you want to check for the whole thing, comparing strings directly might be more useful.

media/js/cms/components/flare26-conditional-display.es6.js
```diff
@@ -0,0 +1,103 @@
...
+        if (conditions.includes(countryCode.toUpperCase())) {
+            el.classList.add('geo-match');
+        } else {
+            el.classList.remove('geo-match');
+        }
+    });
+}
+
+function initDefaultBrowserConditionalDisplay() {
+    isUITourEnabled()
```
lucianoratamero, line 63: we should probably change the implementation on the set as default js to use this promise.

lucianoratamero (changes requested):
I've sent you some tests I've done on slack too :)

lucianoratamero (approved):
looks good! just wait for CI and feel free to merge 🚀

PR #1477, opened 2026-06-04
WT-1262: Custom logo animation

lucianoratamero:
## One-line summary

This PR adds the necessary code to have a custom logo animation on hover, for browsers that support alpha webm.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1262

## Testing

Edit a free form page to add the flare26-kick-page body class, then access it on Firefox/Chrome and Safari.
It should work nicely on FF/Chrome, and fallback to a static logo on Safari.

### FF
CleanShot 2026-06-04 at 18 45 06

### Chrome
CleanShot 2026-06-04 at 18 46 45

### Safari
CleanShot 2026-06-04 at 18 47 37


media/css/cms/components/flare26-two-column-cards.css
```diff
@@ -99,6 +99,10 @@
     gap: var(--token-spacing-lg);
 }
 
+.fl-two-column-card-content .fl-heading-size-4 {
```
maribedran: This component shouldn't be using an `h4`. I noticed that bug recently with the heading level logic. The right selector here should be the general `fl-heading`.

maribedran (approved):

PR #1479, opened 2026-06-05
[WT-1250] Use `text-transform: uppercase` conditionally by language

knowler:
## One-line summary

`text-transform: uppercase` can have a negative effect for some locales, so instead we’ll apply it conditionally using `:lang()` and style queries (since `letter-spacing` is often paired with `text-transform`).

This change only affects the Flare26 styles.

- `:lang(en)` works as a wildcard by default, so `en` will match `en-US`, `en-CA`, `en-GB`.
- Future syntax changes to `:lang()` that could be useful:
  - We can use comma-separated `:lang()` once it’s widely supported (e.g. `:lang(en, fr)`. Currently only Firefox and Safari supports this. https://caniuse.com/mdn-css_selectors_lang_argument_list
  - We can use a different wildcard format to target text based on script or other factors (e.g. `:lang(*-Latn)` will match latin script). Currently, Firefox and Safari support this. https://caniuse.com/mdn-css_selectors_lang_wildcards
- Uses a style query for the custom property matching `--lang-supports-uppercase: true`.
  - As style queries only recently became Baseline Newly Available as of Firefox 151 (https://caniuse.com/css-container-queries-style — the partial support is likely due to style queries originally being spec’d to support more than just custom properties, but custom properties are supported in all Baseline browsers), this means **Firefox versions prior to a few weeks ago will get the non-uppercased design. Since this is just stylistic, I think this is a good use of progressive enhancement.**
  
Future uses of uppercased text should use:

```css
.thing {
  @container style(--lang-supports-uppercase: true) {
    text-transform: uppercase;
    /* Then any additional design considerations for the uppercased text */
    letter-spacing: 0.02em;
  }
}
```

This is something we could add to the docs and perhaps create a lint rule for (I don’t know how easy it is to do custom stuff like that with Stylelint).

## Issue / Bugzilla link

- https://mozilla-hub.atlassian.net/browse/WT-1250
- https://github.com/mozmeao/springfield/issues/1426

## Testing

- Affected pages (incomplete list):
  - Global footer small legal links at the very bottom.
  - `/download/all/`
  - `/channel/desktop/developer/` “Firefox Developer Edition” eyebrow above “The browser made for developers” in second section.
  - `/features/protection/`, `/features/control/`, `/features/focus/`, `/user-privacy/` eyebrow in the main heading of the hero.

### DIFF
--- a/media/css/cms/components/flare26-footer.css
+++ b/media/css/cms/components/flare26-footer.css
@@ -219,7 +219,10 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
     max-inline-size: var(--token-width-desktop-layout);
     padding: var(--token-spacing-lg);
     position: relative;
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
 
 .fl-footer-secondary a:hover,
--- a/media/css/cms/flare26-defaults.css
+++ b/media/css/cms/flare26-defaults.css
@@ -17,6 +17,17 @@ theme layer.
 
 html {
     font-size: var(--fl-theme-font-size-body-sm);
+
+    /**
+     * Only allow uppercase text transform for supported languages.
+     *
+     * Use a container style query to apply uppercase and related styles.
+     *
+     * See: https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#css-issues
+     */
+    &:is(:lang(de), :lang(en), :lang(fr)) {
+        --lang-supports-uppercase: true;
+    }
 }
 
 body {
--- a/media/css/cms/flare26-utilities.css
+++ b/media/css/cms/flare26-utilities.css
@@ -44,7 +44,9 @@ style rules should have any fallbacks needed for IE8+ and Safari 12.1+
 }
 
 .fl-u-text-uppercase {
-    text-transform: uppercase;
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
 
 .hidden {
--- a/media/css/cms/components/flare26-article.css
+++ b/media/css/cms/components/flare26-article.css
@@ -134,7 +134,10 @@ Stacked article list
 .fl-article-item .fl-superheading {
     color: var(--fl-theme-color-brand-text);
     font-size: var(--fl-theme-label-xs);
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
 
 /* =====================================================================
@@ -179,7 +182,10 @@ Related articles
 
 .fl-related-articles-header-link {
     font-size: var(--fl-theme-font-size-body-2xs);
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
 
 @media (--viewport-md-up) {
--- a/media/css/cms/components/flare26-banner.css
+++ b/media/css/cms/components/flare26-banner.css
@@ -126,10 +126,13 @@
 .fl-banner .fl-superheading {
     color: var(--fl-theme-color-link);
     font-size: var(--fl-theme-label-caps-sm);
-    letter-spacing: 0.0175rem;
     line-height: 1.2;
     margin-block-end: var(--token-spacing-md);
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        letter-spacing: 0.0175rem;
+        text-transform: uppercase;
+    }
 }
 
 .fl-banner-media {
--- a/media/css/cms/components/flare26-card-gallery.css
+++ b/media/css/cms/components/flare26-card-gallery.css
@@ -39,7 +39,10 @@ Card Gallery - Flare 2026 Design System Card Gallery Component
     color: var(--fl-theme-card-superheading-color);
     font-size: var(--fl-theme-font-size-body-2xs);
     margin: 0;
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
 
 .fl-card-gallery-card .fl-buttons {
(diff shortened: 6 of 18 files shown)


media/css/cms/flare26-defaults.css
```diff
@@ -30,6 +30,19 @@ body {
...
+:lang(fr) {
+    --optional-uppercase: uppercase;
+}
+
+:not(
+    :lang(en),
+    :lang(fr)
+) {
+    /* Use for `letter-spacing` and include uppercase `letter-spacing` value as a fallback. */
+    --optional-uppercase-letter-spacing: unset;
```
maureenlholland: Bit of a wild swing here (with the huge caveat I haven't actually tested it), but this approach is in my head, curious to know your thoughts:

1. Default styling: no uppercase, no uppercase letter spacing
2. Register `--optional-uppercase` `@property` with initial value of `none`
3. As you've got on L33, allowed uppercase languages set a new property value
4. Style query when `--optional-uppercase` is not initial value to apply the "if uppercase style" letter spacing 

`@property` support: https://caniuse.com/mdn-css_at-rules_property
Style queries support: https://caniuse.com/css-container-queries-style

Downsides: could be wildly over-engineered, still a while away from wide baseline support, might be cross-browser bugs in current support

knowler: I don’t think that’s overengineered at all — this is the perfect application of style queries. We can actually get away without the `@property` since that’s not required for a style query. Something like:

```css
:lang(en) {
  --lang-supports-uppercase: true;
}

.thing {
  @container style(--lang-supports-uppercase: true) {
    text-transform: uppercase;
    letter-spacing: 0.02em;
  }
}
```

This would work nicely as a progressive enhancement and like you’ve suggested, since the default styling would not apply the uppercase text, the older nonsupporting browsers with languages that don’t play well with uppercase text would be safe.

knowler: Resolved in https://github.com/mozmeao/springfield/pull/1479/commits/e0a619970e6bc040479cfc2af26ff256e4e70485

media/css/cms/components/flare26-footer.css
```diff
@@ -219,7 +219,10 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
     max-inline-size: var(--token-width-desktop-layout);
     padding: var(--token-spacing-lg);
     position: relative;
-    text-transform: uppercase;
+
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
```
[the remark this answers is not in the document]
knowler: No `container-type` (or `container-name`) is required for container style queries as style queries do not require any specific containment to be enabled. Since unregistered custom properties inherit, the query container in this situation will always be the parent.

From [the CSS Containment Module Level 3](https://www.w3.org/TR/css-contain-3/#query-container:~:text=Unless,type%2E):

> Unless otherwise noted, all elements are [query containers](https://www.w3.org/TR/css-contain-3/#query-container) for the purpose [container queries](https://www.w3.org/TR/css-contain-3/#container-query) that do no require explicit containment (such as [container style queries](https://www.w3.org/TR/css-contain-3/#container-style-query)), reagardless of the specified [container-type](https://www.w3.org/TR/css-contain-3/#propdef-container-type).

Also, I have specifically (manually) tested the footer.

media/css/cms/flare26-utilities.css
```diff
@@ -44,7 +44,9 @@ style rules should have any fallbacks needed for IE8+ and Safari 12.1+
 }
 
 .fl-u-text-uppercase {
-    text-transform: uppercase;
+    @container style(--lang-supports-uppercase: true) {
+        text-transform: uppercase;
+    }
 }
```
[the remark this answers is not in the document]
knowler: All elements are query containers for the sake of style queries. For extended answer, see: https://github.com/mozmeao/springfield/pull/1479#discussion_r3389918484

media/css/cms/flare26-defaults.css
```diff
@@ -30,6 +30,12 @@ body {
     padding: 0;
 }
 
+:lang(de),
+:lang(en),
+:lang(fr) {
+    --lang-supports-uppercase: true;
+}
```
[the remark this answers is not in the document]
knowler: This is true. Generally, I would leave this as not specific to the `html` element since it would account for parts of the page being in a different language. I don’t think that would be super relevant for this project though. I imagine just one `lang` attribute is ever set for the entire page.

If this change is accepted, we can use `html:is()` or `html:where()` rather than duplicating the type selector for each language.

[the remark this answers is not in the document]
knowler: Resolved in https://github.com/mozmeao/springfield/compare/726891ddff75c15491b344a32ed4e121a55e73d1..cbd989692de312010cf5d7d883df1370538bf19d

media/css/cms/flare26-defaults.css
```diff
@@ -17,6 +17,10 @@ theme layer.
 
 html {
     font-size: var(--fl-theme-font-size-body-sm);
+
+    &:is(:lang(de), :lang(en), :lang(fr)) {
```
maureenlholland, line 28: **non-blocking, maintenance**: might be a good place for a comment that links to the l10n docs to explain why we only conditionally apply uppercase

maureenlholland (approved):
r+ ✨ 

confirmed uppercase styling & letter-spacing applied where expected across current Chrome/Safari/Firefox

uppercase styling does not apply for browsers that don't support style queries or locales that are not part of the "allow uppercase" lang list

media/css/cms/flare26-defaults.css
knowler: Good idea! Added the comment here and here:

- https://github.com/mozmeao/springfield/compare/3415b8faa224efabf8827c7fcd074d0b273ed3fe..e2d4710c7118af686953f5ef059aad8ae19bf8bd
- https://github.com/mozmeao/springfield/compare/e2d4710c7118af686953f5ef059aad8ae19bf8bd..5235044edfa987513adeb2d21041cbafe1169ff8

PR #1481, opened 2026-06-08
Revert essential campaign update

maureenlholland:
## One-line summary

Unnecessary update. Inline context for "essential" campaign and how it will evolve in future

## Testing
n/a - the campaigns are not used with `force` attribute in CMS


stevejalim (approved):

maureenlholland:
merging as the previous PR is not yet in prod, so we can make this update before next deploy

PR #1488, opened 2026-06-09
Fix failing local stub attribution code tests

maureenlholland:
## One-line summary

Fixes local tests after updates to default python stub attribution code values

Views update:
- Switch analytics granted check to look for session_id field rather than ga4 field (ga4 could be null and stripped out of request, session_id should always be set if analytics is granted)
- Only apply fallback values if analytics are granted and those fields are not set

Test update:
- When no session id, expect all fields except dlsource to be (not set)
- Add session_id param on tests where analytics fields should be set. In these cases, expectations should remain the same.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1323

## Testing
`pytest springfield/firefox/tests/test_views.py::TestStubAttributionCode` should show no errors

### DIFF
--- a/springfield/firefox/tests/test_views.py
+++ b/springfield/firefox/tests/test_views.py
@@ -45,8 +45,8 @@ class TestStubAttributionCode(TestCase):
 
     def test_no_valid_param_names(self):
         final_params = {
-            "source": "www.firefox.com",
-            "medium": "(none)",
+            "source": "(not set)",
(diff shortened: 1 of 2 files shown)


maureenlholland:
no valid param data/name still mismatched between CI & local (passing local, failing CI now)

stevejalim (approved):

PR #1493, opened 2026-06-09
fix webpack error

wen-2018:
## One-line summary

This PR fixes the webpack loading error caused by a Dependabot PR (#1212) bumped postcss-custom-media from v10 → v12. Version 12 is ESM-only ("type": "module"), but webpack.config.js loaded it with CommonJS require(), which Node refuses for .mjs ESM modules — crashing the dev server on startup.

## Testing

`npm start` should no longer show webpack loading error.


knowler (approved):
r+ pulled down, updated node deps, and ran `npm start` with no errors.

PR #1500, opened 2026-06-11
Fix console errors when UI Tour isn't available

maribedran:
## One-line summary

Verify that UI Tour is defined before calling it, and handle the promise rejection on the call that verifies if UI Tour is enabled.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1341

## Testing

Open any page in Chrome (UI Tour isn't available) and check that no errors show up in the console


maureenlholland (approved):
r+ tested with http://localhost:8000/en-US/whatsnew/148/ to check Fx & non-Fx works as expected ✅

PR #1502, opened 2026-06-11
WT-1344 inject Plausible analytics for EU visitors

stephendherrera:
## One-line summary

Inject the Plausible analytics script for EU/consent-country visitors only, gated behind a `plausible` waffle switch.

- **EU-only gating is server-side.** A new `plausible_enabled(country_code)` helper ([misc.py](springfield/firefox/templatetags/misc.py)) is the single source of truth — it returns true only when the `plausible` waffle switch is on, `PLAUSIBLE_DOMAIN` is configured, and the visitor is in a `DATA_CONSENT_COUNTRIES` (EU/EEA + CH + UK) country. Reused across all three base templates so non-EU visitors never receive the script or its data attributes.
- **Client script** ([plausible.es6.js](media/js/base/plausible/plausible.es6.js)) reads `data-plausible-domain` / `data-plausible-src` at call time, bails on DNT/GPC, defines the `window.plausible` queue stub, then injects the deferred script — following Plausible's documented install.
- **CSP** ([settings/__init__.py](springfield/settings/__init__.py)): the script host is added to `script-src` and `connect-src` only when `PLAUSIBLE_DOMAIN` is set. Note: assumes events post to the same host as the script (true for Plausible Cloud); a self-hosted instance ingesting on a separate host would need that added too.
- **Settings**: `PLAUSIBLE_DOMAIN` (default empty → disabled) and `PLAUSIBLE_SCRIPT_URL` (default `https://plausible.io/js/script.js`). Demo env var added but commented out so demos stay dark by default.

## Issue / Bugzilla link

[WT-1344](https://mozilla-hub.atlassian.net/browse/WT-1344)

## Testing

1. **Automated:** new Jasmine spec for the client script (DNT/GPC bail-out, queue stub, script injection, missing-attr guard) and `pytest` cases for the helper (EU on / non-EU off / switch off / no-domain off). Run `npm run jasmine` and `make test`.
2. **Manual (non-prod):** set `PLAUSIBLE_DOMAIN`, create+enable the `plausible` waffle switch, then:
   - Load any page with `?geo=DE` → confirm the Plausible `<script>` and `data-plausible-*` attributes are present in the `<head>`.
   - Load with `?geo=US` → confirm the script and attributes are **absent**.
   - With `?geo=DE` and DNT or GPC enabled in the browser → confirm the script does **not** load.


gcp/springfield-demos/cloudrun/cloudrun-demo.env.yaml
```diff
@@ -16,6 +16,7 @@ DEV: "True"
 GTM_CONTAINER_ID: "GTM-MSTZDDXN"
 LOG_LEVEL: INFO
 LOCAL_DB_UPDATE: "True"
+# PLAUSIBLE_DOMAIN: "firefox.com"
```
janriokrause: 🚗 No a full review, just passing through, but why did you comment it out? If default is disabled, maybe consider the var being empty/null.

Configuration should never live in comments. It creates ambiguity, breaks tooling that audits env vars, and makes the default behavior opaque. The codebase already has a documented default (empty string).

janriokrause: We can then also get rid of the `noqa: F405`.

slightlyoffbeat:
Took this for a spin locally

**Verification**

- Brought up runserver with `PLAUSIBLE_DOMAIN=firefox.com`.
- `?geo=DE` (and `FR`/`GB`/`CH`/`NO`): `data-plausible-domain`, `data-plausible-src`, and the `plausible` bundle `<script>` all render in `<head>` ✓
- `?geo=US` (and `CA`/`JP`): zero `plausible` markers on the page ✓
- CSP header includes `plausible.io` in both `script-src` and `connect-src` ✓
- Compiled bundle contains the expected `data-plausible-domain` / `data-plausible-src` / `data-domain` / `defer` literals ✓
- `pytest -k plausible` → 4/4 pass; `npm run jasmine` → 7/7 pass

**Asks before merge**

1. **Consolidate the four new helper tests into one `@pytest.mark.parametrize`**. Suggested:

   ```python
   @pytest.mark.parametrize(
       "country_code, domain, switch_on, expected",
       [
           ("DE", "firefox.com", True, True),
           ("US", "firefox.com", True, False),
           ("DE", "firefox.com", False, False),
           ("DE", "", True, False),
       ],
   )
   def test_plausible_enabled(country_code, domain, switch_on, expected):
       with override_settings(PLAUSIBLE_DOMAIN=domain), patch.object(misc, "switch", return_value=switch_on):
           assert misc.plausible_enabled(country_code) is expected
   ```

2. +1 to @janriokrause on the `cloudrun-demo.env.yaml` line

**Non-blocking, but worth resolving while it's cheap**

3. **Queue stub is skipped on DNT/GPC.** In `plausible.es6.js`, `init()` bails before `defineQueueStub()` when DNT or GPC is set, so `window.plausible` is never defined. Today that's fine (nothing calls it). The moment a follow-up PR adds `window.plausible('Some Event', ...)` somewhere, every EU + DNT visitor throws `TypeError: window.plausible is not a function` at that callsite. Plausible's own install docs define the stub unconditionally for exactly this reason. Two-line reorder:

   ```js
   Plausible.init = () => {
       Plausible.defineQueueStub();
       if (gpcEnabled() || dntEnabled()) {
           return;
       }
       Plausible.loadScript();
   };
   ```

Happy to take this as a follow-up if you'd rather keep PR scope tight, but it's cheap to do now and matches the upstream install pattern.

springfield/firefox/tests/test_helper_misc.py
```diff
@@ -1104,3 +1104,17 @@ def test_fxa_button_email(self):
 def test_needs_data_consent(country_code, expected):
     template = "{{ needs_data_consent('%s') }}" % country_code
     assert render(template) == expected
+
+
+@pytest.mark.parametrize(
```
maureenlholland, line 1109: 👏

maureenlholland (approved):
r+ thanks for the tests

janriokrause:
Apologize to bug again and thank you for the concern about commented-out config, but I think that approach went in the wrong direction. Maybe I wasn't clear enough with what I meant.

The issue wasn't that the setting existed in the YAML, it was how it was presented. Instead of uncommented-but-empty, it went with not-in-file-at-all. Both leave us with a problem: developers have no way to know this feature exists without reading Python code.

Why it matters: The YAML file is the source of truth for what this deployment needs. It's the infrastructure-as-code contract. If `PLAUSIBLE_DOMAIN` isn't listed there, it's invisible. And six months from now, we have to dig through `base.py` to find out.

Better design: Keep it in the YAML, uncommented, with an empty value and a maybe clarifying comment:

```yaml
# Analytics - disabled by default
PLAUSIBLE_DOMAIN: ""
PLAUSIBLE_SCRIPT_URL: "https://plausible.io/js/script.js"
```

And the corresponding `config()` call in `base.py`:

```python
# springfield/settings/base.py
PLAUSIBLE_DOMAIN = config("PLAUSIBLE_DOMAIN")
PLAUSIBLE_SCRIPT_URL = config("PLAUSIBLE_SCRIPT_URL")
```

This way:
* The parameter is **visible** in the config file
* It's clear what the default is (empty = disabled)
* Future maintainers know this feature exists
* To enable it in a different environment, you change the value. No code search needed

If there's no headspace now, I'm happy to tackle this as a follow-up PR.

PR #1503, opened 2026-06-12
WT-1327: What's next (Roadmap page) QA Fixes

lucianoratamero:
## One-line summary

This PR implements various fixes and changes to the Roadmap page. It previously had a refactor to the roadmap item buttons, but this change will be implemented in a separate PR.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1327

## Testing

Load the page fixtures and validate that every sub-item on the Jira ticket is implemented.


media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,194 @@
...
+        });
+
+        if (activeFilters.size > 0 && visibleCount === 0) {
+            section.classList.add('hidden');
+            if (itemsList) {
+                itemsList.classList.add('hidden');
+            }
+        } else {
+            section.classList.remove('hidden');
+        }
```
slightlyoffbeat, line 85: **Bug:** the `if` branch above adds `.hidden` to both `section` and `itemsList`, but this `else` only removes it from the section. Since `.hidden` is the global `display: none` utility ([flare26-utilities.css:52](https://github.com/mozmeao/springfield/blob/main/media/css/cms/flare26-utilities.css#L52)), once an items list is hidden it stays hidden for the rest of the page session.

**Repro:** select a filter that empties section A → switch to a different filter that re-populates A → A's header reappears with no items below it. Clear-all reproduces the same way.

```suggestion
        } else {
            section.classList.remove('hidden');
            if (itemsList) {
                itemsList.classList.remove('hidden');
            }
        }
```

media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,194 @@
...
+            toggle.addEventListener('click', () => {
+                const collapsed = section.classList.toggle('is-collapsed');
+                toggle.setAttribute('aria-expanded', String(!collapsed));
+            });
+        });
+    }
+
+    handleAccordion();
+    mq.addEventListener('change', handleAccordion);
+}
```
slightlyoffbeat, line 195: **Bug:** each call to `handleAccordion()` attaches a *new* click listener to every toggle. The function runs on init AND on every `change` event where `mq.matches` becomes true. Listeners are never removed, so they stack on every mobile→desktop→mobile round-trip.

**Repro:** load page <900px → click toggle (works). Resize ≥900px → resize back <900px. Click toggle — does nothing. Two listeners each flip `is-collapsed` once, net zero. Each subsequent round-trip adds another listener; odd counts = one net toggle, even counts = no-op.

Suggested fix: attach listeners once at init unconditionally, and gate the behavior on `mq.matches` inside the handler. Drop the `matchMedia` change listener entirely.

```suggestion
function initRoadmapAccordion() {
    const mq = window.matchMedia('not (min-width: 900px)');
    const toggles = document.querySelectorAll('.fl-roadmap-section-toggle');

    toggles.forEach((toggle) => {
        const section = toggle.closest('.fl-roadmap-list-section');
        if (!section) return;

        toggle.addEventListener('click', () => {
            if (!mq.matches) return;
            const collapsed = section.classList.toggle('is-collapsed');
            toggle.setAttribute('aria-expanded', String(!collapsed));
        });
    });
}
```

media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,194 @@
...
+        '.fl-roadmap-filter-button-clear-all'
+    );
+
+    if (clearAllButton && activeFilters.size) {
+        clearAllButton.disabled = false;
+        clearAllButton.querySelector('.fl-tag').classList.add('is-selected');
+    } else {
+        clearAllButton.disabled = true;
+        clearAllButton.querySelector('.fl-tag').classList.remove('is-selected');
+    }
```
slightlyoffbeat: The `if` guards `clearAllButton &&`, but the `else` immediately dereferences `clearAllButton.disabled` and `clearAllButton.querySelector(...)`. Today the template always renders clear-all alongside filter buttons so this won't crash in production — but it's a latent crash if anything (pattern library, A/B variant, future template change) ever renders filter buttons without clear-all.

Tightening this up also eliminates the duplicated `.fl-tag` lookup:

```suggestion
    if (clearAllButton) {
        clearAllButton.disabled = !activeFilters.size;
        const tag = clearAllButton.querySelector('.fl-tag');
        if (tag) {
            tag.classList.toggle('is-selected', Boolean(activeFilters.size));
        }
    }
```

springfield/cms/templates/cms/blocks/roadmap-list-section.html
```diff
@@ -15,10 +15,20 @@
...
 <include:roadmap-list-section>
   <content:heading>
-    <{{ heading_level }} class="fl-heading">{{ value.headline }}</{{ heading_level }}>
-    {% if value.subheadline %}
-      <p class="fl-subheading">{{ value.subheadline }}</p>
-    {% endif %}
+    <button
+      type="button"
+      class="fl-roadmap-section-toggle"
+      aria-expanded="true"
```
slightlyoffbeat, line 21: `aria-expanded` is hardcoded `"true"` and JS only updates it on click. Combined with the desktop CSS override at [`flare26-roadmap-list-section.css:307-310`](https://github.com/mozmeao/springfield/blob/WT-1327-whats-next-qa/media/css/cms/components/flare26-roadmap-list-section.css#L307-L310) that forces `.fl-roadmap-list` visible on desktop regardless of `.is-collapsed`, this leads to a screen-reader / visual mismatch:

**Repro:** on mobile, collapse a section (aria-expanded flips to `"false"`, `.is-collapsed` added). Resize to desktop — list visually reappears via the CSS override, but `aria-expanded` stays `"false"`. VoiceOver / NVDA announce the section as collapsed even though it isn't.

Cleanest fix is to reset both `aria-expanded` and `.is-collapsed` when the viewport leaves mobile. In the accordion handler (after fixing #2), add a `mq.addEventListener('change', ...)` that wipes the collapsed state on desktop, e.g.:

```js
mq.addEventListener('change', (e) => {
    if (!e.matches) {
        toggles.forEach((toggle) => {
            toggle.setAttribute('aria-expanded', 'true');
            toggle.closest('.fl-roadmap-list-section')?.classList.remove('is-collapsed');
        });
    }
});
```

(No suggestion block here since the fix lives in the JS, not this template line.)

media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,194 @@
...
+                tags.some((tag) => activeFilters.has(tag));
+
+            item.classList.toggle('hidden', !visible);
+
+            item.querySelectorAll('span[data-tag]').forEach((tagEl) => {
+                const tag = tagEl.querySelector('.fl-tag');
+                if (tag) {
+                    tag.classList.toggle(
+                        'is-selected',
+                        visible && activeFilters.has(tagEl.dataset.tag)
```
slightlyoffbeat: Minor: `is-selected` is stripped whenever the item is hidden, but the user's filter selection hasn't changed — the tag styling is then stale if anything ever re-shows the item without re-running `applyFilter`. Today there's no such path so it's latent, but cleaner to compute it independently:

```suggestion
                        activeFilters.has(tagEl.dataset.tag)
```

slightlyoffbeat:
Pulled this down and visually confirmed every change in the PR description: accordion behavior on mobile, clear-all button, primary button reorder, icon swaps, spacing tweaks, dark mode. Looks great.

Left a few inline comments on the JS, couple of bugs around filter-state and accordion listeners, plus an a11y note on aria-expanded. Nothing blocking the visuals, but worth a pass before merge.

slightlyoffbeat (approved):
Looks good. Only item is "clear all" is an English only string. Not blocking. Something we can follow up on.

R+ from me

springfield/cms/templates/components/roadmap-list-filter.html
```diff
@@ -17,18 +17,17 @@
...
-            </include:tag>
+            <include:tag text="{{ tag.label }}" before_icon={{ tag.icon }} after_icon="close" />
           </button>
         {% endfor %}
+        <button
+          type="button"
+          class="fl-roadmap-filter-button fl-roadmap-filter-button-clear-all"
+          aria-pressed="false"
+          disabled
+        >
```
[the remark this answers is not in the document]
slightlyoffbeat: fixed

springfield/cms/templates/components/roadmap-list-filter.html
```diff
@@ -17,18 +17,17 @@
...
-                <content:before_icon>
-                  <include:icon icon_name="{{ tag.icon }}" hidden=true extra_class="icon-left" />
-                </content:before_icon>
-              {% endif %}
-              <content:after_icon>
-                <include:icon icon_name="close" hidden=true extra_class="icon-right" />
-              </content:after_icon>
-            </include:tag>
+            <include:tag text="{{ tag.label }}" before_icon={{ tag.icon }} after_icon="close" />
           </button>
```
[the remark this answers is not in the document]
slightlyoffbeat: fixed

media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+/**
+ * Initializes the mobile accordion for roadmap sections.
+ * Only active below 900px viewport width.
+ */
+function initRoadmapAccordion() {
+    const mq = window.matchMedia('not (min-width: 900px)');
+    const toggles = document.querySelectorAll('.fl-roadmap-section-toggle');
```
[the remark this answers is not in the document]
slightlyoffbeat: disagree this is a blocker, but this does match current patterns better, so change is made.

media/js/cms/components/flare26-roadmap.es6.js
```diff
@@ -0,0 +1,204 @@
...
+            section.classList.add('hidden');
+            if (itemsList) {
+                itemsList.classList.add('hidden');
+            }
+        } else {
+            section.classList.remove('hidden');
+            if (itemsList) {
+                itemsList.classList.remove('hidden');
+            }
+        }
```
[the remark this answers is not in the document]
slightlyoffbeat: Disagree. Not a blocker. Consistent with current code. Sections with zero visible items are hidden entirely on both mobile and desktop, which is what currently ships and matches what the QA tests cover. 

We can discuss potential changes as a follow up.

slightlyoffbeat:
Pushed three small fixes from the Copilot pass.

One that copilot lists as 'blocking', I would disagree with. Can be a follow up.

Tracking the following as follow-ups (none blocking merge):

- **"Clear all" l10n** — currently hardcoded English; should move to a Fluent string.
- **`<button>` content model** in the section toggle — `<h2>`/`<p>` inside `<button>` is technically invalid HTML. Modern parsers don't actually break, but it affects screen-reader announcement and HTML validators. Cleanest fix is to invert nesting (`<button>` inside the heading, subheading as sibling outside). Applies to both the production template and pattern-library variant.
- **`applyFilter` docstring vs. behavior** — code hides the section entirely on both viewports; docstring suggests mobile keeps the header visible. Either align docstring with current behavior or revisit if the documented behavior is intended. Tied to a dead `.hidden .fl-roadmap-list` CSS rule that can be cleaned up at the same time.
- **Toggle button still focusable on desktop with no behavior** — minor a11y nit; could be wrapped into the same fix as the button-content-model item.
- **`secondary_button_*` named "Primary"** in `blocks.py` — already has a TODO referencing the button-list refactor; flagging for visibility.

PR #1507, opened 2026-06-15
WT-1215: Fixes for RTAMO download attribution

maureenlholland:
## One-line summary

Return to AMO functionality should work for `www.firefox.com` links

Template rendering
- when `s=direct` is in URL, all possible paths should render the `rtamo` template

FTL 
- `rtamo` template should use `download/desktop` FTL file

Validation
- the content is more important than the campaign for installer
   - when campaign is `rtamo`
       - essential only: add content param instead of campaign
       - essential and analytics: use UTM values (essential campaign will be used internally for validation, but UTM campaign will be passed to installer)

Order of operations
- if essential request happens first, and returns before analytics request is triggered, the download does not wait for analytics: it initiates as part of essential success callback, with essential-only information 
    - fix ensures essential-only flow is aware of incoming analytics request and does not proceed to make request. Only the combined essential + analytics request is made.

## Issue / Bugzilla link
https://mozilla-hub.atlassian.net/browse/WT-1215

## Testing

Local
- remove [document.referrer](https://github.com/mozmeao/springfield/pull/1507/changes#diff-ffdc92868013461974feefe6c79bc4314c1c3c78056d486551683bf72fda3552R650) and hardcode `'https://addons.mozilla.org'` 

These URLs check CMS => rtamo template render (`rtamo` in HTML force campaign attribute)

consent required
go to http://localhost:8000/en-US/thanks/?s=direct&utm_campaign=amo-fx-cta-607454&utm_content=rta%3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org&geo=FR 
- [x] `attribution_code` attached to download link should include essential only information (content, dlsource)

consent not required
http://localhost:8000/en-US/thanks/?s=direct&utm_campaign=amo-fx-cta-607454&utm_content=rta%3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org&geo=US
- [x] `attribution_code` attached to download link should include essential + analytics information  (all UTMs, dlsource, session id, ga4 client id)

The following two checks should show `rtamo` in an HTML force campaign attribute

Static (non-CMS) => rtamo template
- [x] http://localhost:8000/fi/thanks/?s=direct&utm_campaign=amo-fx-cta-607454&utm_content=rta%3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org

Static (non-CMS) basic experience => rtamo template
- [x] http://localhost:8000/fi/thanks/?s=direct&utm_campaign=amo-fx-cta-607454&utm_content=rta%3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org&xv=basic

Demo 4
Setup [extension](https://docs.google.com/document/d/1cAUFF9hozAST2JyUlb7kiSdvyfocbFdfpOFGiOj4vXg/edit?tab=t.0#heading=h.pv18ccwnp44g) [moz only] to test with production addons site: download links should point to `www-demo4.springfield.moz.works/thanks`

Go to https://addons.mozilla.org/en-US/firefox/addon/amo-info/ in non-Firefox browser

Click download and check the following
- [x] the rendered demo page has `rtamo` campaign in HTML data attributes
- [x] the download auto-starts (network request has expected `attribution_code` param)
- [x] the try again download link has an `attribution_code` param that decodes to show content prefixed with `rta`

After deployed to prod (easier to test referer header from `www.firefox.com`)
- [ ] when download is loaded to [attribution reader](https://williamdurand.fr/fx-attribution-data-reader/), it shows the same attribution code data (i.e. content prefixed with `rta`)
- [ ] when download is opened, there is special onboarding to install the extension from the adds ons page
    - note: this requires the download link to have a [valid referrer](https://github.com/mozilla-services/stubattribution/blob/master/attributioncode/validator.go#L20) (`www.firefox.com`)

### DIFF
--- a/springfield/firefox/tests/test_views.py
+++ b/springfield/firefox/tests/test_views.py
@@ -357,6 +357,7 @@ class TestStubAttributionCode(TestCase):
 
 
 @override_settings(DEV=False)
+@override_switch("ENABLE_ATTRIBUTION_REFACTOR", active=True)
 @patch("springfield.firefox.views.l10n_utils.render", return_value=HttpResponse())
 class TestFirefoxDownload(TestCase):
     def test_post(self, render_mock):
@@ -462,7 +463,7 @@ class TestFirefoxDownload(TestCase):
         view = views.DownloadThanksView.as_view()
         view(req)
         template = render_mock.call_args[0][1]
-        assert template == ["firefox/download/desktop/thanks_direct.html"]
+        assert template == ["firefox/download/rtamo.html"]
 
     @patch.object(views, "ftl_file_is_active", lambda *x: False)
     def test_thanks_basic_direct(self, render_mock):
@@ -471,7 +472,7 @@ class TestFirefoxDownload(TestCase):
         view = views.DownloadThanksView.as_view()
         view(req)
         template = render_mock.call_args[0][1]
-        assert template == ["firefox/download/basic/thanks_direct.html"]
+        assert template == ["firefox/download/rtamo.html"]
 
     # end /thanks?s=direct URL - issue 10520
 
--- a/tests/playwright/specs/download-attribution/create.spec.js
+++ b/tests/playwright/specs/download-attribution/create.spec.js
@@ -21,7 +21,7 @@ test.beforeEach(async ({ page }) => {
     await removeDownloadAsDefault(page);
 });
 
-test.describe.skip('analytics download attribution', () => {
+test.describe('analytics download attribution', () => {
     test('has expected cookie values', async ({ page, browserName }) => {
         await page.addInitScript(mockGetGtagClientID);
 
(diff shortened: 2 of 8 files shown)


tests/playwright/specs/download-attribution/create.spec.js
maureenlholland, line 1: I'm on the fence about including this change. Can put in separate PR. But these were originally skipped when the switch was still inactive. Now it is active, the tests should pass on all envs.

successful test run: https://github.com/mozmeao/springfield/actions/runs/27617292242

springfield/firefox/tests/test_views.py
```diff
@@ -357,6 +357,7 @@ def test_no_hmac_key_set(self):
 
 
 @override_settings(DEV=False)
+@override_switch("ENABLE_ATTRIBUTION_REFACTOR", active=True)
```
stevejalim, line 360: I wondered if we need version where the switch is off, but we're probably fine as turning it off does not look likely anyway

stevejalim (approved):

PR #1514, opened 2026-06-16
Minor adjustments on What's Next page

maureenlholland:
## One-line summary

Minor adjustments on what's next page

## Issue / Bugzilla link
n/a

## Testing
Pull prod data and go to http://localhost:8000/en-US/whatsnext/

Check accessible names on "Learn more" links
Check light theme subheadings are visible

### DIFF
--- a/media/css/cms/components/flare26-section.css
+++ b/media/css/cms/components/flare26-section.css
@@ -41,6 +41,7 @@
 }
 
 .fl-heading,
+.fl-subheading,
 .fl-heading-group {
     color: var(--fl-theme-color-text);
 }
--- a/springfield/cms/templates/cms/blocks/button.html
+++ b/springfield/cms/templates/cms/blocks/button.html
@@ -7,6 +7,7 @@
 {% set block_text = block_text ~ " - " ~ value.label %}
 
 <include:button
+  id="{{ value.id }}"
   theme_class="{{ value.theme_class() }}"
   label="{{ value.label }}"
   size_class="{{ value.size_class() }}"
@@ -17,4 +18,5 @@
   analytics_text="{{ block_text }}"
   analytics_position="{{ block_position }}"
   analytics_id="{{ value.settings.analytics_id }}"
+  aria_labelledby="{{ value.aria_labelledby }}"
 />
--- a/springfield/cms/templates/cms/blocks/button.html
+++ b/springfield/cms/templates/cms/blocks/button.html
@@ -7,7 +7,6 @@
 {% set block_text = block_text ~ " - " ~ value.label %}
 
 <include:button
-  id="{{ value.id }}"
   theme_class="{{ value.theme_class() }}"
   label="{{ value.label }}"
   size_class="{{ value.size_class() }}"
@@ -18,5 +17,4 @@
   analytics_text="{{ block_text }}"
   analytics_position="{{ block_position }}"
   analytics_id="{{ value.settings.analytics_id }}"
-  aria_labelledby="{{ value.aria_labelledby }}"
 />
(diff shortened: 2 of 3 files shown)


springfield/cms/templates/cms/blocks/roadmap-list-section.html
```diff
@@ -90,8 +91,10 @@
             {% if learn_more_url %}
               {% set button_position = item_block_position ~ ".learn-more" %}
               <include:button
+                id="{{ item.learn_more_analytics_id }}"
```
maureenlholland, line 96: needs a unique id to reference for aria-labelledby

media/css/cms/components/flare26-section.css
```diff
@@ -41,6 +41,7 @@
 }
 
 .fl-heading,
+.fl-subheading,
```
maureenlholland, line 44: light theme is stuck on a constant token color instead of contextual theme color
Image

springfield/cms/templates/cms/blocks/roadmap-list-section.html
```diff
@@ -90,8 +91,10 @@
             {% if learn_more_url %}
```
maureenlholland, line 93: Too many indistinguishable "Learn more"s
Image

maureenlholland: (the "Try now" text is from CMS, might be worth adding a field to allow visually hidden text for context in those cases?)

springfield/cms/templates/cms/blocks/button.html
```diff
@@ -17,4 +18,5 @@
   analytics_text="{{ block_text }}"
   analytics_position="{{ block_position }}"
   analytics_id="{{ value.settings.analytics_id }}"
+  aria_labelledby="{{ value.aria_labelledby }}"
```
maribedran: The button block doesn't have these fields. If you want to pass those props down to the component, we can add them to the block's fields.

springfield/cms/templates/cms/blocks/roadmap-list-section.html
maribedran: What's that list? For analytics, we use the nearest heading on the `data-cta-text` to add context.

maureenlholland: This is from the WAVE extension to see link text out of context: https://wave.webaim.org/extension/

You can see the issue on any individual link from dev accessibility tools.

Whatever the analytics solution is, it would be good to provide an equivalent for interactive element accessible names

springfield/cms/templates/cms/blocks/button.html
maureenlholland: > If you want to pass those props down to the component, we can add them to the block's fields.

It looks like these fields do already exist: https://github.com/mozmeao/springfield/blob/main/springfield/cms/templates/components/button.html#L23

I can see the expected attributes rendering

maureenlholland: oh, if you mean, should we show these fields in the Wagtail dashboard for Admins to edit, I'd say no. They can be derived from content as needed.

maribedran: Right, the fields are available in the component consumed by the block, but the CMS block doesn't have those values, so it doesn't make sense to pass down values that don't exist.

maureenlholland: @maribedran I removed from the block as these aren't fields I would want to make editable

springfield/cms/templates/cms/blocks/roadmap-list-section.html
maribedran: We can add `aria-labeled-by` to the buttons passing the card heading text.

maribedran: I'll make that change

maribedran (approved):

maureenlholland:
Looks much better to me, thanks @maribedran !
Screenshot 2026-06-17 at 5 07 50 PM

PR #1515, opened 2026-06-16
Improve mobile downloads on firefox.com

slightlyoffbeat:
## Related issues

- Partially addresses #1498 — mobile users no longer route through `/thanks/` on download CTA clicks, so they don't see the "Your download was interrupted" copy in the common path. Direct visits to `/thanks/` on mobile (bookmarks, shares, hand-rolled links) still flash the page; the copy/UX revamp for that case is in [follow-up issue].
- Partially addresses #1499 — mobile CTA clicks no longer produce the `/thanks/` interstitial that triggers the "come back → reload → redirect again" loop. The direct-visit cohort still loops; covered by [follow-up issue].

## Summary

Closes the no-campaign gap from #1366: previously, mobile users clicking a `/thanks/`-bound download button on a page with no campaign declared still routed through `/thanks/` and saw "Your download was interrupted" flash for ~1s. They now go directly to the store with a default `fxcomdefault` campaign value.

`fxcomdefault` is distinct from the existing `download` value baked into `GOOGLE_PLAY_FIREFOX_LINK_UTMS` so attribution dashboards can tell "firefox.com mobile fallback" apart from explicit campaigns named `download`.

## Touched files

- `media/js/base/mobile-attribution.js` — `DEFAULT_CAMPAIGN = 'fxcomdefault'` fallback in `init()`.
- `tests/unit/spec/base/mobile-attribution/mobile-attribution.js` — new tests for the iOS and Android fallback paths.

## Before / after

| Scenario | Before | After |
|---|---|---|
| Mobile click on a `/thanks/`-bound CTA, page declares campaign | Rewritten to attributed store URL | Unchanged |
| Mobile click on a `/thanks/`-bound CTA, **no campaign** | Routes through `/thanks/` → 1s flash → bare store URL | Rewritten directly to store with `utm_campaign=fxcomdefault` (Android) / `ct=fxcomdefault` (iOS) |
| Mobile direct visit to `/thanks/` (any path) | Renders desktop page → 1s flash → bare store URL | Unchanged in this PR — addressed in [follow-up issue] |
| Desktop visit to `/thanks/` | Renders desktop page | Unchanged |

## Analytics impact — for Martech review

- **Android Play Console**: traffic that previously showed `utm_campaign=download` on no-campaign mobile CTA clicks (via the `/thanks/` auto-redirect to `GOOGLE_PLAY_FIREFOX_LINK_UTMS`) will now appear under `utm_campaign=fxcomdefault` instead. Same volume, new bucket.
- **iOS App Store Connect**: traffic that previously had no `ct=` value will now appear under `ct=fxcomdefault`.

## How to test

- [ ] **Mobile UA (devtools), page with a download CTA, no campaign**: open a page with a nav download button, click → goes directly to `play.google.com/...&utm_campaign=fxcomdefault` (Android) or `apps.apple.com/.../&ct=fxcomdefault` (iOS). No `/thanks/` flash.
- [ ] **Mobile UA, page with a download CTA, `?utm_campaign=test-19`**: campaign value flows through; URL ends with `test-19`, not `fxcomdefault`.
- [ ] **Desktop UA, any page**: download CTA still routes to `/thanks/` as before (no mobile changes apply).
- [ ] **Real Android device** (pre-launch): tap a nav download button on a no-campaign page → Play Store opens → install → `utm_campaign=fxcomdefault` appears in Play Console attribution.
```


slightlyoffbeat:
After a slack conversation with @stevejalim, I was reminded that Fastly totally ignores the Vary header.

With this in mind, I'm narrowing the scope of this PR, and the updated description reflects that.

A follow up would be updating the copy of the page that mobile users see before being redirected. We can pair that with trying to send users directly to the android/ios store directly whenever possible.

stephendherrera (approved):
LGTM and approved from a martech/analytics perspective.

stevejalim (approved):
r+ from me. leaving to @slightlyoffbeat to merge, in case there are spinning plates (eg copy changes) that should land at the same time

PR #1517, opened 2026-06-16
Adding merch store to footer

slightlyoffbeat:
## Summary

Adds a "Merch store" link to the firefox.com footer Resources section, between Learn and Support, pointing to https://shop.mozilla.com/.

- New Fluent string `footer-merch-store` in `l10n/en/footer-firefox.ftl`
- Link added to both the Flare26 footer (`flare26-footer/resources.html`) and the Protocol footer (`includes/footer/footer.html`) so it shows consistently across pages
- Outbound icon renders automatically via the existing `a[href^='http']::after` CSS in both footers — no markup or class needed
- Untranslated locales fall back to English until Pontoon syncs

## Test plan

- [ ] Load a Flare26 page (e.g. homepage) — "Merch store" appears between Learn and Support with the outbound icon, link goes to shop.mozilla.com with footer UTM params
- [ ] Load a Protocol-based page — same check
- [ ] `data-link-text="Merch Store"` and `data-link-position="footer"` present on the rendered `<a>` (for GA)

### DIFF
--- a/l10n/en/footer-firefox.ftl
+++ b/l10n/en/footer-firefox.ftl
@@ -47,6 +47,8 @@ footer-learn = Learn
 footer-support = Support
 footer-addons = Add-ons
 footer-blog = Blog
+# Link to Mozilla's merch store. Link points to https://shop.mozilla.com/
+footer-merch-store = Merch store
 
 ## Links to social media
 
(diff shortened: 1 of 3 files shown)


knowler (approved):
r+ LGTM tested locally following the steps in the PR

PR #1521, opened 2026-06-17
Preserve query string in release-notes and sysreq redirects

Muhtasim-Munif-Fahim:
## Description

When `latest_notes` or `latest_sysreq` redirect to the versioned release URL, the incoming query string is dropped. This causes the `redirect_source` parameter (used during the mozilla.org → firefox.com split) to be lost on the final hop.

Fixes #1520

## Changes

- **`springfield/releasenotes/views.py`**: Both `latest_notes` and `latest_sysreq` now append the incoming query string to the redirect target when present.
- **`springfield/releasenotes/tests/test_views.py`**: Added parameterized tests covering both views with and without query params.

## Testing

- `pytest springfield/releasenotes/tests/` passes
- The fix mirrors the [suggested approach](https://github.com/mozmeao/springfield/issues/1520#issue-2893495032) from the issue

### DIFF
--- a/springfield/releasenotes/tests/test_views.py
+++ b/springfield/releasenotes/tests/test_views.py
@@ -6,7 +6,7 @@ from unittest.mock import patch
 
 import pytest
 
-from springfield.releasenotes.views import releases_index, show_android_sys_req
+from springfield.releasenotes.views import latest_notes, latest_sysreq, releases_index, show_android_sys_req
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


janriokrause (changes requested):
Thank you for tackling this issue.

The changes in your second commit are unrelated to the original bug. Please remove to keep the history clean.

janriokrause (approved):
r+ Thank you for your first contribution 🥳

janriokrause:
Sorry, I didn't review this closely enough before merging and I'll fix it in a follow up, the parametrized redirect tests had the `@patch` mock argument in the wrong position so pytest failed collection, and it just needs `latest_release_mock` moved to the front of the signature.

PR #1528, opened 2026-06-22
chore(security): Python dep bump for cryptography; other bumps while passing by

stevejalim:
Also running integration tests because of changes to pytest and selenium deps ✅ https://github.com/mozmeao/springfield/actions/runs/27945872470


maureenlholland (approved):
r+

PR #1529, opened 2026-06-22
chore (security): update npm deps

maureenlholland:
## One-line summary

Update dev dependencies & dependencies of dev dependencies

Updated with `npm audit fix --force`. Includes major webpack dev server update. NPM overrides updated too.

The same change will apply to bedrock

## Testing
Running `npm install` and `npm start` runs with no errors

successful integration test run: https://github.com/mozmeao/springfield/actions/runs/27950607844

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "swiper": "^12.2.0",
         "terser-webpack-plugin": "^5.6.1",
         "webpack": "^5.107.2",
-        "webpack-cli": "^7.0.3"
+        "webpack-cli": "^6.0.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -11197,6 +11175,31 @@
         "node": ">= 6"
       }
     },
+    "node_modules/webpack-dev-server/node_modules/http-proxy-middleware": {
```
maureenlholland, line 11178: tested an http-proxy-middleware override, but it has to wait until webpack-dev-server updates its use of http-proxy-middleware to v3 (this is a superficial move of the code as a result of that test)

stevejalim (approved):
R+ thank you!

PR #1530, opened 2026-06-22
Remove ESR 115 download for Linux 32-bit and Windows ARM64/AArch64

knowler:
## One-line summary

ESR 115 no longer supports Linux 32-bit or Windows ARM64/AArch64 as of `115.37.0esr`, so we should not show the download link for them.

This just adds some new exclusions for when to show the “Download ESR 115” button. It builds on existing logic that would conditionally show that button when the user is trying to download ESR.

## Issue / Bugzilla link

- https://mozilla-hub.atlassian.net/browse/WT-1353
- These ESR 115 builds were disabled here: https://bugzilla.mozilla.org/show_bug.cgi?id=2040496

## Testing

These should not show a “Download ESR 115” button under the regular “Download Now” button for any language:

http://localhost:8000/en-US/download/all/desktop-esr/linux/en-US/
http://localhost:8000/en-US/download/all/desktop-esr/win64-aarch64/en-US/

The existing exclusion of Linux ARM64/AArch64 should also still apply:

http://localhost:8000/en-US/download/all/desktop-esr/linux64-aarch64/en-US/

The button should still show up for platforms that are still receiving builds, for example, Windows 64-bit:

http://localhost:8000/en-US/download/all/desktop-esr/win64/en-US/


springfield/firefox/tests/test_firefox_all.py
```diff
@@ -294,8 +295,8 @@ def mock_get_filtered_full_builds(channel, query=None):
                         "win64": {
                             "download_url": "https://download.mozilla.org/?product=firefox-esr-next-latest-ssl&os=win64&lang=en-US",
                         },
-                        "linux": {
-                            "download_url": "https://download.mozilla.org/?product=firefox-esr-next-latest-ssl&os=linux&lang=en-US",
+                        "linux64": {
```
knowler, line 297: `linux` here is 32-bit. `linux64` is probably better anyway.

springfield/firefox/tests/test_firefox_all.py
```diff
@@ -271,7 +272,7 @@ def test_firefox_esr(client, os, lang):
         assert "https://support.mozilla.org/kb/install-firefox-linux" in linux_link.attr("href")
 
 
-@pytest.mark.parametrize("os, lang", [("win64", "en-US"), ("win64", "de"), ("osx", "en-US"), ("linux", "en-US")])
+@pytest.mark.parametrize("os, lang", [("win64", "en-US"), ("win64", "de"), ("osx", "en-US"), ("linux64", "en-US")])
 def test_firefox_esr_next(client, os, lang):
     # Note: Only testing a few os/lang pairs to avoid mocking too much. We're mostly checking that 2 buttons show up.
```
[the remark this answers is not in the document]
janbrasna: There's more variants, and the explanations are usually commented next to the assertions, so instead of the suggestion above I'd go for something as:

```suggestion
def test_firefox_esr_next(client, os, lang):
    # Note: Only testing a few os/lang pairs to avoid mocking too much. We're mostly checking that all button and link types show up.
```

springfield/firefox/views.py
```diff
@@ -396,8 +396,8 @@ class MobileRelease:
                 # ESR115 builds do not exist for "sat" and "skr" languages (issue: mozilla/bedrock#15437)
                 if locale in ["sat", "skr"]:
                     download_esr_115_url = None
-                # ESR115 builds do not exist for "linux64-aarch64" platform (see issue #467)
-                if platform == "linux64-aarch64":
+                # ESR115 builds do not exist for "linux64-aarch64" (see issue #467), "linux" (32-bit), or "win64-aarch64" platforms
```
janbrasna: I've tried to shorten the comment a little to smuggle the bug reference in, maybe it's worth to have inline, something like:
```suggestion
                # ESR115 builds do not exist for "linux64-aarch64" (springfield#467); "linux" (i686) and "win64-aarch64" no longer ship (bug 2040496)
```

(feel free to tweak the language as you see fit)

springfield/firefox/tests/test_firefox_all.py
janbrasna: Yes! The next esr-next actually rolling out in a month will not have linux i686 builds available anymore anyway — so there will be more similar housekeeping to have the mocks actually reflect the current products shipped;) Good move!

maureenlholland (approved):
looks good to me, including the bug inline is a good suggestion

r+ but leaving unmerged in case you want to update comment

successful test run: https://github.com/mozmeao/springfield/actions/runs/28021887065

PR #1527, opened 2026-06-19
Ignore `play()` rejections to stop Sentry noise

bluewave41:
## One-line summary
Sentry errors are flooding in due to video playing issues. This PR fixes it.

## Testing
- [ ] Open https://www.firefox.com/en-US/whatsnew/152/
- [ ] Open inspect element tools, console. Observe the error
- [ ] Open the debugger tab
- [ ] Right click the highlighted file in the image below and add script override saving the file somewhere
- [ ] Open the saved file and replace it with the ENTIRE file from https://github.com/mozmeao/springfield/blob/1cfc5a6da88ede361c4eb63281f0a714f92deebb/media/js/cms/pages/flare26-kick-page.es6.js
- [ ] Reload the page
- [ ] Hover over the logo in the top left and keep your mouse there. The animation should still play
- [ ] Open the console tab and observe the error has disappeared and doesn't come back if you refresh

image

PR #1534, opened 2026-06-23
Ignore logo mouseover errors

bluewave41:
## One-line summary
Slow networks don't immediately load the logo SVG. Hovering over it before it's loaded still try to play it which flood Sentry with issues.

Changes copied from https://github.com/mozmeao/springfield/pull/1527/changes

## Testing
- [ ] Repeat steps from https://github.com/mozmeao/springfield/pull/1527 using the changes in this PR instead
- [ ] For the last 2 steps set your network throttling to Good 3G
- [ ] Hover in and out of the logo quickly as it's loading

### DIFF
--- a/media/js/cms/pages/flare26-kick-page.es6.js
+++ b/media/js/cms/pages/flare26-kick-page.es6.js
@@ -38,7 +38,12 @@ const setupKickPage = () => {
     logoLink.appendChild(wrapper);
 
     video.addEventListener('mouseover', () => {
-        video.play();
+        const playPromise = video.play();
+        if (playPromise !== undefined) {
+            playPromise.catch(() => {
+                // Ignore `play()` rejections (e.g. `AbortError` when paused before playback begins)
+            });
+        }
     });
     video.addEventListener('mouseout', () => {
         video.pause();
--- a/media/js/cms/pages/flare26-kick-page.es6.js
+++ b/media/js/cms/pages/flare26-kick-page.es6.js
@@ -33,36 +33,26 @@ const setupKickPage = () => {
     source.src = 'https://assets.mozilla.net/wc/logo-1-alpha.webm';
     source.type = 'video/webm';
 
-    let playPromise;
-
     video.appendChild(source);
     wrapper.appendChild(video);
     logoLink.appendChild(wrapper);
 
-    video.addEventListener('mouseover', () => {
-        playPromise = video.play();
-        if (playPromise !== undefined) {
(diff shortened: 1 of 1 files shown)


media/js/cms/pages/flare26-kick-page.es6.js
```diff
@@ -38,7 +38,12 @@ const setupKickPage = () => {
     logoLink.appendChild(wrapper);
 
     video.addEventListener('mouseover', () => {
-        video.play();
+        const playPromise = video.play();
+        if (playPromise !== undefined) {
+            playPromise.catch(() => {
+                // Ignore `play()` rejections (e.g. `AbortError` when paused before playback begins)
+            });
+        }
```
stevejalim: This looks cool - thanks @bluewave41 !

Do you think we're at risk of video.pause() also failing once the play() error is handled? Right now .play blows up and might be masking the potential for .pause() to also explode if the video hasn't loaded yet?

bluewave41: > This looks cool - thanks @bluewave41 !
> 
> Do you think we're at risk of video.pause() also failing once the play() error is handled? Right now .play blows up and might be masking the potential for .pause() to also explode if the video hasn't loaded yet?

I just thought of this too but it doesn't want to error for me locally. Let me go set up a script override again and test how the live site reacts.

media/js/cms/pages/flare26-kick-page.es6.js
knowler, line 1: Consider this non-blocking feedback, but, taking a step back, maybe we shouldn’t add the `mouseover`/`mouseout` event handlers until the video is loaded.

```js
// Do whatever you need to do to trigger loading the video

// Once the video is loaded, we progressively enhance it to play/pause on hover.
video.addEventListener('canplaythrough', () => {

  // Might still need to catch errors related to interruptions.
  video.addEventListener('mouseover',  () => { video.play(); });
  video.addEventListener('mouseout',  () => { video.pause(); });
}, { once: true });
```

bluewave41: I like this a lot better and makes more sense. I can't replicate the sentry error with this.

knowler (approved):
r+ ⚽️

I pulled this down and tested locally. I also checked quickly hover + unhovering the logo *after it loaded* to make sure we didn’t need the promise stuff we removed still during the hover effect (different then how we’re using handling it during loading).

PR #1536, opened 2026-06-23
feat(Plausible): product_download event addition

stephendherrera:
## One-line summary

Forward `product_download` events to Plausible alongside the existing GA4/dataLayer events.

- **New `Plausible.trackEvent(name, props)`** in `media/js/base/plausible/plausible.es6.js`: sends a custom event to Plausible. No-ops safely when the Plausible script isn't loaded (`window.plausible` not a function) or the visitor has opted out via GPC/DNT, reusing the existing `gpcEnabled()`/`dntEnabled()` consent helpers (same gate as `Plausible.init`).
- **New `TrackProductDownload.sendPlausibleEvent(eventObject)`** in `media/js/base/datalayer-productdownload.es6.js`: forwards every download field except the GA4 `event` name as Plausible props, fired from `sendEventFromURL` right after the existing GA4 `sendEvent` call. Wrapped in try/catch so analytics can't break the download flow.
- **Telemetry note:** this adds a new outbound analytics event:
  - Consent gating should be enforced at send time, not just load time.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1375

## Testing

- `npm run jasmine`:full front-end unit suite passes (**516 specs, 0 failures**), including new coverage for `Plausible.trackEvent` (queue-undefined, GPC, DNT, with/without props) and `TrackProductDownload.sendPlausibleEvent` / `sendEventFromURL` (fields forwarded, `event` dropped, both GA4 + Plausible fire on valid URLs, neither fires on invalid).
- `npx eslint`: clean on both changed source files.

### DIFF
--- a/media/js/base/plausible/plausible.es6.js
+++ b/media/js/base/plausible/plausible.es6.js
@@ -66,7 +66,7 @@ Plausible.trackEvent = (name, props) => {
         return;
     }
 
-    if (gpcEnabled() || dntEnabled()) {
+    if (Plausible.analyticsDenied()) {
         return;
     }
 
--- a/media/js/base/plausible/plausible.es6.js
+++ b/media/js/base/plausible/plausible.es6.js
@@ -56,4 +56,25 @@ Plausible.init = () => {
     Plausible.loadScript();
 };
 
+/**
+ * Send a custom event to Plausible. No-ops when Plausible isn't loaded or the visitor has opted out.
+ * @param {String} name - the custom event name (e.g. 'product_download')
+ * @param {Object} [props] - optional custom properties
+ */
+Plausible.trackEvent = (name, props) => {
+    if (typeof window.plausible !== 'function') {
+        return;
+    }
+
+    if (gpcEnabled() || dntEnabled()) {
+        return;
+    }
+
+    if (props) {
+        window.plausible(name, { props: props });
+    } else {
+        window.plausible(name);
+    }
+};
+
 export default Plausible;
--- a/media/js/base/datalayer-productdownload.es6.js
+++ b/media/js/base/datalayer-productdownload.es6.js
@@ -4,6 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+import Plausible from './plausible/plausible.es6';
+
 const TrackProductDownload = {};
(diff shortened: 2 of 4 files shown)


maureenlholland:
successful integration test run: https://github.com/mozmeao/springfield/actions/runs/28101920875

media/js/base/plausible/plausible.es6.js
```diff
@@ -42,4 +42,25 @@ Plausible.init = () => {
...
+ * Send a custom event to Plausible. No-ops when Plausible isn't loaded or the visitor has opted out.
+ * @param {String} name - the custom event name (e.g. 'product_download')
+ * @param {Object} [props] - optional custom properties
+ */
+Plausible.trackEvent = (name, props) => {
+    if (typeof window.plausible !== 'function') {
+        return;
+    }
+
+    if (gpcEnabled() || dntEnabled()) {
```
maureenlholland: possible we also need to check `moz-consent-pref` analytics is not denied? Unfortunately not as straightforward a check, consent cookie helpers are from '../consent/utils.es6'
```
if (hasConsentCookie()) {
        const cookie = getConsentCookie();
        if (!cookie.analytics) {
            return;
        }
}
```

maureenlholland (approved):
r+wc

tests always appreciated 🎉 

We're only checking 2 out of 3 explicit analytics denial signals (`moz-consent-pref` cookie check is missing). Is it a redundant check at this point?

No regressions in GA4 events or download behaviour testing with a direct download link (http://localhost:8000/en-US/browsers/desktop/mac/?geo=FR) and redirect to /thanks download (http://localhost:8000/en-US/thanks/?geo=FR)

I haven't seen any plausible events in the network though, just the loading of scripts
Image
Is there anything else I should be looking for here?

media/js/base/plausible/plausible.es6.js
stephendherrera: Resolved via https://github.com/mozmeao/springfield/pull/1539

maureenlholland (commented):
r++ tests 
- geo=FR (plausible loads, localhost console event)
- geo=US (no plausible data attributes)
- geo=FR with moz pref cookie for denied analytics (plausible does not load, no local console event) 

All downloads working as expected

PR #1537, opened 2026-06-24
WT-1119: rework /healthz-cdn/ to render critical pages

stevejalim:
## Summary

Replaces the previous `/healthz-cdn/` implementation (disabled in #1380 because it took down the site on first deploy) with a symptom-based check: render a small set of critical pages via an in-process `django.test.Client` and return 500 if any returns 4xx/5xx or raises. If old code can still render against a newer DB (benign migration), the check passes; if it can't (destructive schema change), it fails — without needing to model the DB state directly. Template errors, broken context processors, and DB outages all surface the same way.

## What's in here

- Re-enable the `/healthz-cdn/` route and the `SUPPORTED_NONLOCALES` entry.
- New `HEALTHZ_CDN_DB_CHECKS_ENABLED` Waffle switch that short-circuits the view to `200 pong` when off. Defaults to `settings.DEV`, so a fresh prod deploy is safe by default and operators opt in by creating/enabling the switch.
- `HEALTHZ_CDN_CRITICAL_PATHS = (homepage, /en-US/whatsnew/general/, /de/whatsnew/150/)` covering the download view, a CMS-served Wagtail page, and a locale-prefixed versioned whatsnew.
- 30s per-process cache (`SimpleDictCache`) so Fastly probe bursts collapse into one render per pod per interval. Production runs Granian with 1 worker × 1 blocking thread per pod (enforced in the webservices-infra Helm chart), so no in-pod thundering herd to design around.
- Inline-import of `_disable_apm` retained from the previous version (it's a private watchman API; guarded defensively against upstream rename).

## Why not the previous approach

The disabled version iterated `connection.introspection.get_table_description()` over every managed model on every probe — ~200 DB queries per healthcheck for Wagtail. Even at Fastly's nominal 5s interval this overwhelmed the DB. The new approach is intentionally cause-agnostic: we don't try to detect deploy-skew patterns directly (forward/reverse migration drift), because:

- It's easy to get wrong (reverse-skew detection false-positives on every rollout that includes any migration — even benign additive ones).
- It misses non-DB causes of outage.
- A "did the page render?" check is cheaper and more accurate at the same time.

## Trade-offs noted in the design

- Each healthcheck blocks the pod's single blocking thread for the duration of 3 page renders (~100–300ms healthy, several seconds if the DB is slow). The 30s cache TTL bounds how often that blocking window opens.
- The Waffle switch flip isn't instantaneous across pods (waffle caches per-process in `LocMemCache`); a `kubectl rollout restart` forces immediate propagation if needed.
- `LocMemCache` is per-pod, so check load scales linearly with pod count. At the 25–100 pod range it's small (~25–100 healthcheck-induced queries/sec fleet-wide). A separate branch will explore Redis as a shared cache to break the linear scaling.

## Test plan

- [x] `uv run pytest springfield/base/tests/test_urls.py` — 9 tests pass
- [x] `uv run pytest springfield/base/` — 339 tests pass
- [x] `uv run ruff check` clean on changed files
- [ ] Verify in dev / staging that `/healthz-cdn/` returns 200 with the switch on
- [ ] Verify that flipping `HEALTHZ_CDN_DB_CHECKS_ENABLED` off returns 200 pong without rendering
- [ ] Confirm Fastly probe behaviour against the new endpoint in dev

### DIFF
--- a/springfield/base/tests/test_urls.py
+++ b/springfield/base/tests/test_urls.py
@@ -2,8 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+from unittest.mock import patch
+
+from django.core.cache import cache
 
 import pytest
+from waffle.testutils import override_switch
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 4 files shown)


stevejalim:
Closing without merge — switching to a Fastly-side approach (real-time 5xx fallback in VCL) instead of an app-side healthcheck. See WT-1119 follow-up.

PR #1538, opened 2026-06-24
WT-1119: explore shared Redis cache for /healthz-cdn/ (on top of #1537)

stevejalim:
**Exploration / discussion PR — depends on #1537.** This branch is layered directly on top of `WT-1119--make-healthz-cdn-better`, so the diff against `main` includes the page-render work in #1537 plus the Redis layer; the [compare against the PR branch](https://github.com/mozmeao/springfield/compare/WT-1119--make-healthz-cdn-better...WT-1119--healthz-cdn-redis-cache) shows the Redis delta in isolation.

## Why

The LocMemCache approach in #1537 is per-pod, so check load scales linearly with pod count. At 45 pods that's ~45 healthcheck-induced query bursts per 30s across the fleet; at the 100-pod ceiling it's ~100. Each render holds the pod's single Granian blocking thread for the duration. A shared cache collapses the fleet to ~1 render per TTL regardless of pod count, and a single-flight mutex prevents cold-cache or simultaneous-expiry stampedes.

## What's in here (delta over #1537)

- New `healthz` named cache in `settings.CACHES`. Uses the existing `REDIS_URL` (Cloud Memorystore in prod, 1GB) on DB 1, keeping healthz keys separate from the RQ task queues on DB 0. Falls back to a per-process LocMemCache when `REDIS_URL` is unset (local dev, CI) so test behaviour is unchanged.
- `healthz_cdn` switches from the default `cache` to `caches[\"healthz\"]` and adds a `cache.add()` lock around the render. Lock holder runs the check and populates the cache; concurrent cache-missers return `200 pong` best-effort and let the winner's result land. The lock has a 15s TTL backstop in case a worker is killed mid-check.
- All cache operations are wrapped in `_safe_cache_*` helpers so a Redis hiccup never takes the healthcheck down: a failed get/add falls through to an uncoordinated check; a failed set/delete is logged but doesn't affect the response.

## Trade-offs to discuss vs the LocMemCache version in #1537

- Adds a soft dependency on Redis. With the safe wrappers it's not a hard dependency — Redis being down loses the cross-pod dedup and every pod runs its own check, but the healthcheck itself still works.
- Adds ~1–3ms of Redis round-trip per probe (one GET, plus SET and DELETE on the lock holder).
- Adds keys in DB 1 of the existing Redis instance — worth confirming the 1GB memory budget has headroom (it should: this cache holds at most one result tuple and one lock entry).
- The winner-takes-all approach can briefly delay detection of a fresh failure: if pod A wins the lock and is still rendering, pod B–Z return 200 pong best-effort even if the prior cached result was 500. Bounded by `HEALTHZ_CDN_LOCK_TIMEOUT` (15s).

## When this is worth it

My own read: marginal at 25–45 pods (the LocMemCache load is already small); a clearer win if you scale above ~100 or if connection pressure on Cloud SQL becomes a thing. Mostly here so we have a concrete shape to compare against, rather than a recommendation to land.

## Test plan

- [x] `uv run pytest springfield/base/tests/test_urls.py` — 13 tests pass
- [x] `uv run pytest springfield/base/` — 343 tests pass
- [x] `uv run ruff check` clean on changed files
- [ ] Confirm Memorystore DB 1 is reachable from the springfield namespace
- [ ] Verify behaviour in dev / staging with Redis present and Redis briefly stopped
- [ ] Compare check-rate metrics against the #1537 baseline

### DIFF
--- a/springfield/base/tests/test_urls.py
+++ b/springfield/base/tests/test_urls.py
@@ -2,8 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+from unittest.mock import patch
+
+from django.core.cache import cache
 
 import pytest
+from waffle.testutils import override_switch
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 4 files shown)


stevejalim:
Closing without merge — switching to a Fastly-side approach (real-time 5xx fallback in VCL) instead of an app-side healthcheck. See WT-1119 follow-up.

PR #1539, opened 2026-06-24
Wt 1375 plausible consent gating

stephendherrera:
## One-line summary

Skip loading Plausible analytics when a visitor has opted out.

- Adds a `Plausible.analyticsDenied()` helper that centralizes the opt-out checks (GPC, DNT, and an explicit analytics opt-out in the consent cookie).
- `Plausible.init()` now uses that helper instead of the inline GPC/DNT check, so the script is not loaded at all when a visitor has opted out.
- **Point to review:** a missing consent cookie is intentionally *not* treated as an opt-out, so the default loading behavior is unchanged; only an explicit opt-out suppresses loading.
- **Point to review:** gating happens at init/script-load time. An opt-out made *after* the script has already loaded in the same session would need a per-event check to be covered (noted as a follow-up).

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1388

## Testing

- `npm run jasmine` passes (465 specs, 0 failures), stable across multiple random seeds.
- Added unit coverage for `analyticsDenied()` (GPC, DNT, cookie opt-out, cookie opt-in, and no-cookie cases) and for `init()` loading / not loading the script accordingly.
- `npx eslint` clean.

### DIFF
--- a/media/js/base/plausible/plausible.es6.js
+++ b/media/js/base/plausible/plausible.es6.js
@@ -10,9 +10,7 @@ const Plausible = {};
 
 /**
  * Determines if the visitor has opted out of analytics, via GPC, DNT, or by
- * explicitly declining analytics in the consent banner. A missing cookie is
- * NOT a denial: Plausible runs cookieless by default (e.g. for EU visitors who
- * haven't made a choice yet), so only an explicit `analytics: false` opts out.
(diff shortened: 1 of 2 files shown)


maureenlholland (approved):
r+

PR #1541, opened 2026-06-25
WT-1119: remove /healthz-cdn/ from Springfield

stevejalim:
## Summary

Removes the disabled `/healthz-cdn/` endpoint and all related scaffolding.

We're moving the failover-on-bad-origin logic to Fastly's VCL (real-time 5xx response handling
plus stale-on-error) rather than trying to express it as an app-side healthcheck — so the surface
area can just go away. The simple `/healthz/` ping (watchman) and the `/healthz-cron/` status
dashboard are unchanged.

## What's removed

- `healthz_cdn` view and its `logger` in `springfield/base/views.py`
- Now-unused imports: `django.apps.apps`, `django.db.connections`,
  `InconsistentMigrationHistory`, `MigrationExecutor`, `HttpResponse`, `logging`
- The commented-out URL pattern in `springfield/urls.py`
- The commented-out `SUPPORTED_NONLOCALES` entry in `springfield/settings/base.py`
- The four commented-out tests in `springfield/base/tests/test_urls.py`

## Context

Two earlier exploratory PRs (#1537 page-render approach, #1538 Redis-cache variant) have been
closed without merging. The takeaway from those iterations was that an active healthcheck either
loads the DB too much or amplifies slow-page conditions, and the cleaner fix is on the Fastly side
— observed-5xx triggered fallback, with `stale_if_error` for cacheable pages. That work will land
separately in `webservices-infra/springfield`, starting with Dev only.

## Test plan

- [x] `uv run pytest springfield/base/` — 333 tests pass
- [x] `uv run ruff check` clean
- [x] No remaining references to `healthz_cdn`, `healthz-cdn`, or `HEALTHZ_CDN` in the codebase

### DIFF
--- a/springfield/base/tests/test_urls.py
+++ b/springfield/base/tests/test_urls.py
@@ -36,57 +36,6 @@ def test_readiness(client):
     )
 
 
-# def test_healthz_cdn(client):
-#     _test(
-#         url="/healthz-cdn/",
-#         client=client,
-#         expected_content="pong",
-#     )
-
-
-# def test_healthz_cdn_fails_when_migrations_pending(client):
-#     from django.db.migrations.executor import MigrationExecutor
-
-#     class FakeMigration:
-#         app_label = "fake_app"
-#         name = "0001_fake"
-
-#     with patch.object(MigrationExecutor, "migration_plan", return_value=[(FakeMigration(), False)]):
-#         _test(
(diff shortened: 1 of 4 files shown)


janriokrause (approved):
r+ and +1 for the Copilot comment

PR #1544, opened 2026-06-25
Ignore play sentry error

bluewave41:
## One-line summary
Filter out play errors

https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-ignore-errors


knowler (approved):
r+

PR #1546, opened 2026-06-25
Fix sitemap cataloguing deprecated paths (WT-1169)

janriokrause:
# Summary

The sitemap generator runs without `ENABLE_CMS_REFRESH_REDIRECTS=True`, so it catalogued the old deprecated paths that 404 in production.

# Details

* Make the CMS refresh download URLs unconditional and remove the legacy else branch
* Point the `unsupported-systems` `noindex` rule at its new path

The `/mobile/get-app/` to `/mobile/` redirect and its test already exist in the unconditional `redirectpatterns`, so removing the page route keeps it working.

# Heads up

💡 Hint for code reviewers: use the "Hide whitespace" option in the "diff view settings" (gear icon) on the "Files changed" tab to make review easier.

# Follow up

To be discussed:

1. Leftover asymmetry: `urls.py` is now unconditional, but `redirects.py:135` still gates the `browsers/…` → `download/…` redirects behind `if settings.ENABLE_CMS_REFRESH_REDIRECTS`. So with the flag off (local dev default), legacy `/browsers/…` URLs now 404 instead of redirecting. Irrelevant for the sitemap script, but a comment in `urls.py` pointing at the `redirects.py` counterpart would keep the follow-up from missing it.
2. Stale test mechanism: `test_cms_refresh_redirects.py` still toggles `ENABLE_CMS_REFRESH_REDIRECTS=True` and reloads the urlconf. That reload is now a no-op for URL patterns (tests still pass).
3. Orphaned template `firefox/browsers/mobile/get-app.html` (+ its FTL) is now route-less: harmless, a cleanup candidate.

# Context

* https://mozilla-hub.atlassian.net/browse/WT-1169
* Closes #1427


stevejalim:
Thanks for this @janriokrause!

Following up

1. I'd go with Copilot's suggestions to provide FTL files to the prefer_cms decorator. Happy to chat directly if you have any questions about those canges

2. Following up on the asymmetry callout, let's drop the `ENABLE_CMS_REFRESH_REDIRECTS` flag entirely, since `True` is the only behavior we want to keep going forward.

Specifically, that loks like

**`springfield/firefox/redirects.py`**
- Drop the `if settings.ENABLE_CMS_REFRESH_REDIRECTS:` conditional and the surrounding wiring on 121-122.
- Fold the 10 `redirect(...)` calls into `redirectpatterns` with `permanent=settings.PERMANENT_CMS_REFRESH_REDIRECTS` inlined. Keep them as a named `refresh_redirects` tuple alongside `redirectpatterns` so the existing `test_refresh_redirect_destinations` parametrization in `test_redirects.py` keeps working without rewrite.

**`springfield/settings/base.py`**
- Remove `ENABLE_CMS_REFRESH_REDIRECTS` as a const. `PERMANENT_CMS_REFRESH_REDIRECTS` can stay for now, but let's ticket that to zap straight after - `True` is the only mode we want for that setting now,

**`springfield/cms/tests/test_cms_refresh_redirects.py`**
- The `cms_refresh_redirects_enabled` fixture (lines 28-39) becomes redundant, so let's delete it, adjust the file's docstring and rename the file (eg `test_download_url_cms_preference.py`?) 

**`springfield/firefox/tests/test_redirects.py`**
- `test_refresh_redirects_not_in_redirectpatterns_when_disabled` can go in the woodchipper

**`tests/redirects/test_redirects.py:31`**
- Drop the `override_settings(ENABLE_CMS_REFRESH_REDIRECTS=True)` wrapper.

The orphaned `mobile/get-app.html` template + FTL from the original follow-up list can ride along in the same PR or a separate cleanup, your call.

janriokrause:
I applied Copilot's suggestions and removed `mobile/get-app` in this PR. The removal of `ENABLE_CMS_REFRESH_REDIRECTS`, based on your comment, is in #1593.

springfield/firefox/urls.py
```diff
@@ -149,83 +148,54 @@
...
-            prefer_cms(
-                L10nTemplateView.as_view(template_name="firefox/browsers/mobile/focus.html", ftl_files=["firefox/browsers/mobile/focus"]),
-                fallback_ftl_files=["firefox/browsers/mobile/focus"],
-            ),
-            name="firefox.browsers.mobile.focus",
+        name="firefox.browsers.desktop.chromebook",
+    ),
+    path("download/linux/", prefer_cms(views.PlatformViewLinux.as_view()), name="firefox.browsers.desktop.linux"),
+    path("download/mac/", prefer_cms(views.PlatformViewMac.as_view()), name="firefox.browsers.desktop.mac"),
+    path("download/windows/", prefer_cms(views.PlatformViewWindows.as_view()), name="firefox.browsers.desktop.windows"),
```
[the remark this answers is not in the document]
janriokrause: Asking myself why Copilot didn't mention that in the first place?

tests/redirects/test_redirects.py
```diff
@@ -28,15 +27,14 @@
...
-    with override_settings(ENABLE_CMS_REFRESH_REDIRECTS=True):
-        importlib.reload(firefox_redirects_module)
-        importlib.reload(firefox_urls_module)
-        importlib.reload(root_urls_module)
-        redirects_util.redirectpatterns.clear()
-        redirects_util.register(firefox_redirects_module.redirectpatterns)
-        clear_url_caches()
-        url["base_url"] = base_url
-        assert_valid_url(**url)
+    importlib.reload(firefox_redirects_module)
```
stevejalim, line 30: Now that we're no longer trying two states for the redirects (on and off) we don't need to reload these URL modules, I think

springfield/settings/base.py
```diff
@@ -56,12 +56,9 @@ def data_path(*args):
...
-# CMS refresh redirect controls (toggled by infra during URL/content migrations).
-# ENABLE_CMS_REFRESH_REDIRECTS turns the redirects on; defaults to false so new behavior
-# is opt-in and can be rolled out gradually. PERMANENT_CMS_REFRESH_REDIRECTS switches
-# the redirects from temporary (302) to permanent (301); defaults to false so we can
-# verify the rollout and later make redirects permanent or remove them as part of cleanup.
-ENABLE_CMS_REFRESH_REDIRECTS = config("ENABLE_CMS_REFRESH_REDIRECTS", default="false", parser=bool)
+# PERMANENT_CMS_REFRESH_REDIRECTS switches the CMS-refresh redirects from temporary (302)
+# to permanent (301); defaults to true. The redirects themselves are now always
+# on. TODO: remove this setting once the redirects are confirmed permanent (follow-up).
 PERMANENT_CMS_REFRESH_REDIRECTS = config("PERMANENT_CMS_REFRESH_REDIRECTS", default="true", parser=bool)
```
stevejalim, line 62: We should do a follow-up task to remove `PERMANENT_CMS_REFRESH_REDIRECTS` too now and just have them always set to permanent

stevejalim (approved):

stevejalim:
Let's wait for integration tests to pass before we merget this to be extra sure.

PR #1551, opened 2026-06-29
Get hidden fields' POST value in ContactPage falling back to default

lucianoratamero:
## One-line summary

This PR fixes the way we're getting hidden fields' data in the contact page. Now, it tries to get data from POST, before falling back to the default value.

## Testing

Add a debugger/logger to the contact page _collect_form_data method, open a contact page passing a `ls=test` query param, and validate that the `lead_source` hidden field is now populated with `test` instead of the default value.


kkellydesign (approved):
works for me 👍

PR #1552, opened 2026-06-29
Escape CMS values in Firefox Account button markup (WT-1397)

janriokrause:
# Summary

`_fxa_product_button` assembled the Firefox Account button anchor by string concatenation and returned it as `Markup`, emitting the CMS button label, the optional attribute values and the label-derived `href` without escaping. Any CMS editor who can edit and publish a page with a Firefox Account Button could store active HTML in the label and trigger stored XSS for every visitor of the public page.

The fix suggested in the report only escaped the button label, which closes the anchor body but leaves two more injection points open: the optional attribute values (reachable through the rich-text `<fxa>` path in `cms_tags.richtext`) and the `href`, into which the label flows via `utm_campaign` and `entrypoint`. Escaping therefore has to happen at the sink so all three are covered at once.

# Fix

Build the anchor with `format_html` and `format_html_join` so all interpolated values are escaped. `inner_html` remains the single trusted slot via `mark_safe`, which is safe because the component template escapes the label before it reaches that slot.

Escaping the `href` renders its `&` separators as `&amp;`, which is valid HTML and decodes identically, so the existing exact-string button tests are updated to match.

# Tests

New `TestFxAButtonXSS` class:

* `test_button_text_label_is_escaped` confirms a label payload renders as inert text in the anchor body.
* `test_optional_attribute_value_is_escaped` confirms an attribute value cannot break out of its attribute.
* `test_label_in_href_entrypoint_is_escaped` confirms a label reaching the `href` cannot break out of the URL.
* `test_trusted_inner_html_is_preserved` confirms the trusted `inner_html` slot is still rendered as-is.

The three escaping tests were confirmed to fail against the unpatched sink, and the `inner_html` test fails if its `mark_safe` is removed.

# Context

* https://mozilla-hub.atlassian.net/browse/WT-1397
* https://bugzilla.mozilla.org/show_bug.cgi?id=2044038

### DIFF
--- a/springfield/firefox/templatetags/misc.py
+++ b/springfield/firefox/templatetags/misc.py
@@ -13,6 +13,8 @@ from django.contrib.staticfiles.finders import find as find_static
 from django.template.defaultfilters import slugify as django_slugify
 from django.template.defaulttags import CsrfTokenNode
 from django.utils.encoding import smart_str
+from django.utils.html import format_html, format_html_join
+from django.utils.safestring import mark_safe
 
 import jinja2
 from django_jinja import library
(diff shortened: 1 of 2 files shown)


maribedran (approved):

janriokrause:
Bedrock port: mozilla/bedrock#17273

PR #1553, opened 2026-06-29
Update prettier to match pre-commit CI job

maribedran:
## One-line summary

Pre-commit and local prettier versions drifted, so the pre-commit and the test workflows disagreed on the right format for the same file.


maribedran:
https://github.com/mozmeao/springfield/actions/runs/28407675280/job/84173683674?pr=1516

lucianoratamero (approved):
tested locally, worked just fine!

PR #1555, opened 2026-06-30
(hotfix): Add clean-up migration to drop stale ContentTypes 

stevejalim:
…following the deletion of models

`9b63e06a` deleted some 2025-era page models, but didn't also delete the related ContentType which then caused the recently-edited-pages tab to blow up the CMS dashboard.

This changeset fixes that by deleting the now-stale content types


stevejalim (commented):
Self review: r+

PR #1558, opened 2026-06-30
Refactor Contact Page to properly handle hidden inputs from query param

maribedran:
## One-line summary

Simplify how the Contact Page uses query parameters to set the value on hidden fields.

- Add a new `query_param_override` option to the `HiddenFieldBlock` to let its default value be set from query parameters
- Remove the JS and request session logic used to pass the query parameters to the field's value
- Prevent fields that aren't user-facing from being translated in field blocks

## Testing

- Access http://localhost:8000/en-US/flare-docs/pages/test-contact-page/?ls=something
- Inspect the page and check that the form has the hidden field `<input type="hidden" name="lead_source" value="something">` with the value passed on the `?ls=` query param
- Remove the `required` attribute from any of the form fields so that the browser allows you to submit an invalid post
- Submit the form and check that the value of the hidden field is still the one passed to the query param
image

### DIFF
--- a/media/js/cms/contact-form.js
+++ b/media/js/cms/contact-form.js
@@ -23,20 +23,8 @@
         }
     }
 
-    function setCorrectLeadSource() {
-        var leadSourceQueryParam = new URLSearchParams(
-            window.location.search
-        ).get('ls');
-        var leadSourceInput = document.querySelector('input[name=lead_source]');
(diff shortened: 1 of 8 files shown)


lucianoratamero (approved):
copilot thingys and you're good!

kkellydesign (approved):
works well 👍

PR #1567, opened 2026-07-01
WT-1119: add SyntheticServerErrorMiddleware to force 5xx for cascade testing

stevejalim:
## Summary

This changeset adds a token-gated Django middleware that returns HTTP 500 when a request carries a matching
`X-Springfield-Cascade-Test` header. The purpose is to force user-facing 5xx from Springfield in a
controlled way, so we can test Fastly's [origin failover cascade](https://github.com/mozilla/webservices-infra/pull/11575) without breaking anything real.

## Why we need this

If we scale down the cluster to 0 to simulate GKE dying, then Fastly's redirector logic sends traffic to the failover static site, but that doesn't use the cascade/fallback logic that we want to add. 

The "cascade" behaviour (details below) is triggered by 500s in pages even when `/healthz/` (which is pretty naive) confirm the application is technically available.

In Dev it doesn't really make a difference, but in Stage and Prod we want to be able to run a failover cascade from the US cluster to the EU cluster and _then_ to the static site in the worst case. This new middleware lets us test that in by simulating 500s from pages which are not the `/healthz/` or `/readiness/` pages.

## What it does

- **When `SYNTHETIC_5XX_TOKEN` env var is unset**: middleware raises `MiddlewareNotUsed` at construction so Django drops it from the chain entirely. Zero cost, zero risk. This is the default in every environment.
- **When `SYNTHETIC_5XX_TOKEN` is set** (via a k8s Secret, only for environments where cascade testing is desired): a request carrying the same token value in the `X-Mozilla-Ops-Canary` header returns HTTP 500 with a synthetic body. Other requests pass through.
- **Healthcheck paths (`/healthz/`, `/readiness/`, `/healthz-cron/`) are always passed through** even with the matching token. This is deliberate and load-bearing: we need to force user-facing 5xx WHILE keeping `/healthz/` green, so that Fastly's probe stays healthy and only the restart-based cascade can rescue the request. Otherwise the probe would trip and director-fallback would kick in, obscuring what we're trying to test.
- **Token comparison uses `hmac.compare_digest`** for constant-time comparison, so nothing about the token can be inferred from response timing.

## Test plan

- [x] New unit tests in `test_middleware.py` covering: no-op when unset, fires on matching header,
      passthrough without header, passthrough with wrong header, skips each of the three healthcheck
      paths, constant-time compare prefix-resistance
- [x] `uv run pytest springfield/base/tests/test_middleware.py` (62 tests pass)
- [x] `ruff check` clean on changed files
- [ ] After Secret is set in Dev, run the actual cascade test via curl and confirm `restarts=1, backend=F_springfield_backup, response_status=200` in BQ

### DIFF
--- a/springfield/base/middleware.py
+++ b/springfield/base/middleware.py
@@ -11,6 +11,7 @@ the locale codes.
 
 import base64
 import contextlib
+import hmac
 import inspect
 import logging
 import time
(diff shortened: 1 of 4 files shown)


stevejalim:
Test driven by @diox as well - thanks!

PR #1570, opened 2026-07-01
[WT-1407] Regenerate analytics ids on page copy

maribedran:
## One-line summary

When pages are copied in the admin, regenerate all analytics ID values for the new page.

- Create the `regenerate_analytics_ids_on_copy` hook for `after_copy_page` that updates the values of all UUID blocks on the copied page, unless it's opted out
- Override Wagtail's page copy form with a checkbox to opt out of UUID regeneration

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1407

## Testing

- Choose any existing page and copy it through the Wagtail admin (Edit page -> ... button -> Copy)
  - Verify that buttons in the copied page have different analytics IDs
- Copy the page and check the "Keep analytics IDs" checkbox on the page copy form
  - Verify that the copied page has the same analytics ids as the original page
- Copy the page and check the "Alias" checkbox on the page copy form
  - Verify that the copied page has the same analytics ids as the original page
- Translate the page
  - Verify that the translated page has the same analytics ids as the original page


stevejalim:
@maribedran 

> @maribedran requested a review from https://github.com/apps/copilot-pull-request-reviewer [9 hours ago](https://github.com/mozmeao/springfield/pull/1570#event-27452322978)

FWIW i think this comes from _your_ copilot credits/allowance, not Mozilla's, so feel free to ping me/someone to ask for one of the MoCo devs to trigger Copilot if you want

springfield/admin/templates/wagtailadmin/pages/copy.html
```diff
@@ -0,0 +1,50 @@
+{% extends "wagtailadmin/pages/copy.html" %}
+
+{% comment %} SKIP LICENSE INSERTION {% endcomment %}
```
stevejalim, line 3: Looks like the comment failed to tell the pre-commit hook what to do :/  Let's see if we can remove the MPL block below

springfield/admin/templates/wagtailadmin/pages/copy.html
```diff
@@ -0,0 +1,50 @@
...
+
+        {% if form.keep_analytics_ids %}
+          <li>{% formattedfield form.keep_analytics_ids %}</li>
+        {% endif %}
+      </ul>
+
+      <input type="submit" value="{% trans 'Copy this page' %}" class="button">
+    </form>
+  </div>
+{% endblock %}
```
stevejalim, line 50: Shall we add a custom Copilot rule to check this file against new Wagtail source every time we bump the version of Wagtail?

springfield/cms/wagtail_hooks.py
```diff
@@ -577,3 +580,45 @@ class PencilBannerSnippetViewSet(LocaleDefaultingSnippetViewSet):
...
+
+    Applies to the copied page and every copied descendant, since recursive copies
+    build subpages without calling ``Page.copy()`` either.
+    """
+    post = getattr(request, "POST", {})
+    user = getattr(request, "user", None)
+
+    if post.get("keep_analytics_ids"):
+        return
+    if new_page.alias_of_id:
```
stevejalim, line 604: Does this line identify a translation, or one of our alias locales?

springfield/admin/templates/wagtailadmin/pages/copy.html
stevejalim: Might be good to warn that this markup might break in future Wagtail versions, and would need updating

stevejalim (approved):
r+wc - makes sense to me, but I think we should get eyes from @stephendherrera too re the analytics field

springfield/admin/templates/wagtailadmin/pages/copy.html
maribedran: That's a great idea. I'd add a general rule to check all our overrides, since we have a custom page model and tweaks to Wagtail behavior all over the project.

springfield/cms/wagtail_hooks.py
maribedran: This identifies that the page is an alias of another page. When you copy a page, you have the option to make the copy an alias of the original, so all the changes to the original can be synced to the copy.

springfield/admin/templates/wagtailadmin/pages/copy.html
maribedran: The pre-commit check passed here. Where did you see the failure?

springfield/admin/templates/wagtailadmin/pages/copy.html
maribedran: @stevejalim https://github.com/mozmeao/springfield/pull/1570/commits/367f6ed45e9361ebffecbb1412f9a0148f778f9f

stephendherrera:
Logic for this PR looks good to me.

stephendherrera (approved):

PR #1571, opened 2026-07-02
WT-1119: drop Cache-Control override from synthetic 500 response

stevejalim:
Setting `Cache-Control: no-store, private` (and `Surrogate-Control: no-store`) on the
synthetic 500 caused Fastly to switch the response into pass state before the failover
cascade in `vcl_fetch` could fire. BQ evidence: `restarts=0, response_state=PASS` on test
requests.

Drops both headers. Fastly's default is to not cache 5xx anyway, so the belt-and-braces was
low-value; the cascade firing correctly is not.

Test and docstring updated accordingly.


stevejalim (commented):
self review r+

PR #1573, opened 2026-07-02
Page fixtures improvements

maribedran:
## One-line summary

Make sure that all page fixtures have no duplicate block IDs and can be translated.

- Create the `with_fresh_ids` helper to make sure that page fixtures don't use duplicate IDs (most fixtures use the same blocks on the upper and lower content sections of the page)
- Create the `get_or_create_page` helper to make sure that page fixtures always use the en-US locale to avoid picking a translation instead of the original page when fixtures are loaded 
- Add a new test to guarantee that all page fixtures can be translated

## Issue / Bugzilla link

The issue with duplicate IDs came up during https://github.com/mozmeao/springfield/pull/1570 development

## Testing

- Access http://localhost:8000/en-US/flare-docs and check that the page fixtures look the same
- Pick any fixture page and translate it
- Delete the original page and load fixtures again
- Make sure that the original page is recreated


lucianoratamero (approved):
followed the test steps and all worked as intended 🎉

PR #1574, opened 2026-07-02
Fix BreadcrumbList to exclude non-routable ancestors

slightlyoffbeat:
## What changed

Fixes the `BreadcrumbList` JSON-LD emitted on CMS pages. Previously position 1 was `{"name": "Welcome to your new Wagtail site!", "item": null}` — a Wagtail default title leaking on every CMS page (invalid schema per Google's spec, since only the last position can omit `item`).

Root cause: template filtered ancestors by `depth__gt=1`, keeping the depth-2 non-routable per-locale root pages that sit above `Site.root_page` in production (see migration 0060).

- New `AbstractSpringfieldCMSPage.get_breadcrumb_ancestors()` — filters by `url_path` prefix against the cached `Site.get_site_root_paths()`, so only ancestors inside the routable subtree are returned.
- Also chains `.public()` to keep view-restricted ancestor titles/URLs out of public JSON-LD.
- Template updated to call the helper.
- New `prod_shape_site` fixture that mirrors production's 3-level tree, plus 6 new tests covering the helper (including one that would catch a regression to `depth__gt=1`, and one that would catch dropping `.public()`).

## Things to note

- **Cost:** +1 DB query per CMS page render (`.public()` does a `PageViewRestriction` lookup). Bumped `test_blog_index_no_n_plus_one_queries` (17→18) and `test_blog_all_no_n_plus_one_queries` (20→21) with an explanatory comment.
- **Homepage behavior change:** rendering `Site.root_page` itself now emits no `BreadcrumbList` at all (its own url_path is not a prefix of any ancestor's). Before, it emitted a broken 2-item list with `"Welcome..."` at position 1. Confirm SEO is fine with a home page having no BreadcrumbList — I think it's the right call.
- **Data-side cleanup (not in this PR):** the depth-2 per-locale root pages still have the default `"Welcome to your new Wagtail site!"` title in Wagtail. Not user-visible after this fix, but I'll rename them manually in prod for admin readability.
- **Preview mode caveat:** `.live()` filters draft ancestors, so previews under an unpublished parent show a truncated breadcrumb. Considered out of scope — preview honestly reflecting "what would ship if only this page were published" is arguably correct.

## How to test

```bash
pytest springfield/cms/tests/test_structured_data.py springfield/cms/tests/test_blog_pages.py -v

### DIFF
--- a/springfield/cms/models/base.py
+++ b/springfield/cms/models/base.py
@@ -10,7 +10,7 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.cache import never_cache
 
 from wagtail.admin.panels import FieldPanel
-from wagtail.models import Locale, Page as WagtailBasePage
+from wagtail.models import Locale, Page as WagtailBasePage, Site
 from wagtail_localize.fields import SynchronizedField
 
 from lib import l10n_utils
@@ -205,3 +205,26 @@ class AbstractSpringfieldCMSPage(WagtailBasePage):
     def noindex(self):
         """By default, don't add the robots meta tag to CMS pages, but allow child classes to override this if needed."""
         return False
+
+    def get_breadcrumb_ancestors(self):
+        """Live, publicly-visible ancestors of this page for a BreadcrumbList.
+
+        Restricts the ancestor chain to pages within the current Site's
+        routable subtree by matching `url_path` prefixes against the cached
+        `Site.get_site_root_paths()`. Pages above the routable subtree — the
+        Wagtail system root (depth 1) and per-locale root pages (depth 2 in
+        production, see migration 0060_create_alias_locale_records) — have
+        `url_path`s outside every SiteRootPath prefix and are excluded.
+
+        Applies `.public()` so ancestors with a `PageViewRestriction`
+        (password-required, group-restricted, login-required) never leak
+        into public JSON-LD.
+
+        Filters on `url_path` rather than calling `get_site()` / iterating
+        `full_url` on each ancestor — both would fire extra queries per
+        page render. Don't refactor to `full_url is not None` in Python.
+        """
+        for site_root_path in Site.get_site_root_paths():
+            if self.url_path.startswith(site_root_path.root_path):
+                return self.get_ancestors().live().public().filter(url_path__startswith=site_root_path.root_path)
+        return WagtailBasePage.objects.none()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: springfield/cms/models/base.py [hunk 1]
```diff
@@ -205,3 +205,26 @@ def og_description(self):
...
+        (password-required, group-restricted, login-required) never leak
+        into public JSON-LD.
+
+        Filters on `url_path` rather than calling `get_site()` / iterating
+        `full_url` on each ancestor — both would fire extra queries per
+        page render. Don't refactor to `full_url is not None` in Python.
+        """
+        for site_root_path in Site.get_site_root_paths():
+            if self.url_path.startswith(site_root_path.root_path):
+                return self.get_ancestors().live().public().filter(url_path__startswith=site_root_path.root_path)
```
janriokrause:
```suggestion
                return (
                    self.get_ancestors()
                    .live()
                    .public()
                    .filter(url_path__startswith=site_root_path.root_path)
                    .order_by("path")
                )
```

`BreadcrumbList` treats `position` as authoritative, and the template derives it from `loop.index`. Correct ordering today is a happy accident of Wagtail's default `Page.Meta.ordering = ["path"]` — nothing in `get_breadcrumb_ancestors()` states that dependency. A future subclass override or a Wagtail default change could silently reorder ancestors, producing syntactically valid but semantically wrong JSON-LD that gets ingested by Google on every affected CMS page. `.order_by("path")` is free (the DB already sorts that way) and locks the contract in at the call site.

springfield/cms/models/base.py
```diff
@@ -205,3 +205,26 @@ def og_description(self):
...
+        production, see migration 0060_create_alias_locale_records) — have
+        `url_path`s outside every SiteRootPath prefix and are excluded.
+
+        Applies `.public()` so ancestors with a `PageViewRestriction`
+        (password-required, group-restricted, login-required) never leak
+        into public JSON-LD.
+
+        Filters on `url_path` rather than calling `get_site()` / iterating
+        `full_url` on each ancestor — both would fire extra queries per
+        page render. Don't refactor to `full_url is not None` in Python.
```
janriokrause, line 225: The `"Don't refactor to full_url is not None in Python"` warning in the docstring is doing important work, but it's currently guarded only by prose. Consider adding a query-count test that varies ancestor depth to turn the invariant into a tripwire.

Maybe something like this as a test:

```python
article = prod_shape_site["article"]
with django_assert_max_num_queries(2):
    list(article.specific.get_breadcrumb_ancestors())
```

stevejalim (approved):

PR #1575, opened 2026-07-02
Fix reference to renamed flare 26 files

maribedran:
https://mozilla.sentry.io/issues/7590081760/?query=is%3Aunresolved&referrer=issue-stream


janriokrause (approved):
r+

PR #1576, opened 2026-07-02
Contact page improvements

maribedran:
## One-line summary

Instantiate a Django Form from the Contact Page field blocks and use it to validate the submission.

- Bind each field block to a corresponding Django form field with the appropriate attributes
- Instantiate a Django form with the page's form fields and use it to validate the submission
- Fix hidden fields handling, invalidating the form if those are not sent
- Fix the logic to consider an empty form submission invalid
- Organize the `serve` method for better readability
- Improve the email message formatting
- Organize tests
- Only allow the Contact Page to use a limited set of paths on production

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1354

## Testing

- To test locally, you need to run [Basket](https://github.com/mozmeao/basket) alongside Springfield
  - Replace "8000:8000" with "8100:8000" in docker-compose.yml, so the ports don't conflict with Springfield ports
  - Run `make build` and `make run`
- Run `./manage.py load_page_fixtures` and access http://localhost:8000/en-US/flare-docs/pages/test-contact-page/
  - Submit the form and verify that the page shows the success message
  - Edit the form's HTML, removing the `required` attribute from any field so the browser allows an invalid post, and check that the error message is displayed
- Edit the page and set it to send an email instead of the basket API in the "Settings" panel
  - Submit the form and check the server logs for the email message
- Edit the page and set a "Redirect to" in the "Settings" panel
  - Submit the form and check that you get redirected to the page


springfield/cms/models/pages.py
```diff
@@ -1945,178 +1948,168 @@ def clean(self):
...
-            identifier = value["internal_identifier"]
-            if field.block_type == "hidden_field":
-                post_value = request.POST.get(identifier, "")
-                data[identifier] = post_value if post_value else value.get("default_value", "")
-            elif field.block_type == "checkbox_group_field":
-                data[identifier] = ", ".join(request.POST.getlist(identifier))
-            else:
-                data[identifier] = request.POST.get(identifier, "")
-        return data
+            form_fields[value["internal_identifier"]] = value.get_form_field()
```
stevejalim, line 2087: Any risk that we'd get no `internal_identifier` key in the Field that returned?

springfield/cms/models/pages.py
```diff
@@ -1945,178 +1948,168 @@ def clean(self):
...
 
-    def get_form_data_for_context(self, post_data):
-        """Return submitted form values keyed by internal_identifier for template use.
+        ContactForm = type("ContactForm", (forms.Form,), form_fields)
```
stevejalim, line 2089: This works, but I wonder if something more idomatic with a concrete Form base class would be easier to scan? Something like

```
class DynamicContacForm(forms.Form):

   def __init__(self, form_fields, *args, **kwargs):
        # form creation here


```

Am easy either way, but what do you think?

stevejalim (approved):
Neat. r+wc!

springfield/cms/models/pages.py
maribedran: No, the field is required.

springfield/cms/models/pages.py
maribedran: I thought of that too. I wasn't sure which one looked better. Let me try it.

maribedran: The class does look better:
```
        class ContactForm(forms.Form):
            def __init__(_self, *args, **kwargs):
                super().__init__(*args, **kwargs)
                _self.fields.update(form_fields)

            def clean(_self):
                # The honeypot must stay empty, and every hidden field must have a value
                honeypot = _self.data.get("office_fax")
                empty_hidden_fields = any(not _self.data.get(identifier) for identifier in hidden_identifiers)
                if honeypot or empty_hidden_fields:
                    raise forms.ValidationError(ftl_lazy("contact-form-error-sending", ftl_files=self.ftl_files))
                # Only flag an empty submission when no per-field error already exists
                has_any_data = any(_self.cleaned_data.get(identifier) for identifier in visible_identifiers)
                if not has_any_data and not _self.errors:
                    raise forms.ValidationError(ftl_lazy("contact-form-error-empty", ftl_files=self.ftl_files))
                return _self.cleaned_data
  ```

PR #1589, opened 2026-07-07
W-T1415 Allow searching for page content and block types 

maribedran:
## One-line summary

Index all pages' content and block types, and add a new admin view to search it.

- Add search indexes to all page models with their content fields
- Add block types to `StreamField`'s searchable content so blocks' content and type keys are both indexed
- Create a new "Content Search" admin page to allow searching for pages' content*

* The default Wagtail search only includes fields added through the `AutocompleteField` index. The option for the new page was to keep the existing one simple and fast. Adding all pages' content to that index would potentially slow it down when the content volume grows.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1415

## Testing

Go to `/cms-admin/content-search/` and search for any block type slug. Ex: "illustration_card"

## Deploy

Once this is deployed, we need to run the `./mange.py update_index` command for the search index to be updated


springfield/cms/fields.py
```diff
@@ -76,6 +78,27 @@ def to_python(self, value):
 
         return stream_value
 
+    def get_searchable_content(self, value):
+        content = super().get_searchable_content(value)
+        content.extend(self.collect_block_types(value))
```
stevejalim, line 83: nit - nonblocking, but maybe worth a comment:

If individual blocks set `search_index=False`, we don't fully respect it in this override. 

`get_searchable_content` in the superclass bails early when `search_index=False`, but `collect_block_types` here doesn't check it and will still add the type slug. 

Not a real problem today --= no `search_index=False` usage shows up in the codebase and I can't see why we wouldn't want to index the fields (unless maybe they were huge?)

stevejalim (approved):
This is great - thank you @maribedran!

Added one comment inline plus Claude had this to say around l10 impact

>  1. Tokenising is language-agnostic (probably English). Springfield doesn't set WAGTAILSEARCH_BACKENDS.SEARCH_CONFIG (checked settings/base.py — no override), so the Postgres backend passes config=None to to_tsvector() (modelsearch/backends/database/postgres/postgres.py:684), which falls through to Postgres's server-level default_text_search_config GUC — almost always english. So French/German/Portuguese body content still gets indexed and matched on exact tokens and prefixes, but stemming rules are wrong (German compound words, French elisions etc. won't tokenise idiomatically). Exact-match search remains fine; fuzzy-match quality is worse for non-English content. Autocomplete already uses simple config so that path is unaffected.
> 2. No accent-folding. Without unaccent enabled, searching cafe won't match café. Not new to this PR, but relevant because the new indexed fields are much more likely to contain accented text than the title-only status quo.
> 3. Search returns cross-locale results by default. ContentSearchView extends IndexView, whose base filterset (PageFilterSet at wagtail/admin/views/pages/listing.py:70) has filters for date/owner/edited-by/site/child-pages — but no locale filter. So an editor searching "email" gets matches across every locale their permissions expose. Wagtail's admin does have a locale sidebar / switcher elsewhere in the UI when WAGTAIL_I18N_ENABLED=True (which Springfield has), but whether it applies to this specific view depends on template inheritance — worth spot-checking in the browser. If it doesn't filter, editors may want a locale filter added; if it does, ignore.

I'll work on a separate infra PR to run `wagtail_update_index` after a deploy to Dev/Stage/Prod and also demos

springfield/cms/fields.py
maribedran: The only reason I can think of for not including the block in the search index would be if the content were too big. Even in that case, I think we'd want to know that the page includes that block.

maribedran:
@stevejalim Points 1 and 2 are relevant, but from the suggestions Claude is proposing to address them, I don't think it's worth the effort. Exact words still match, so if an editor needs to find specific text in another language, it will still work. I think most of the time, editors will search English text anyway.

PR #1595, opened 2026-07-09
User Routing for WNP

slightlyoffbeat:
## One-line summary

Client-side User Routing framework for What's New Pages — marketing authors typed rules in Wagtail, browser resolves signals and navigates.

The framework lets marketing route Balrog-flow visitors to variant WNPs based on signals (country, locale, Firefox version, default-browser status, UITour state, etc.) via typed rules in the admin — no handwritten conditional rendering. WNP is the first consumer; the dispatcher is page-type-agnostic so landing pages / downloads can adopt it next.

The server does **no rule matching**. When `utm_source=update` is present and the canonical has a live rule, the dispatcher renders a lightweight resolver page carrying the full rule set as JSON. The browser resolves all signals — geo from server-rendered `data-country-code`, UA-derived signals from `Mozilla.Client`, browser state from UITour — evaluates AND-within-rule / OR-across-rules with priority-strict ordering, and navigates. `preview_rule` keeps a server-side 302 shortcut so authors can eyeball variant pages without matching signals; `preview_signal` now flows through a JSON blob into the resolver.

**The client resolver — `media/js/cms/user-routing-resolver.js` — is the most critical part to review.** Correctness of the tri-state evaluator (True / False / null-pending), priority-strict ordering, per-signal + global timeouts, and parity of each sync signal with the deleted server resolver all live in that one file. If the tri-state semantics are wrong, rules that should match won't, or vice versa. If the timeouts are wrong, users hang on the resolver page. If a signal's client-side output diverges from what the admin dropdown implies, marketing authors get silent no-ops.

Grouped topics, most-significant first:

- **Client resolver (`media/js/cms/user-routing-resolver.js`).** Sync-first design: DOM + `Mozilla.Client` resolve inline, then UITour ping + `getConfiguration` for browser-state signals only when a rule actually references one. `GLOBAL_TIMEOUT_MS=1500`, `SIGNAL_TIMEOUT_MS=500`, `PING_TIMEOUT_MS=800`, `LAPSED_MIN_GAP=5` (matches the deleted server constant). Signal parity fixes vs. the deleted server: `platform` coerces `"other"` → `"other-os"` for the admin dropdown; `firefox_version` guards on `isFirefox === true` and rejects the version-0 sentinel that non-Firefox / iOS Firefox report; `lapsed_user` accepts the optional `rv:` prefix on `oldversion`. `is_firefox` intentionally now includes FxiOS (arguably more correct than the server's substring check).

- **Dispatcher + resolver template (`springfield/cms/routing/dispatcher.py`, `templates/cms/routing/resolver.html`).** Renders the resolver page with four JSON blobs (`rules`, `signal-metadata`, `canonical-url`, `preview-signals`). `preview_rule` still 302s; `preview_signal` is now embedded into the preview-signals blob. Router URL responses carry `Link: <canonical>; rel="canonical"` for SEO/AEO. Target URLs in the rules blob get `routed_from` / `routed_rule` / `routed_mode=client` attribution pre-appended; incoming spoofed values stripped.

- **Signal registry (`springfield/cms/routing/__init__.py`, `signals.py`).** Metadata-only catalog (no server resolvers). Twelve signals registered; server-side vs. client-side is a categorization for the admin dropdown, not a resolution mechanism (all resolution is client-side).

- **Models + admin (`springfield/cms/models/routing.py`, `wagtail_hooks.py`).** `RoutingRule` (ParentalKey to `WhatsNewPage2026`, Orderable, status enum). `RoutingCondition` (one-to-many, AND semantics). `WhatsNewPage2026.routing_paused` kill switch. Nested "User Routing" admin menu (Rules snippet + auto-generated Signals reference).

- **WNP wrapper (`springfield/firefox/views.py::wnp_dispatch`).** Enforces the `utm_source=update` gate; organic traffic bypasses routing entirely. The `$` anchor in `urls.py` is load-bearing — regression test guards the redirect-loop it caused during development.

- **Test rework.** 88 Python integration tests assert dispatch gate, resolver JSON contract, preview_rule/preview_signal flows, spoofed-param stripping, canonical link header, no-store on preview responses. 34 Jasmine specs cover the client resolver's pure logic + each SYNC_RESOLVER + UITour extractor.

- **Deletions.** `evaluator.py` (189 lines) and `server_resolvers.py` (153 lines). All server-side rule matching and per-signal resolvers removed. `Signal.cache_safe` becomes documentation-only — Fastly VCL work is no longer a prerequisite for going Live under the client-side architecture (client-resolved signals don't affect the cache key).

## Testing

**Setup**

```bash
./manage.py migrate
./manage.py runserver
```

**1. Create pages in Wagtail admin (`/cms-admin/`)**

Under Pages → What's New Index (create one if missing):
1. Add child page **What's New Page 2026** with slug `120` → publish. This is the canonical.
2. Add another **What's New Page 2026** as a child of the `120/` page with slug `us-variant` → publish. (Variant must be a descendant of canonical.)

**2. Attach a routing rule**

Edit the canonical, User Routing tab:
- Rule name `us-users-120`, target `us-variant`, condition signal=`country` op=`is` values=`US`, status Live. Save → Publish.

**3. Verify the gate**

Set `DEV_GEO_COUNTRY_CODE=US`, restart the server.
- `/en-US/whatsnew/120/` → canonical served (no marker → gate closed).
- `/en-US/whatsnew/120/?utm_source=organic` → still canonical.

**4. Verify client-side resolution**

- `/en-US/whatsnew/120/?utm_source=update` → resolver page renders briefly, browser navigates to `/whatsnew/120/us-variant/` with `routed_from=120&routed_rule=us-users-120&routed_mode=client` appended.
- Change env to `DEV_GEO_COUNTRY_CODE=DE`, restart → resolver page renders, browser navigates back to canonical (rule doesn't match).

**5. Verify AND conditions**

Edit the rule and add a second condition: `lapsed_user` `is` `true`. Save.
- `.../whatsnew/120/?utm_source=update` (no oldversion) → resolver renders, then canonical (`lapsed_user` unresolved → AND blocks).
- `.../whatsnew/120/?utm_source=update&oldversion=115` → resolver renders, then variant.
- `.../whatsnew/120/?utm_source=update&oldversion=rv:115` → same result (rv: prefix accepted).

**6. Verify UITour signal**

Firefox `about:config`: `browser.uitour.testingOrigins = http://localhost:8000`.
Change the rule to condition `default_browser` = `true`. Visit `.../whatsnew/120/?utm_source=update` in Firefox → resolver briefly shows "Contacting Firefox…", then navigates based on whether Firefox is default.

**7. Verify preview**

While logged in to Wagtail admin:
- `.../whatsnew/120/?preview_rule={rule_id}&utm_source=update` → 302 to variant. Response `Cache-Control: no-store`.
- `.../whatsnew/120/?preview_signal=country:DE&utm_source=update` → resolver renders with the fake `country=DE` in the preview-signals blob → navigates to canonical (rule misses). `no-store`.
- Preview URLs when logged out are silently ignored.

**8. Verify pause routing**

Toggle `routing_paused` on the canonical → Save.
- `.../whatsnew/120/?utm_source=update` → canonical (routing skipped, no resolver).
- `.../whatsnew/120/?preview_rule={id}&utm_source=update` as admin → still 302 to variant (preview bypasses pause).

**9. Verify Signals reference + Rules admin**

Wagtail admin: **User Routing → Signals reference** lists all 12 signals grouped by resolver type. **User Routing → Rules** index shows the rule columns.

**10. Verify variant-URL loop guard**

`/en-US/whatsnew/120/us-variant/?utm_source=update` → renders variant directly, no re-entry.

**11. Run the tests**

```bash
pytest springfield/cms/routing springfield/cms/tests/test_routing_rule.py springfield/firefox/tests/test_wnp_dispatch.py
npm run jasmine
```


springfield/cms/migrations/0115_remove_routingrule_condition_and_more.py
```diff
@@ -0,0 +1,52 @@
...
+import springfield.cms.models.routing
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0114_alter_routingrule_options_and_more"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
stevejalim: We should get claude to drop and regenerate these migrations so that there's no unnecessary additon of a field (in 0113) then removal of it here.

stevejalim (changes requested):
Sorry - blocking merge on this until we can work out the CDN implications. Right now, I think this won't fly because as far as the CDN is concerned, it may not be able to differentiate reliably between request for a WNP with content decided by server-side rules and one that's handled client-side - and especially so if there's a combination of both URL-params and UITour signals.

Will have a think about it, and I might be wrong, but flagging this early in case it got merged via someone else's approval

springfield/cms/templates/cms/routing/resolver.html
```diff
@@ -0,0 +1,77 @@
...
+      margin: 0;
+    }
+    @media (prefers-color-scheme: dark) {
+      #user-routing-status { color: #aaa; }
+    }
+  </style>
+</head>
+<body>
+  <main>
+    <p id="user-routing-status">Preparing your page…</p>
```
stephaniehobson, line 78: This needs a `<noscript>` fallback.

I also want to note that the UITour check can add several seconds to the response and we should monitor to make sure that doesn't result in fewer total views.

stephaniehobson (changes requested):
Commenting to block this until I've done a full PR review.

springfield/cms/models/routing.py
```diff
@@ -0,0 +1,345 @@
...
+        on_delete=models.CASCADE,
+        related_name="routing_rules",
+        help_text="The canonical page this rule attaches to.",
+    )
+
+    target_page = models.ForeignKey(
+        "wagtailcore.Page",
+        on_delete=models.PROTECT,
+        related_name="routing_rules_as_target",
+        help_text="The variant page to route matching users to.",
```
stephaniehobson, line 198: Should this be restricted to only child pages of this canonical page?

springfield/cms/migrations/0116_remove_routingrule_expected_value_and_more.py
```diff
@@ -0,0 +1,73 @@
...
+        migrations.AddField(
+            model_name="routingrule",
+            name="expected_values",
+            field=models.TextField(
+                blank=True,
+                default="",
+                help_text=(
+                    "One or more values, one per line (commas also accepted). "
+                    "For enum-valued signals, values must match the exact allowed set "
+                    "(check the signal's help entry). Booleans: 'true' or 'false'. "
```
stephaniehobson: **(Enhancement, non-blocking):** It'd be nice to link to the help entry here.

springfield/cms/models/routing.py
```diff
@@ -0,0 +1,345 @@
...
+            "values": parsed,
+        }
+
+    @condition.setter
+    def condition(self, value):
+        """Accept both the legacy ``{signal, equals}`` shape and the new
+        ``{signal, op, values}`` shape, decomposing either into the
+        structured fields.
+
+        Legacy support preserves the fluent ``RoutingRule(condition={...})``
```
stephaniehobson: **Todo (blocking):**

Since this code is net-new there's no "legacy" signal to preserve. Please re-write the tests to consider only the structure we plan to publish.

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,350 @@
...
+
+    var navigated = false;
+    function navigate(url) {
+        if (navigated) return;
+        navigated = true;
+        setStatus('Redirecting…');
+        try {
+            window.location.replace(url);
+        } catch (e) {
+            window.location.href = url;
```
stephaniehobson, line 243: **Bug (blocking):** If there is no matching variant, the user will be sent back to the original URL... which will check for variants again. 

From the PR description, it sounds like you ran into this problem during development with the child pages as well and only fixed it there.

Maybe appending a param to the query string so it doesn't loop.

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,350 @@
...
+    // Entry
+    // -----------------------------------------------------------------------
+
+    if (!rules.length) {
+        done(canonicalUrl);
+        return;
+    }
+
+    if (
+        !window.Mozilla ||
```
stephaniehobson, line 491: **todo (blocking)**: Don't forget to include the UI Tour library ;)

springfield/cms/routing/evaluator.py
```diff
@@ -0,0 +1,148 @@
...
+    operator = _condition_operator(rule)
+    if operator == "is":
+        return is_in_list
+    if operator == "is_not":
+        return not is_in_list
+    # Unknown operator (shouldn't happen with clean() validation).
+    return False
+
+
+def evaluate_rules(rules: Iterable, resolved_signals: dict[str, Any]) -> EvaluationResult:
```
stephaniehobson: At the moment this function is only resolving server side rules and higher-priority client-side rules are not part of the evaluation.

(So, if your first priority is to target people with Firefox set as their default browser but have a lower priority rule for people in GEO=de --People with Firefox set as default in de, will see the de page, not the set as default one.

stephaniehobson: I think this means that the client side resolver will have to be able to handle all server side cases.

Which begs the question - why do all the work twice? Maybe we should only have the client side resolver (good for maintainability, not necessarily for page speed)

springfield/firefox/views.py
```diff
@@ -938,6 +938,72 @@ def detect_channel(version):
...
+    # Use ``.filter().first()`` rather than ``.get()`` so a Locale table with
+    # a duplicate row (data-import race, manual DB edit) still falls through
+    # to the Fluent evergreen path instead of 500-ing on MultipleObjectsReturned.
+    # The fallback is deliberately catch-all here — the whole point of the
+    # try block is "don't let locale lookup take down the WNP path."
+    locale = WagtailLocale.objects.filter(language_code=lang_code).first()
+    if locale is None:
+        return _fallback()
+
+    canonical = WhatsNewPage2026.objects.live().public().filter(locale=locale, slug=version).first()
```
stephaniehobson: This needs the same filters that scope the slug lookup to the WhatsNewIndexPage's direct children as are on the _get_evergreen_wnp_redirect (line 1113ish)

stephaniehobson (commented):
Two high-level pieces of feedback:

1) We need more defined parent/child scoping.
  - children should only be allowed at depth 1, not infinitely nested
  - children should not allow user routing rules to be defined
  - choosing variant pages would be faster if the routing only included child pages in the page selector

2) We need a strategy for handling a mix of server and client side rules
  - maybe all processing happens client side? (though, I do like the speed of the server side processing, but I don't like duplicating the rules more than once)
  - maybe server can evaluate true/false for all its rules and then pass to the client for final evaluation if there are client rules defined

springfield/cms/migrations/0115_remove_routingrule_condition_and_more.py
slightlyoffbeat: consolidated to 1, and I'll continue to consolidate as we work through this.

springfield/cms/templates/cms/routing/resolver.html
slightlyoffbeat: Added  `<noscript><meta http-equiv="refresh">` at resolver.html, sends JS-disabled users to canonical.

springfield/cms/models/routing.py
slightlyoffbeat: Yes, `RoutingRule.clean()` enforces descendant-only now

springfield/cms/models/routing.py
slightlyoffbeat: tests rewritten under the final structure in test_routing_rule.py and test_wnp_dispatch.py.

media/js/cms/user-routing-resolver.js
slightlyoffbeat: Fixed. The client falls back to a URL carrying `routed_mode=none`, and wnp_dispatch now short-circuits whenever any `routed_mode` is present. Regression test at test_wnp_dispatch.py::test_routed_mode_param_short_circuits.

media/js/cms/user-routing-resolver.js
slightlyoffbeat: now in the user_routing_resolver bundle

springfield/cms/routing/evaluator.py
slightlyoffbeat: Yup. Chatted with Steve, and decided to move in the direction of clietn side resolver. WIll chat more on this in its own comment below since this PR has changed a good amount.

springfield/firefox/views.py
slightlyoffbeat: Fixed

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,529 @@
...
+        },
+
+        // getConfiguration('fxa')
+        fxa_signed_in: function (data) {
+            if (!data || typeof data.setup !== 'boolean') return undefined;
+            return data.setup;
+        },
+
+        // getConfiguration('aiControls')
+        ai_controls: function (data) {
```
stevejalim, line 195: **Null not guarded in `ai_controls` extractor**

If Firefox returns `{ default: null }`, this extractor returns `null` (not `undefined`). The signal is then considered resolved (key exists in `resolvedSignals`), so `conditionMatches` runs `values.indexOf(null)` against the string enum. Under `op: "is"` that is harmless. But under `op: "is_not"` it returns `true` — a rule like `ai_controls is_not [enabled]` silently matches every user whose Firefox returned `null`.

The other extractors guard `typeof x === 'boolean'` to avoid exactly this class of bug. This one should guard `typeof data.default === 'string'` (or at minimum `!== null`) before returning.

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,529 @@
...
+        // higher-priority rule has been definitively decided (matched or
+        // not-matched). If any earlier-in-priority rule is still pending,
+        // hold the decision.
+        //
+        // Takes ``ruleList`` as an argument (rather than reading a closure)
+        // so the pure evaluator is testable in isolation. Boot callers pass
+        // the module-scope ``rules`` array; tests pass constructed rule
+        // lists.
+        var pending = false;
+        var sorted = ruleList.slice().sort(function (a, b) {
```
stevejalim, line 304: **`Array.sort` stability not guaranteed for priority ties**

When two rules share the same `sort_order` (common case: both `null` → both `0`), the resolved order is undefined on pre-V8-7.0 engines. The Jasmine test `"treats missing priority as 0"` asserts `result.matched === r1`, which assumes `r1` stays ahead of `r2` — that relies on stable sort.

Modern Firefox/Chrome sort stably, but the priority-strict guarantee the evaluator documents is not enforced for ties. Easiest fix: secondary sort key on `rule.name`, or have the server inject a stable tiebreaker field so ties resolve deterministically regardless of engine.

springfield/firefox/views.py
```diff
@@ -938,6 +941,115 @@ def detect_channel(version):
...
+    if request.GET.get(trigger_param) != trigger_value:
+        return _fallback()
+
+    # Loop-breaker: if the request already carries ``routed_mode``, routing
+    # has already run for this navigation (either the client resolved to
+    # canonical with ``routed_mode=none``, or a downstream link cycled
+    # back through the WNP URL). Re-entering the router would render the
+    # resolver again → the client would re-navigate → infinite loop.
+    # See dispatcher.ROUTED_MODE_NONE for the fallback attribution.
+    if request.GET.get("routed_mode"):
```
stevejalim, line 1053: **Spoofed `routed_mode` bypass is intentional but unexplained in code**

This breaks on _any_ non-empty `routed_mode`, including `?utm_source=update&routed_mode=xyz` crafted by a user. The PR description acknowledges the tradeoff, but the in-code comment only mentions the legitimate `routed_mode=none` loop-breaker case.

Without a note here, a future reviewer is likely to "fix" this to `== ROUTED_MODE_NONE` to tighten it — which re-introduces the redirect loop for any future `routed_mode` value the framework might emit. One sentence stating that spoofed values are an intentional, accepted bypass would prevent that regression.

media/js/cms/user-routing-resolver.js
slightlyoffbeat: Fixed. Extractor now guards on `typeof data.default === 'string'`, matching the bool-extractor pattern. Null / non-string values stay unresolved instead of coercing under `is_not` and matching every user whose Firefox returned null. Regression test at tests/unit/spec/cms/user-routing-resolver.js covering `{ default: null }`, `{ default: 0 }`, and `{ default: true }`.

media/js/cms/user-routing-resolver.js
slightlyoffbeat: Fixed. Server payload now emits `id` (DB PK) on each rule; JS sort key is `(priority, id)` ascending, so equal-priority rules order deterministically regardless of engine sort-stability. Chose `id` over `name` since rule renames shouldn't reshuffle evaluation order.

springfield/firefox/views.py
slightlyoffbeat: Comment expanded to spell out the tradeoff and add a "do not tighten" warning for future reviewers, the any-non-empty check is deliberate future-proofing (any new framework-emitted mode is honored without a code change here) at the cost of accepting client-crafted values as a bypass.

springfield/cms/migrations/0116_remove_routingrule_expected_value_and_more.py
slightlyoffbeat: Consolidated into 0122_user_routing_framework.py. Live signal help is now the auto-generated Signals reference at /cms-admin/user-routing/signals/, linked from the User Routing sidebar submenu.

media/static-bundles.json
```diff
@@ -841,6 +850,12 @@
       ],
       "name": "wagtailadmin-locale-badges"
     },
+    {
+      "files": [
+        "js/cms/wagtailadmin-user-routing-condition-help.js"
+      ],
+      "name": "wagtailadmin-user-routing-condition-help"
```
stevejalim, line 857: nit: this is `kebab-case` but all the other bundles are `snake_case`

springfield/admin/templates/cms/routing/admin/rules_index.html
```diff
@@ -0,0 +1,35 @@
...
+ This Source Code Form is subject to the terms of the Mozilla Public
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+{#
+ This Source Code Form is subject to the terms of the Mozilla Public
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
```
stevejalim, line 11: Change: delete repeated licence comment

```suggestion
```

springfield/admin/templates/cms/routing/admin/rules_index.html
```diff
@@ -0,0 +1,35 @@
...
+{% block listing %}
+    <div class="nice-padding" style="max-width: 44rem; padding-block: 0.75rem 1rem; color: #666;">
+        <p style="margin: 0; line-height: 1.5;">
+            Rules appear here once their parent canonical page has been
+            <strong>published</strong>. Draft-only pages hold their rules in a
+            page revision, not in this table. A rule's own <em>status</em>
+            (Draft / Live / Archived) is separate — only <strong>Live</strong>
+            rules are actually evaluated by the router; Draft and Archived
+            rules still show here so authors can find and edit them.
```
stevejalim, line 31: comment, nonblocking: This definitely helps, but I still think people will have questions. Do we want a complementary comment on the draft revision page?

springfield/cms/models/__init__.py
```diff
@@ -6,3 +6,4 @@
 from .snippets import *  # noqa
 from .pages import *  # noqa
 from .images import *  # noqa
+from .routing import *  # noqa
```
stevejalim, line 9: Comment, nonblocking: This separation is handy - if we're going to apply this to Bedrock as well, it would be clean to move it all to a separate Python package

springfield/cms/models/pages.py
```diff
@@ -1294,10 +1294,22 @@ def serve(self, request):
 
 
 class WhatsNewPage2026(PageThemeMixin, PreFooterImageMixin, UTMParamsMixin, QRCodeFloatingSnippetMixin, AbstractSpringfieldCMSPage):
-    """A 2026 version of the What's New page with optional upper/lower split layout."""
+    """A 2026 version of the What's New page with optional upper/lower split layout.
 
-    parent_page_types = ["cms.WhatsNewIndexPage"]
-    subpage_types = []
+    Two roles in the page tree:
```
stevejalim, line 1299: Suggestion, nonblocking

```suggestion
    This page type has two roles in the page tree:
```

springfield/cms/models/pages.py
```diff
@@ -1294,10 +1294,22 @@ def serve(self, request):
...
+
+    * **Canonical** — direct child of ``WhatsNewIndexPage`` at
+      ``/whatsnew/<version>/``. Rendered to organic traffic and to any
+      Balrog-flow user whose request doesn't match a routing rule.
```
stevejalim, line 1303: Suggestion, clarity/remove tribal knowledge from Claude

```suggestion
      user sent to WNP following an update whose request doesn't match a routing rule.
```

springfield/cms/models/pages.py
```diff
@@ -1294,10 +1294,22 @@ def serve(self, request):
...
+    * **Variant** — child of a canonical WNP at
+      ``/whatsnew/<version>/<variant-slug>/``. Reached only via a routing
+      rule 302 (server-side) or the resolver page (client-side, Step 3b).
```
stevejalim, line 1306: Stale comment? We no longer have server-side rules

```suggestion
      rule via the resolver page.
```

springfield/cms/models/pages.py
```diff
@@ -1317,6 +1329,17 @@ class WhatsNewPage2026(PageThemeMixin, PreFooterImageMixin, UTMParamsMixin, QRCo
         use_json_field=True,
     )
 
+    routing_paused = models.BooleanField(
```
stevejalim, line 1332: Comment, nonblocking: once we're happy with the routing behaviour as tested on WNP, we should move as much as possible into a mixin class, making it easier to add the fields and `panel` setup to other pages

springfield/cms/models/pages.py
```diff
@@ -1351,10 +1374,47 @@ class WhatsNewPage2026(PageThemeMixin, PreFooterImageMixin, UTMParamsMixin, QRCo
...
+    # Rules attached here are evaluated when this page is the canonical for a
+    # /whatsnew/{version}/ request. Variants (children of a canonical) may
+    # also technically render this tab today; rules on variants are inert
+    # because dispatch only queries rules attached to the canonical.
+    user_routing_panels = [
+        HelpPanel(
+            content=(
+                "<p>Rules here apply when <strong>this page is the canonical "
+                "for a release</strong> (a direct child of the What's New Index). "
+                "Adding a rule on a variant page is rejected at save time — "
```
stevejalim, line 1387:
```suggestion
                "Adding a rule on a variant page will be rejected at save time: just "
```

springfield/cms/models/pages.py
```diff
@@ -1372,6 +1432,64 @@ def get_utm_campaign(self):
     def noindex(self):
         return True
 
+    @staticmethod
+    def _is_canonical_wnp(page):
+        """Single-source-of-truth for 'is ``page`` a canonical WNP?' — a
+        WhatsNewPage2026 whose direct parent is a WhatsNewIndexPage
+        (i.e., depth-2 in the /whatsnew/ subtree, the shape Balrog targets).
```
stevejalim, line 1439: Drop the depth assumption (may not _always_ be true) and Balrog tribal-knowledge ref - not needed here
```suggestion
```

springfield/cms/models/pages.py
```diff
@@ -1372,6 +1432,64 @@ def get_utm_campaign(self):
...
+
+        Consumed by :meth:`can_create_at` (rejecting variants-of-variants),
+        :meth:`can_move_to` (same for moves), and
+        :meth:`can_host_routing_rules` (framework hook). Any future
+        refactor that widens/narrows the canonical definition changes one
+        place and all three checks stay in lockstep.
```
stevejalim, line 1445: Comment style is not pythonic 

```suggestion
        Consumed by `self.can_create_at()` (rejecting variants-of-variants),
        `self.can_move_to()` (same for moves), and
        `self.can_host_routing_rules()` (framework hook). Any future
        refactor that widens/narrows the canonical definition changes one
        place and all three checks stay in lockstep.
```

springfield/cms/models/pages.py
```diff
@@ -1372,6 +1432,64 @@ def get_utm_campaign(self):
...
+        that is itself a variant (i.e., not itself a canonical).
+        """
+        if not super().can_create_at(parent):
+            return False
+        if isinstance(parent.specific, WhatsNewPage2026) and not cls._is_canonical_wnp(parent):
+            return False
+        return True
+
+    def can_move_to(self, parent):
+        """Mirror :meth:`can_create_at` for the Wagtail move flow.
```
stevejalim, line 1469:
```suggestion
        """Mirror `self.can_create_at()` for the Wagtail move flow.
```

springfield/cms/models/pages.py
```diff
@@ -1372,6 +1432,64 @@ def get_utm_campaign(self):
...
+        rejects at creation. Same predicate, same rejection.
+        """
+        if not super().can_move_to(parent):
+            return False
+        if isinstance(parent.specific, WhatsNewPage2026) and not self._is_canonical_wnp(parent):
+            return False
+        return True
+
+    def can_host_routing_rules(self):
+        """Framework hook (see ``RoutingRule.clean()``): only canonical WNPs
```
stevejalim, line 1483:
```suggestion
        """Framework hook (see RoutingRule.clean()): only canonical WNPs
```

springfield/cms/models/pages.py
```diff
@@ -1372,6 +1432,64 @@ def get_utm_campaign(self):
...
+        can host rules.
+
+        The hook is consumer-defined so the User Routing framework doesn't
+        need to know anything about WNP-specific tree structure — a future
+        consumer implements its own ``can_host_routing_rules()`` on its
```
stevejalim, line 1488:
```suggestion
        consumer implements its own can_host_routing_rules() on its
```

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * User Routing — client-side resolver.
+ *
+ * Runs the full rule evaluator. The server does no rule matching under
+ * this architecture — it just ships the resolver page with:
```
stevejalim, line 11: Removing earlier-iteration comment cruft
```suggestion
 * Runs the full rule evaluator. The server does no rule matching 
 * at all – it just ships the resolver page with:
```

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        locale: function () {
+            var v = document.documentElement.getAttribute('lang');
+            return v || undefined;
+        },
+
+        // Derived: (target_version from URL path) - (oldversion from query) >= LAPSED_MIN_GAP.
+        // Target version comes from the ``/whatsnew/{version}/`` URL segment.
+        // ``oldversion`` accepts an optional ``rv:`` prefix (matches the shape
+        // some Gecko builds emit in UA strings) — the deleted server resolver
+        // stripped it, so we keep parity here.
```
stevejalim, line 83: Dropping stale comment content

```suggestion
        // some Gecko builds emit in UA strings)
```

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        lapsed_user: function () {
+            var pathMatch =
+                window.location.pathname.match(/\/whatsnew\/(\d+)\//);
+            if (!pathMatch) return undefined;
+            var target = parseInt(pathMatch[1], 10);
+            if (isNaN(target)) return undefined;
+            var params = new window.URLSearchParams(window.location.search);
+            var oldRaw = params.get('oldversion');
+            if (!oldRaw) return undefined;
+            var oldMatch = /^(?:rv:)?(\d{1,3})/.exec(oldRaw);
```
stevejalim, line 93: Question, blocking: are we 100% surew that `oldversion` is always `XXX` and never `XXX.Y` or `XXX.Y.Z`

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+            // module. Coerce so a marketer's ``platform is other-os`` rule
+            // still matches the audience the label implies.
+            return p === 'other' ? 'other-os' : p;
+        },
+        // Note: ``is_firefox`` includes iOS Firefox (FxiOS). Mozilla.Client
+        // treats FxiOS as Firefox; the deleted server resolver only matched
+        // the literal string "Firefox" in the UA and excluded FxiOS. The
+        // client's answer is arguably more correct — but rules that want to
+        // exclude iOS should combine ``is_firefox=true`` with
+        // ``platform is [osx, linux, windows, android]``.
```
stevejalim, line 119: Question (that might answer itself as I read down the PR): Is this potential footgun documented in the UI?

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+            return null;
+        }
+    }
+
+    var rules = readJSON('user-routing-rules') || [];
+    var signalMetadata = readJSON('user-routing-signal-metadata') || {};
+    var canonicalUrl = readJSON('user-routing-canonical-url') || '/';
+    var previewSignals = readJSON('user-routing-preview-signals') || {};
+    var status = document.getElementById('user-routing-status');
+
```
stevejalim, line 226: I'm assuming all of these values from the JSON can handle being cached at the CDN for 10 mins after being changed

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+    function done(url) {
+        window.clearTimeout(globalTimer);
+        navigate(url);
+    }
+
+    // -----------------------------------------------------------------------
+    // Rule evaluation — tri-state, AND-within-rule, OR-across-rules
+    // -----------------------------------------------------------------------
+
+    function conditionMatches(condition, resolved) {
```
stevejalim, line 260: Imagining a year from now, I'd welcome more comments or a docstring to hand-hold me through this

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    function ruleMatches(rule, resolved) {
+        var conditions = rule.conditions || [];
+        if (!conditions.length) return false;
+        var anyPending = false;
+        for (var i = 0; i < conditions.length; i++) {
+            var outcome = conditionMatches(conditions[i], resolved);
+            if (outcome === false) return false;
+            if (outcome === null) anyPending = true;
```
stevejalim, line 282: nit: here's a case where braces help readability, IMO
```suggestion
            if (outcome === null) {
                anyPending = true;
            }
```

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        }
+        if (anyPending) return null;
+        return true;
+    }
+
+    function evaluate(ruleList, resolved) {
+        // Priority-strict: a lower-priority rule can only "win" once every
+        // higher-priority rule has been definitively decided (matched or
+        // not-matched). If any earlier-in-priority rule is still pending,
```
stevejalim, line 291: what does `earlier-in-priority` mean in this context? evaluated earlier but we're waiting to see if a higher-priority rule will be evaluated that would beat this one?

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        // hold the decision.
+        //
+        // Takes ``ruleList`` as an argument (rather than reading a closure)
+        // so the pure evaluator is testable in isolation. Boot callers pass
+        // the module-scope ``rules`` array; tests pass constructed rule
+        // lists.
+        //
+        // Ties on ``priority`` fall back to ``id`` (server-emitted DB PK)
```
stevejalim, line 299:
```suggestion
        // Ties on ``priority`` fall back to ``id`` (generated by the server; the primary key for the database row)
```

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        for (var i = 0; i < conditions.length; i++) {
+            var outcome = conditionMatches(conditions[i], resolved);
+            if (outcome === false) return false;
+            if (outcome === null) anyPending = true;
+        }
+        if (anyPending) return null;
+        return true;
+    }
+
+    function evaluate(ruleList, resolved) {
```
stevejalim, line 288: Renaming `resolved` to `resolvedSignals` might improve first-read understanding

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        typeof window.Mozilla.UITour.ping !== 'function'
+    ) {
+        abandonUITourSignals();
+        onSignalUpdate();
+        if (navigated) return;
+        done(canonicalUrl);
+        return;
+    }
+
+    setStatus('Contacting Firefox…');
```
stevejalim, line 502: Couple of things here: 

1. We should localize that string (can pull it from a data-attr in the resolver template where we have the string translated via the CMS or Fluent)
2. Phrasing: maybe 'Checking Firefox status' or 'Checking your Firefox's capabilities', rather than suggesting we're pingning the FF mothership?

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+
+    // -----------------------------------------------------------------------
+    // Navigation — every code path ends here at most once
+    // -----------------------------------------------------------------------
+
+    var navigated = false;
+    function navigate(url) {
+        if (navigated) return;
+        navigated = true;
+        setStatus('Redirecting…');
```
stevejalim, line 239: We should add L10N support for this and all `setStatus` messages - see similar comment later

media/js/cms/user-routing-resolver.js
```diff
@@ -0,0 +1,542 @@
...
+        done(canonicalUrl);
+    }, PING_TIMEOUT_MS);
+
+    try {
+        window.Mozilla.UITour.ping(function () {
+            if (pingSettled) return;
+            pingSettled = true;
+            window.clearTimeout(pingTimer);
+
+            setStatus('Checking your settings…');
```
stevejalim, line 520: Same L10N comment as before

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
...
+    the rule model to parse the operator's expected value from admin input.
+    """
+
+    BOOL = "bool"
+    STRING = "string"
+    INT = "int"
+
+
+@dataclass(frozen=True)
+class Signal:
```
stevejalim, line 65: concern, semi-blocking: `Signal` is a _very_ loaded term in the Django ecosystem because Django also has a `Signal` model.

I suggest we rename this to `RoutingSignal` or similar to namespace it

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
+Signal registry — the catalog of every conditional signal the routing layer
+knows about.
+
+A signal is a named thing that rules reference by name (e.g. ``lapsed_user``,
+``ai_controls``, ``country``). Under the client-side architecture, every
```
stevejalim, line 10:
```suggestion
``ai_controls``, ``country``). Defined in the CMS and served to the client, every
```

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
...
+from dataclasses import dataclass
+from enum import Enum
+
+
+class ResolverType(str, Enum):
+    """How a signal's value is obtained (informational; the client-side
+    resolver has the actual per-signal resolution logic)."""
+
+    SERVER_SIDE = "server_side"
+    CLIENT_SIDE_STATE = "client_side_state"
```
stevejalim, line 48: Do we need this `ResolverType` class at all now if everything is resolved client-side?

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@dataclass(frozen=True)
+class Signal:
+    """A registered conditional signal.
+
+    Attributes:
+        name: The unique identifier rules reference (e.g. ``"lapsed_user"``).
+        description: Human-readable summary shown in the CMS admin.
+        resolver_type: SERVER_SIDE (server-rendered / UA / URL) or
```
stevejalim, line 71: Again, referencce to SERVER_SIDE may be redundant now

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
...
+        supports_in_page_swap: Whether this signal may gate an in-page block.
+        uitour_key: For CLIENT_SIDE_STATE signals: the ``Mozilla.UITour``
+            ``getConfiguration`` key to fetch (e.g. ``"appinfo"``, ``"fxa"``).
+            Emitted into the resolver page's ``signal_metadata`` JSON so the
+            client knows which key to fetch per signal.
+        uitour_extractor: For CLIENT_SIDE_STATE signals: the name of a
+            client-side extractor that converts the UITour response into
+            the signal's value. Purely for documentation / cross-reference
+            — the client-side resolver library maps signal name to extractor
+            function directly.
```
stevejalim, line 84: If this param is intert / documentation only, does it really have much value vs the risk of user confusion?

springfield/cms/routing/signals.py
```diff
@@ -0,0 +1,155 @@
...
+                f"Signal {signal.name!r} must support at least one of routing / in-page swap",
+            )
+
+        if signal.resolver_type == ResolverType.CLIENT_SIDE_STATE:
+            if not signal.uitour_key or not signal.uitour_extractor:
+                raise SignalRegistrationError(
+                    f"client_side_state signal {signal.name!r} must declare uitour_key and uitour_extractor",
+                )
+        elif signal.resolver_type != ResolverType.SERVER_SIDE:
+            raise SignalRegistrationError(f"Unknown resolver_type: {signal.resolver_type!r}")
```
stevejalim, line 138: I think we can drop this clause, and remove `if signal.resolver_type == ResolverType.CLIENT_SIDE_STATE:` above, too?

springfield/cms/templates/cms/routing/admin/signals_reference.html
```diff
@@ -0,0 +1,154 @@
...
+This is a Django template (extends wagtailadmin/base). Comment syntax is
+the Django comment tag, not Jinja's hash-brace form. Extends must be the
+first tag in the file — the license comment lives right below it.
+{% endcomment %}
+
+{% block titletag %}User Routing signals reference{% endblock %}
+
+{% block extra_css %}
+  {{ block.super }}
+  <style>
```
stevejalim, line 28: blocking: this will need to move to a proper CSS file/bundle because CSP will deny it

springfield/cms/templates/cms/routing/resolver.html
```diff
@@ -0,0 +1,93 @@
...
+  <meta name="robots" content="noindex,nofollow">
+  {# JS-disabled fallback: meta-refresh redirects to the canonical URL so
+     users without JavaScript still land on a real page instead of staring
+     at the transient "Preparing…" loading screen forever. The canonical
+     URL is server-rendered here (not read from the JSON blob below) since
+     no JS is available to parse the blob. #}
+  <noscript>
+    <meta http-equiv="refresh" content="0;url={{ canonical_url_raw|e }}">
+  </noscript>
+  <title>Preparing your page…</title>
```
stevejalim, line 47: L10N klaxon 📯

springfield/cms/templates/cms/routing/resolver.html
```diff
@@ -0,0 +1,93 @@
...
+      margin: 0;
+    }
+    @media (prefers-color-scheme: dark) {
+      #user-routing-status { color: #aaa; }
+    }
+  </style>
+</head>
+<body>
+  <main>
+    <p id="user-routing-status">Preparing your page…</p>
```
stevejalim, line 78: Also need L10N support

springfield/firefox/urls.py
```diff
@@ -142,9 +142,21 @@
     ),
     # START What's New Page (WNP) paths
     # 1. Legacy version format: MAJ.MIN/variant.patch (127.1a, 139.0.1, etc) rather than just MAJ
-    re_path(f"^whatsnew/(?P<version>{version_re})/", views.WhatsnewView.as_view(), name="firefox.whatsnew_legacy"),
-    # 2. New version format, which should be served from the CMS, but falls back to evergreen page
-    re_path(r"^whatsnew/(?P<version>[1-9]\d{2})/", prefer_cms(views.WhatsnewView.as_view()), name="firefox.whatsnew"),
+    # Trailing `$` for the same reason as the 3-digit pattern below: without it,
+    # any URL under `/whatsnew/<version>/*` matches this pattern (with the leftover
+    # path unconsumed) and re-enters WhatsnewView, causing surprises for future
+    # variant / sub-page routing under legacy version formats.
```
stevejalim, line 148: nit: true but overly verbose. Could lose it all

springfield/firefox/urls.py
```diff
@@ -142,9 +142,21 @@
...
+    re_path(f"^whatsnew/(?P<version>{version_re})/$", views.WhatsnewView.as_view(), name="firefox.whatsnew_legacy"),
+    # 2. New version format: `wnp_dispatch` runs first to honor any live
```
stevejalim, line 150:
```suggestion
    # `wnp_dispatch` runs first to honor any live
```

springfield/firefox/urls.py
```diff
@@ -142,9 +142,21 @@
...
+    #    RoutingRule against server-resolvable signals; on no match, it
```
stevejalim, line 151: This looks like it needs a revisit - there are no server-resolvable signals

springfield/firefox/urls.py
```diff
@@ -142,9 +142,21 @@
...
+    #    delegates internally to prefer_cms(WhatsnewView) so the CMS canonical
+    #    (or Fluent evergreen fallback) is rendered as before.
+    #    The trailing `$` is load-bearing: variant URLs like
+    #    `/whatsnew/120/welcomeback/` MUST fall through to Wagtail page routing.
+    #    Without `$`, Django matches the prefix and re-runs wnp_dispatch on the
+    #    variant, which then finds canonical 120 again → routing rule matches
+    #    again → redirect loop.
+    re_path(r"^whatsnew/(?P<version>[1-9]\d{2})/$", views.wnp_dispatch, name="firefox.whatsnew"),
```
stevejalim, line 159: I can't see how this is needed at all now - if WNP can server its resolver via the `routing` module, isn't this `wnp_dispatch` view redundant?

stevejalim (commented):
Moving forward, but I think there's redundant code here, and definitely redundant code comments. Let me know what you think, partic around my question about `wnp_dispatch`? If it's not actually redundant, I thikn we should make it so because I can see a version of this where we don't need it and the routing layer does it. Could even hook that in to the start of Page.serve() to avoid having to patch in stuff

PR #1599, opened 2026-07-10
Adding easter egg logo

slightlyoffbeat:
## Summary
Bring the `/landing/kick` logo easter egg (animated Firefox WebM on hover) to the whole flare-themed side of firefox.com, behind a waffle switch so we can flip it on/off without a deploy. The `/landing/kick` and `/landing/sidekick` pages keep their own scoped implementation untouched.

Hover behaviour: 1s hover-intent before the animation starts, plays through to completion regardless of mouseout, then a ~10s cooldown before it can trigger again.

## Files affected
- `media/js/cms/easter-egg-logo.es6.js` — new. Self-contained: logo swap, hover-intent, cooldown, and all safety gates (kick/sidekick, enterprise theme, Safari/WebM alpha, `prefers-reduced-motion`).
- `media/static-bundles.json` — registers the `easter-egg-logo` JS bundle.
- `springfield/cms/templates/cms/base-flare.html` — includes the bundle in `site_js` behind `switch('easter-egg-logo')`. This template is the only load point; older non-CMS pages (extending `base-protocol.html`) and enterprise-themed pages are unaffected.

## Waffle switch to create
- **Name:** `EASTER_EGG_LOGO`
- **Where:** Django admin → Waffle → Switches → *Add switch* on each environment (stage, prod).
- **Default:** create as **off**, flip on when you want it live. In local dev the switch defaults to **on** if it doesn't exist (existing `switch()` helper behaviour).
- **To retire or swap in a future easter egg:** edit the WebM URL and any sizing tweaks in `easter-egg-logo.es6.js`. Flip the switch off to disable entirely.

## Expected behavior

**Accessibility — `prefers-reduced-motion`**
When the user has requested reduced motion at the OS level (macOS System Settings → Accessibility → Display → Reduce motion; Windows Settings → Accessibility → Visual effects → Animation effects off), the swap is skipped entirely. The user sees the normal static SVG wordmark and no video ever loads or plays. This applies regardless of whether the switch is on.

**Safari and WebM alpha**
Safari (all versions), iOS Safari, and touch-enabled Macs are all detected via UA and touch-point sniffing, and the swap is skipped. Reason: Safari's support for WebM with an alpha channel is unreliable — historically it either failed to decode or rendered the transparent regions as a black background. On those browsers users see the normal static SVG wordmark; no video element is ever inserted.

**Other early-bail conditions**
- Kick/sidekick pages (`body.flare26-kick-page`) — their scoped implementation owns the logo.
- Enterprise-themed pages (`body.fl-theme-enterprise`) — their wider 200x40 wordmark doesn't fit our sizing tweaks.

In all early-bail cases the normal SVG wordmark stays in place; nothing regresses.

## Test plan
- [ ] Create switch `EASTER_EGG_LOGO`, enable it — hover the logo on `/en-US/`, `/firefox/new/`, or any other flare page. Video plays after 1s hover, completes fully, cooldown gates the next trigger.
- [ ] Disable the switch — normal SVG wordmark restored across the site.
- [ ] `/landing/kick` and `/landing/sidekick` still show their existing (immediate hover-play) easter egg regardless of switch state.
- [ ] Enterprise-themed CMS pages (`fl-theme-enterprise` body class) do not swap the logo.
- [ ] Safari (desktop + iOS + touch-Mac) shows the static SVG wordmark — no video attempted (check Network tab: no `.webm` request).
- [ ] With OS-level `prefers-reduced-motion: reduce`, static SVG is shown, no swap (check Elements panel: original `` still in place).
- [ ] Non-CMS pages extending `base-protocol.html` (older channel/marketing pages) do not load the bundle at all — confirm in DevTools Network tab.

### DIFF
--- a/media/js/cms/components/flare-animations.es6.js
+++ b/media/js/cms/components/flare-animations.es6.js
@@ -99,7 +99,10 @@ function initAnimationPauseButtons() {
 
         button.addEventListener('click', function () {
             if (video.paused) {
-                video.play().catch();
+                video.play().catch((error) => {
+                    if (error && error.name === 'AbortError') return;
+                    throw error;
+                });
                 button.setAttribute('aria-label', button.dataset.labelPause);
                 button.classList.remove('is-paused');
                 pauseIcon.hidden = false;
--- a/media/js/cms/components/flare-sliding-carousel.es6.js
+++ b/media/js/cms/components/flare-sliding-carousel.es6.js
@@ -83,7 +83,12 @@ class SlidingCarousel {
     }
 
     playVideo(videoEl) {
-        videoEl.play();
+        // play() rejects with AbortError when a slide change pauses the video
+        // before the play promise resolves; that's expected here.
+        videoEl.play().catch((error) => {
+            if (error && error.name === 'AbortError') return;
+            throw error;
+        });
         videoEl.autoplay = true;
 
         const btn = this.getAnimationButton(videoEl);
--- a/media/js/cms/components/flare-video.es6.js
+++ b/media/js/cms/components/flare-video.es6.js
@@ -26,18 +26,24 @@ function honorReducedMotionForAutoplay() {
     if (!window.matchMedia('(prefers-reduced-motion: reduce)').matches) return;
 
     document.querySelectorAll('video[autoplay]').forEach(function (video) {
-        // Pause immediately — more reliable than removing the autoplay attribute,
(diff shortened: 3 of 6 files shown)


knowler (approved):
r+ ⚽️

Works! I like this behaviour more than on the whatsnew, kick pages, since the user can’t interrupt it.

PR #1601, opened 2026-07-14
Create Navigation Snippet

maribedran:
## One-line summary

Create the Navigation Snippet replicating the structure from the existing header navigation, allow pages to choose a custom navigation, and enable the default site's navigation to be built on the CMS.

- Create the `NavigationSnippet` with the corresponding CMS blocks and components, replicating the existing structure from the `flare-header` component
  - The snippet introduces a new Top Level link, not available on the existing nav. The nav JS was adapted not to try to open a panel for those links.
- Add a new `custom_navigation` field to all pages to allow any page to override the default site header
  - The custom navigation is propagated to the page's children
  - The snippet needs to be published on the page's location, and the fallback for the alias location logic is respected
- When no custom navigation is set, the header falls back to any published Navigation Snippet with the `is_default` flag on the active locale. When not available, it falls back to the existing static version.
- The snippet also allows overriding the header logo and download button

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-437

## Testing

- Run `./manage.py load_page_fixtures` and check http://localhost:8000/cms-admin/snippets/cms/navigationsnippet/edit/1000/
  - Check the header on dark mode and mobile
- Edit any page, go to the Settings panel, and pick the custom nav
- Add the custom nav to any page that has children and check that the children also display the custom nav
- Go to any of the edited pages' translations and check that the custom nav doesn't show
  - Translate the nav snippet and check that the translated pages now show it
- Edit the snippet and check the "Is default" option
  - Check that CMS and static pages on the corresponding locale display it


springfield/cms/templates/components/nav-folder.html
```diff
@@ -0,0 +1,31 @@
...
+  <span
+    class="fl-menu-title"
+    aria-haspopup="true"
+    aria-controls="{{ panel_id }}"
+  >
+    <span class="fl-menu-heading">
+      {{ label }}
+      <span class="fl-icon fl-icon-chevron-down"></span>
+    </span>
+  </span>
```
knowler: As of #1465 this structure has changed slightly.

Because the mobile and desktop views need different semantic structures (i.e. desktop needs buttons for the dropdowns and mobiles doesn’t), we’re now using JS to apply and unapply the semantics for each scenario. I’ve used an `href`-less `<a>` tag here instead in the updated version. Part of the reason for that is because the “Features” one is a link. The other reason for that is because button semantics are easier to apply to an `<a>` element than link semantics are to apply to a `<button>` element.

If we can make this an `<a>` element and conditionally add the `href` that would be great, but it will work without the `href` in any case. The JavaScript is just expecting an `<a>` element here.

```suggestion
  <a class="fl-menu-title">
    <span class="fl-menu-heading">
      {{ label }}
      <span class="fl-icon fl-icon-chevron-down"></span>
    </span>
  </a>
```

springfield/cms/templates/components/nav-folder.html
```diff
@@ -0,0 +1,31 @@
...
+  <div class="fl-menu-panel" id="{{ panel_id }}">
+    <button class="fl-menu-close-button" type="button">
+      <span class="fl-icon fl-icon-close fl-menu-close-button-content">{{ ftl('navigation-close-menu') }}</span>
+    </button>
```
knowler: As of #1465, I don’t include the close button as its redundant (e.g. keyboard users can close the menu with the <kbd>Escape</kbd> key or simply move their focus outside of the menu; pointer users can light dismiss the menu by clicking outside).


```suggestion
```

media/js/cms/flare-navigation.es6.js
knowler, line 1: I imagine the change here can be omitted once you rebase as #1465 refactored pretty much everything to do with the dropdown menus.

knowler (commented):
👋🏻 Just noticed this PR is related to some changes I made to the nav, so I’ve addressed them. Super small changes, but without them the nav will break.

maribedran:
@knowler thanks! I've updated everything here: [6616269](https://github.com/mozmeao/springfield/pull/1601/commits/661626919f4d9e30e4b4fff725f084b836ef90e2)

springfield/cms/templates/components/nav-folder.html
knowler: I see this was partially done in [6616269](https://github.com/mozmeao/springfield/pull/1601/commits/661626919f4d9e30e4b4fff725f084b836ef90e2). `aria-controls` and `aria-haspopup` should also be removed.

- `aria-controls` only makes sense on the client when the element has the button semantics in the desktop view. Semantics that require JS to handle their expected behaviour should be applied with JS and not be serialized by the server.
- `aria-haspopup` should not be used at all in this context.

maribedran: Thanks for catching this! I'll remove.

springfield/cms/models/snippets.py
```diff
@@ -468,6 +478,100 @@ def get_preview_template(self, request, mode_name):
...
+        verbose_name_plural = "Navigation"
+
+    def __str__(self):
+        return f"{self.name} – {self.locale}"
+
+    @classmethod
+    def get_default(cls):
+        """Return the site default navigation, localized to the active locale, or None."""
+        snippet = cls.objects.filter(is_default=True, locale=SpringfieldLocale.get_default()).live().order_by("-last_published_at").first()
+        return snippet.get_localized() if snippet else None
```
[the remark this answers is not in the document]
maribedran: The snippet will always be created in en-US and then translated.

springfield/cms/migrations/0116_navigationsnippet.py
stevejalim, line 1: We already have a 115 on `main` - so we'll either need a merge migration in this branch or a renumbering

stevejalim: Have added to copilot's instructions to see it can catch this 🤞

springfield/cms/models/base.py
```diff
@@ -232,3 +250,25 @@ def get_breadcrumb_ancestors(self):
...
+        return context
+
+    def get_navigation(self):
+        """Resolve this page's custom navigation: its own, else the nearest
+        ancestor's, walking up the tree; each candidate must resolve in the
+        active locale. Returns None when neither the page nor any ancestor has
+        one — the site default navigation is supplied separately by the header
+        template via the get_default_navigation() tag.
+        """
+        candidate_pages = [self, *self.get_ancestors().specific().order_by("-depth")]
```
stevejalim: Do we also want `.live()` in here?

springfield/cms/models/snippets.py
```diff
@@ -468,6 +478,100 @@ def get_preview_template(self, request, mode_name):
...
+        blank=True,
+        use_json_field=True,
+    )
+    logo = models.ForeignKey(
+        "cms.SpringfieldImage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        help_text="Override the header logo. Falls back to the default Firefox logo if unset.",
```
stevejalim, line 504: What size rendition is needed for this `logo` and `logo_dark`? Is there a risk it will get loaded in a size that isn't [one of the specs we pre-generate](https://mozmeao.github.io/platform-docs/cms/images/#allowed-filter-specs)? If so, that can cause a 500 in production

stevejalim: Looking at the constants below, we don't pre-generate `480x160` so we might have a problem, unless however the logo is loaded (eg CSS) uses the `original` sizing, which of course will be there

stevejalim (approved):
Backend eyes - tentative r+wc as long as my comments/queries are addressed:

1) Merge migration needed
2) Logo size/rendition version could bite us given we have to pre-generate the renditions at save time 
3) need for a `.live()` when finding the next-best-fit nav option?

springfield/cms/models/snippets.py
maribedran: The template is rendering `width-400`, so we're covered on the renditions. I set those values arbitrarily as 4x the size of our existing logo, just in case someone wanted a bigger resolution. I forgot to go back to the model and adjust those numbers. Though maybe 480 -> 400 is close enough. @kkellydesign, what do you think?

springfield/cms/migrations/0116_navigationsnippet.py
maribedran: Thanks! I'll bump that number. I don't think we need a Copilot instruction since tests will fail when we merge with main and get migration conflicts.

springfield/cms/models/base.py
maribedran: Good catch! Thanks.

PR #1602, opened 2026-07-14
WT-1438 Add enterprise download component

StuartMacKay:
## One-line summary

Create a component for the download section on the Enterprise Firefox page so the page can be migrated to the Wagtail CMS in the future,

The component is a verbatim copy of the existing HTML and template language constructs. 

1. The existing page includes the template extracted for the component and should display identically to the previous version.
2. media/js/firefox/enterprise/landing.es6.js was updated to ensure the menus were initialized correctly. Since this file is 'global' the change should be reviewed carefully.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1438

## Testing

The download component currently exists in isolation. However it can be added to a "Free Form 2026 Page" to confirm it renders identically to the section at the bottom of the https://www.firefox.com/en-US/browsers/enterprise/

It's also worth noting that during development the download section was replaced by the component and a visual regression test generated to confirm the component rendered identically to the static content it replaced. These are available as a patch, if required.

### DIFF
--- a/media/js/firefox/enterprise/landing.es6.js
+++ b/media/js/firefox/enterprise/landing.es6.js
@@ -7,4 +7,18 @@
 import MzpDetails from '@mozilla-protocol/core/protocol/js/details';
 
 window.MzpDetails = MzpDetails;
-window.MzpDetails.init('.mzp-c-menu-list-title');
+
+function initMenuLists() {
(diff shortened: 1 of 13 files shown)


kkellydesign (approved):
E2E tests work well for me. I'll leave a quick code review to the others. Thanks for the speedy work here, @StuartMacKay.

PR #1603, opened 2026-07-14
Ship springfield sqlite DB to GCS instead of S3

stevejalim:
This changeset moves the upload/download behaviour for sqlite DB exports to use Google Cloud Storage, not AWS S3.

The rest of springfield has been on GCP for a while (CMS media uses `django-storages[google]`), so `google-cloud-storage` is already installed transitively; `boto3` is dropped from `prod.in` since `run-db-upload.py` was its only consumer.

Mirrors mozilla/bedrock#17286.

## Summary

- Rename `bin/db_s3_utils.py` -> `bin/db_utils.py` (no longer S3-specific).
- `bin/run-db-upload.py`: swap boto3 for `google.cloud.storage`; ADC-authenticated `storage.Client()`.
- `bin/run-db-download.py` / `springfield/base/views.py`: switch base URL from `s3-<region>.amazonaws.com` to `storage.googleapis.com`.
- Bucket name still comes from `AWS_DB_S3_BUCKET` (kept unchanged so k8s ConfigMap can carry over).
- Drop `boto3` from `requirements/prod.in` + recompile.

## Pre-cutover checklist (per env: dev / stage / prod)

1. Create the GCS bucket. Using the same name as the S3 bucket is simplest so `AWS_DB_S3_BUCKET` in the k8s ConfigMap can carry over unchanged.
2. Access-control mode. Enable uniform bucket-level access + grant `allUsers` the Storage Object Viewer role.
3. Copy the two live objects from S3 -> GCS. Historical `.db` files are unreferenced and can be skipped:
   - `springfield_db_info.json` (the metadata pointer)
   - the `.db` named by `file_name` inside that JSON
4. Grant the clock pod's GSA `storage.objectAdmin` (or `objectCreator` + `objectViewer`) on the bucket so `storage.Client()` via ADC can upload.
5. Sanity check: `curl https://storage.googleapis.com/<bucket>/springfield_db_info.json` should return the metadata before deploying the code change.

## Test plan

- [x] CI green (unit tests, lint, requirements pinning check)
- [ ] After bucket provisioning per checklist above, deploy to dev and confirm the clock pod uploads the sqlite DB
- [ ] Confirm a pod fetching the DB via `run-db-download.py` gets the object from `storage.googleapis.com` and passes the checksum check
- [ ] `/healthz/cron/` shows current `db_file_url` pointing at the new GCS host


janriokrause (approved):
r+

PR #1605, opened 2026-07-14
add France to allowed territories for smart window enable

slightlyoffbeat:
## One-line summary

This will enable users in France to see the enable smart window experience on /smart-window/

The page is not yet localized, but that is known and understood by the team.


janriokrause (approved):
r+ 🇫🇷

PR #1612, opened 2026-07-14
Rename AWS_DB_JSON_DATA_FILE env var to GCS_DB_JSON_DATA_FILE

stevejalim:
## Summary

- Follow-up to #1603 (S3 → GCS migration): the env var name still referenced AWS. Rename to `GCS_DB_JSON_DATA_FILE` to match the actual storage backend.
- Only two references in the repo: `bin/db_utils.py` and `springfield/base/views.py`.

## Test plan

- [x] Confirm any deployment configs / secrets referencing `AWS_DB_JSON_DATA_FILE` are updated to `GCS_DB_JSON_DATA_FILE` before/with this rollout (the fallback default keeps local dev working without the env var set).

### DIFF
--- a/bin/db_utils.py
+++ b/bin/db_utils.py
@@ -11,7 +11,7 @@ from subprocess import CalledProcessError, check_output
 
 JSON_DATA_FILE_NAME = "springfield_db_info.json"
 DATA_PATH = getenv("DATA_PATH", "data")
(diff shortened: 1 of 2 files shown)


stevejalim (commented):
self review: r+

PR #1620, opened 2026-07-16
Create Bind to UI Tour conditional display option

maribedran:
## One-line summary

Allow a conditional display block to show when it contains a visible UI Tour button.

The UI Tour buttons are only displayed by JS logic that checks if their actions are available for the user. This new conditional display option makes the block visible when it contains a visible UI Tour.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1465

## Testing

- Run `./manage.py load_page_fixtures` and open http://localhost:8000/en-US/flare-docs/blocks/test-conditional-display/
- The "Bound to UI Tour" section should appear on Firefox if UI Tour is correctly set up
- The section should not appear in other browsers or if UI Tour is not set up
image


kkellydesign (approved):
works 👍

PR #1625, opened 2026-07-17
Easteregg fix

slightlyoffbeat:
## Fix easter egg logo on light-theme pages

The easter egg logo WebM is the white wordmark, so it only shows up on a dark nav. On light-theme pages (e.g. `/features/protection`) it rendered white-on-white and disappeared.

### Changes
- **Only run on a dark nav.** Activate the video only when the theme is showing the white wordmark, so it never lands on a white background.
- **React to live theme changes.** Swap the logo in/out when the OS color scheme flips at runtime (devtools, macOS auto dark-at-sunset) instead of needing a refresh.
- **Skip custom/brand logos.** Custom navigation snippets reuse the `fl-logo-fx` class; leave those alone so the Firefox flame never replaces a partner logo.
- **Clean up `play()` error handling.** Reset state instead of rethrowing, avoiding an uncatchable unhandled rejection and a stuck-state bug.

### Testing
- Verify the easter egg still plays on hover on dark-nav pages (e.g. homepage).
- Confirm it no longer appears on light-theme pages.
- Toggle light/dark in devtools and confirm the logo updates without a refresh.
- Confirm custom-logo pages are unaffected.

### DIFF
--- a/media/js/cms/easter-egg-logo.es6.js
+++ b/media/js/cms/easter-egg-logo.es6.js
@@ -58,10 +58,10 @@ const setupEasterEggLogo = () => {
     const activate = () => {
         if (teardown) return;
 
-        const existingLogo = logoLink.querySelector('img');
-        if (existingLogo) existingLogo.remove();
-
-        // Hide the wordmark that .fl-logo-fx paints as a background.
+        // Hide the wordmark that .fl-logo-fx paints as a background. The static
+        // <img> logo is left in the DOM and hidden via CSS
+        // (.fl-logo-fx:has(.fl-video-ready) img) so teardown restores it for
+        // free — no need to remove and re-insert the node here.
         logoLink.style.backgroundImage = 'none';
         // Widen the logo container so the WebM's flame lands at the same visual
         // size as the SVG wordmark, and shift it back so the layout doesn't move.
@@ -94,6 +94,10 @@ const setupEasterEggLogo = () => {
         video.addEventListener(
             'canplaythrough',
             () => {
+                // Signal to CSS that the video is showing its first frame, so
+                // the static wordmark <img> can be hidden.
+                wrapper.classList.add('fl-video-ready');
+
                 video.addEventListener('mouseover', () => {
                     if (isPlaying || hoverTimer) return;
                     if (Date.now() < cooldownUntil) return;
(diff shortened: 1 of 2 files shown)


janriokrause (approved):
Really nice fix. The `--fl-logo-url` signal is a clever way to detect the dark nav, and the `play()` cleanup is a great catch for preventing expected `AbortError`s from spamming Sentry. 🎉

One tiny thing (non-blocking): in `activate()` we remove the `` from the default flare header, but `teardown()` only restores the CSS background and sizing. So after a dark→light flip the wordmark looks identical, but the original DOM node is never restored. This is fine with the current CSS, though it could become noticeable if the logo asset or background-based rendering changes later. Re-inserting the same `` during teardown would make it a more complete inverse of `activate()`, but it's still totally fine to ship as-is. 😊

media/js/cms/easter-egg-logo.es6.js
```diff
@@ -33,74 +33,131 @@ const setupEasterEggLogo = () => {
...
+            .includes('logo-word-hor-white');
+
+    // Set by activate(); calling it removes the video and restores the logo.
+    let teardown = null;
+
+    const activate = () => {
+        if (teardown) return;
+
+        const existingLogo = logoLink.querySelector('img');
+        if (existingLogo) existingLogo.remove();
```
knowler: To address what @janriokrause is saying, but in a slightly different way: we don’t need to remove this. Instead, we can handle this with CSS.

1. As a prerequisite, we can add a `fl-video-ready` class to the video wrapper after `canplaythrough` fires (no poster is needed—the video will show the first frame as its default poster before it even gets to this point).
2. Use `:has(.fl-video-ready)` to conditionally hide the existing logo `` element with CSS.

slightlyoffbeat: Done — went with your CSS approach. `activate()` now leaves the `` in the DOM and adds `.fl-video-ready` to the wrapper on `canplaythrough`; `flare-navigation.css` hides it via `.fl-logo-fx:has(.fl-video-ready) img`. Removing the video on teardown reverses it automatically, so this also covers @janriokrause's point about restoring the original node. Thanks both! 🙏

knowler (approved):
r+ 🍳

PR #1628, opened 2026-07-20
WT-1484: Groundwork for FF refer project

stevejalim:
This changeset lands some groundwork for the FF Referral project. The aim is to boostrap shared surfaces to enable us to work in paralle more easily

Commits:

- 5947f4eb: Add a stable local-dev REFERRAL_INVITE_CODE_KEY_V1 signing key to .env-dist. Distinct from www-dev / stage / prod keys, so dev and CI can share a deterministic value without cross-env leakage. Please copy the value to your .env
- 8ef988e4: Introduce `FirefoxReferralData` model (siloed under springfield/firefox/referral/models.py). This is a Django model mapping `referral_id` to `install_count` with a `last_refreshed_at` timestamp and a `refresh()` manager stub, wiring the ORM entry point for the upcoming Data Eng sync (they hand us (`invite_code, install_count`), we decrypt to `referral_id`. No PII or FxA linkage anywhere in the pipeline).
- d92d8632: Add skeleton `ReferralHubPage` (`/invite/`) and `ReferralGetFirefoxPage` (`/get-firefox/`) as `HomePage` children in the `cms` app. Templates just show the Jinja context raw so BE and FE work can proceed in parallel without stepping on each other. Adds both to the `_allowed_page_models` allowlist and the DB export script.
- 0a400fd5: Override `ReferralHubPage.get_context()` to inject a placeholder `invite_url` into the context

JIRA: https://mozilla-hub.atlassian.net/browse/WT-1484 (parent https://mozilla-hub.atlassian.net/browse/WT-1281)


stephaniehobson (approved):
R+ 

Looks good as empty files to set up a framework.

PR #1630, opened 2026-07-20
Set `type=button` on dialog close button

knowler:
## One-line summary

Sets `type=button` on close button in dialog component so that the `<dialog>` can be nested within `<form>`, since otherwise the close button submits the form.

In a modal state, a dialog element becomes the only active content in a page (i.e. all other content is inert and not a part of the accessibility tree). This means it’s totally fine to nest a `<dialog>` element wherever in a document, including within forms. This pattern can be helpful since it allows one to colocate the `<dialog>` element with its trigger and other relevant content (e.g. imagine creating dialog and dialog triggers in a loop).

## Issue / Bugzilla link

- N/A

## Testing

- Confirm an existing dialog still works.


janriokrause (approved):
r+

PR #1635, opened 2026-07-21
Add overwritable title and content fields for QR codes

bluewave41:
## One-line summary
Adds an overwritable title and content field to QR codes

## Issue / Bugzilla link
https://mozilla-hub.atlassian.net/jira/for-you?tab=assigned&selectedIssue=WT-761

## Testing
- [ ] Checkout this PR
- [ ] http://localhost:8000/cms-admin/pages/321/edit/
- [ ] Toggle the QR code and set a custom title and content
- [ ] Open http://localhost:8000/en-US/download/ios/ and ensure the custom content displays


maribedran (commented):
@bluewave41, if we're overriding all the snippet fields at the page level, wouldn't it make more sense to have an FK to the snippet so editors could pick a different one, or to build the floating QR entirely on the page?
I'm worried about this getting too confusing. The snippet is built separately, and the page only had options to change its state. We later added the option to override the image. Now we have options to override the entire snippet, which makes the original snippet useless. Ideally, I'd like us to determine how this snippet is being used and think of a solution that better addresses editor needs and usability.

PR #1636, opened 2026-07-21
Enterprise revisions

maribedran:
## One-line summary

CMS and CSS revisions for the Enterprise pages

- Only validate the Contact Page's slug when publishing the page
  - The current draft page on production doesn't allow the user to see the preview because of the validation error caused by the slug. That slug will only be fixed once the pages are ready to be published. Since we only need the live page to respect the slug rules, the check was moved to the publish action instead of running on every call to `.clean()`
- Add the `PageThemeMixin` to the Contact Page
- Create fixtures for the Enterprise pages and their custom Navigation Snippet
- Fix layout issues with the banner component

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1485

## Testing

Run `./manage.py load_page_fixtures` and check the pages:
- http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/
- http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/product/
- http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/support/
- http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/contact/
- http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/download/


springfield/cms/models/pages.py
```diff
@@ -2017,9 +2018,36 @@ def __str__(self):
...
+        # `action-publish` is present in the POST data when the editor clicks
+        # "Publish". Draft saves and "Submit for moderation" omit it, so they
+        # skip this check.
+        is_publishing = "action-publish" in self.data
+        slug = cleaned_data.get("slug")
+        if is_publishing and slug and settings.PROD:
+            parent = self.parent_page or self.instance.get_parent()
+            path = parent.url_path + slug + "/" if parent else "/" + slug + "/"
+            # Using .search() instead of .match() because paths will often start with /home/parent/child/
+            if not any(re.search(allowed_path, path) for allowed_path in settings.CONTACT_PAGE_ALLOWED_PATHS):
```
stevejalim, line 2040: @maribedran I still need to make the WAF change to allow the wildcard path on landing/*/get/ -- are you expecting that to be in production already?

stevejalim (approved):
Backend r+wc

springfield/cms/models/pages.py
maribedran: We'll need `/enterprise/contact/` for production for the Enterprise launch

PR #1638, opened 2026-07-22
chore(Plausible): Add a new additive, env-driven PLAUSIBLE_EXTRA_COUNTRIES setting

stephendherrera:
## One-line summary

Enable Plausible analytics for Brazil alongside GA, in order to test analytics parity.

- New `PLAUSIBLE_EXTRA_COUNTRIES` setting (`springfield/settings/base.py`): additive, env-driven, defaults to `BR`.
- `plausible_enabled()` (`springfield/firefox/templatetags/misc.py`) now loads Plausible for consent countries **OR** any country in `PLAUSIBLE_EXTRA_COUNTRIES`. The existing consent-country path is unchanged.

This deliberately does **not** add `BR` to `DATA_CONSENT_COUNTRIES`. That list is shared with the cookie-consent banner and Google Consent Mode (`data-needs-consent` -> `consentRequired()` in `gtm-snippet.es6.js`). Adding `BR` there would flip Brazil into consent-required mode and stop GA4 from loading by default, which would degrade GA data and break the comparison. Keeping `BR` out of the consent list and only in the Plausible extra-list means GA4 keeps loading by default in Brazil while Plausible now loads too.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1510

## Testing

- `test_plausible_enabled`: added `BR` cases (enabled with switch + domain; disabled when switch off or domain empty).
- `test_needs_data_consent`: added `BR -> False` to confirm Brazil is not a consent country (guards GA behavior).

### DIFF
--- a/springfield/settings/base.py
+++ b/springfield/settings/base.py
@@ -1186,10 +1186,25 @@ DATA_CONSENT_COUNTRIES = [
 ]
 
 # Extra countries (beyond DATA_CONSENT_COUNTRIES) where the Plausible
-# analytics script should load. This is intentionally env-driven: set the
-# PLAUSIBLE_EXTRA_COUNTRIES env var (comma-separated ISO codes, e.g. "BR")
-# to enable or disable per environment.
-PLAUSIBLE_EXTRA_COUNTRIES = config("PLAUSIBLE_EXTRA_COUNTRIES", default="", parser=ListOf(str))
+# analytics script should load. This is intentionally env-driven so we can
+# turn a country on or off per environment (dev/stage/prod) without a code
+# change or deploy, and so it can be reverted instantly if needed. Additionally,
+# this is likely a temporary measurement addition.
+# Set the PLAUSIBLE_EXTRA_COUNTRIES env var to a comma-separated list of
+# ISO country codes, e.g.: PLAUSIBLE_EXTRA_COUNTRIES=BR,CA
+# Empty by default (no extra countries). filter(None, ...) drops empty tokens
+# (e.g. a trailing comma) so the list never contains an empty string that
+# could match a blank country code.
+PLAUSIBLE_EXTRA_COUNTRIES = list(
+    filter(
+        None,
+        config(
+            "PLAUSIBLE_EXTRA_COUNTRIES",
+            default="",
+            parser=ListOf(str, allow_empty=True),
+        ),
+    )
+)
 
 
 # RELAY =========================================================================================
--- a/springfield/settings/base.py
+++ b/springfield/settings/base.py
@@ -1185,6 +1185,12 @@ DATA_CONSENT_COUNTRIES = [
     "GB",  # United Kingdom
 ]
 
+# Extra countries (beyond DATA_CONSENT_COUNTRIES) where the Plausible
+# analytics script should load. This is intentionally env-driven: set the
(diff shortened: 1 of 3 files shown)


springfield/settings/base.py
```diff
@@ -1185,6 +1185,23 @@ def before_send(event, hint):
     "GB",  # United Kingdom
 ]
 
+# Extra countries (beyond DATA_CONSENT_COUNTRIES) where the Plausible
+# analytics script should load. This is intentionally env-driven: set the
+# PLAUSIBLE_EXTRA_COUNTRIES env var (comma-separated ISO codes, e.g. "BR")
```
stephaniehobson: @stephendherrera Can you tell me more about why this is ENV driven? And update the comment with an example of an entire line, like `PLAUSIBLE_EXTRA_COUNTRIES=BR,CA` or whatever is appropriate if that's wrong.

stephendherrera: @stephaniehobson Main reason is I'm trying to decouple this from an app release. Since it's likely a temporary change to evaluate GA4 and Plausible parity, and may need a few variant tests along the way, I wanted to keep it in config rather than hardcode it. That said, if this is an anti-pattern for Springfield, just let me know.

springfield/settings/base.py
```diff
@@ -1185,6 +1185,27 @@ def before_send(event, hint):
...
+# analytics script should load. This is intentionally env-driven so we can
+# turn a country on or off per environment (dev/stage/prod) without a code
+# change or deploy, and so it can be reverted instantly if needed. Additionally,
+# this is likely a temporary measurement addition.
+# Set the PLAUSIBLE_EXTRA_COUNTRIES env var to a comma-separated list of
+# ISO country codes, e.g.: PLAUSIBLE_EXTRA_COUNTRIES=BR,CA
+# Empty by default (no extra countries). filter(None, ...) drops empty tokens
+# (e.g. a trailing comma) so the list never contains an empty string that
+# could match a blank country code.
+PLAUSIBLE_EXTRA_COUNTRIES = list(
```
stephaniehobson: This feels unnecessarily complicated, is it possible to simplify?

stephendherrera: We can just omit? @stephaniehobson

stephaniehobson: Some kind of data validation/cohersion is probably needed. Give me a minute and I'll suggest something.

springfield/settings/base.py
stephaniehobson: Nah, it's fine. It seemed like a lot of work to turn BR on for ever but if changes are expected that's okay.

springfield/settings/base.py
```diff
@@ -1185,6 +1185,27 @@ def before_send(event, hint):
...
+    filter(
+        None,
+        config(
+            "PLAUSIBLE_EXTRA_COUNTRIES",
+            default="",
+            parser=ListOf(str, allow_empty=True),
+        ),
+    )
+)
```
stephaniehobson:
```suggestion
PLAUSIBLE_EXTRA_COUNTRIES = {
    c.strip() 
    for c in config("PLAUSIBLE_EXTRA_COUNTRIES", default="").split(",") 
    if c.strip()
}
```

stephendherrera: @stephaniehobson - works for me. applied.

stephaniehobson (approved):
Thanks for the changes R+

PR #1648, opened 2026-07-24
Add FF1 encryption for referral invite-code

janriokrause:
# Add FF1 encryption for referral invite-code and reverse mapping

## What

Adds a reversible, deterministic, format-preserving mapping between a Firefox referral ID (15-char Crockford base32) and a public 16-char invite code, per the referral-invite-code spec. The shared code reveals nothing about the referrer ID, and the website can reverse a code back to the referrer ID to attribute installs, with no database writes on the request path.

## How

* `springfield/firefox/referral/ff1.py` vendors NIST FF1 (SP 800-38G section 5.1) using AES from `cryptography`. `b`/`d` use integer-only bit-length math (exact for the power-of-two radix 32). Every logical block cites its spec step.
* `springfield/firefox/referral/crypto.py` is the thin adapter: `referral_id_to_invite_code`, `invite_code_to_referral_id`, `invite_url_for_code`. The FF1 tweak is empty bytes (the key is single-purpose, so no domain separation is needed). Decode `ValueError`s report to Sentry at warning level, fingerprinted and categorized `bad_format` / `version_not_in_keyring`, and never log the code, referral ID, or key.
* `springfield/firefox/referral/utils.py` holds Crockford validation, normalization, and the startup keyring validator.
* Settings gain a versioned keyring (`REFERRAL_INVITE_CODE_KEYS` + `REFERRAL_INVITE_CODE_ACTIVE_KEY_VERSION`) with boot-time validation that raises `ImproperlyConfigured` on an empty keyring, an active version not in the keyring, a key that is not 32 bytes, or a version identifier that is not a single Crockford char.
* `ReferralHubPage` now builds the shareable URL from the real crypto instead of the placeholder.

## Reviewer materials

* Algorithm: NIST SP 800-38G section 5.1 (Algorithms 7 and 8), https://csrc.nist.gov/pubs/sp/800/38/g/upd1/final
* Test vectors: NIST FF1 sample values (FF1samples.pdf), https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/FF1samples.pdf
* Primary reference (ported from): Capital One `fpe` (Go), tag `v1.2.1`, commit `80b2f950e5923895c32d589c2df0390e80fba22f`, https://github.com/capitalone/fpe/blob/v1.2.1/ff1/ff1.go
* Cross-check reference: str4d `fpe` (Rust), main branch, commit `913e533c9345d5c8e11aac42318fc5743d6f58e9`, `src/ff1.rs`, https://github.com/str4d/fpe/blob/913e533c9345d5c8e11aac42318fc5743d6f58e9/src/ff1.rs
* Design note: Crockford base32 alphabet (radix 32, uppercase, no `I`/`L`/`O`/`U`), empty FF1 tweak (single-purpose key), 256-bit AES keys in a versioned keyring, `ValueError` on bad format or unknown key version, and no integrity check on decode by design (the business layer rejects unattributable IDs).

## Test evidence

* All 9 NIST FF1 vectors pass encrypt and decrypt.
* 20 additional vectors generated from Capital One `fpe` v1.2.1 pass, covering radix 10/26/32/36, AES-128/192/256, empty and non-empty tweaks, unequal Feistel halves, and a long input exercising the S-expansion (`d > 16`) path.
* Round-trip and bijection property tests over the Crockford alphabet.
* Wrapper tests: malformed-input rejection, case-insensitivity, whitespace tolerance, separator rejection, unknown-version rejection, active-version switch, and the Sentry observability paths.
* Per-version regression fixtures, frozen.
* Startup validation tests for all four `ImproperlyConfigured` failure modes.
* Local run: 147 referral and CMS referral-page tests pass. `ruff check`/`format` and `manage.py check` are clean.

## Notes and follow-ups

* Referral-ID length is 15 chars, producing 16-char invite codes. **Deferred DB migration (required follow-up):** `FirefoxReferralData.referral_id` is still `max_length=10` in `models.py` and must be raised to 15 with a new Django migration before real 15-char referral IDs can be stored (the DE import path), and the `bootstrap_dummy_referral_data` dummy IDs refreshed to 15 chars. The crypto already mints and decodes 15-char codes now. Only storage is gated on the migration.
* `invite_url_for_code` returns `https://www.firefox.com/get-firefox/?invitation={code}` to match the shipped hub wiring. This intentionally differs from the spec's Function-2 text (`/invite/?code=`), which is stale relative to the committed product URL.
* Citation correction: the spec text references SP 800-38G `section 6.2` and attributes the vectors to `Appendix A`, but the published standard defines FF1 in `section 5.1` (Algorithms 7 and 8) and the sample values live in `FF1samples.pdf` linked from `Appendix D`. The code and these reviewer materials cite the correct `section 5.1` / `FF1samples.pdf`, verified against the live NIST PDF. The spec's `6.2` / `Appendix A` are errors in the spec, not the code.
* `REFERRAL_INVITE_CODE_ACTIVE_KEY_VERSION` reads from config with a default of `1`, so a rotation's "flip active version" step can be a config-only env change. Set the real per-environment `REFERRAL_INVITE_CODE_KEY_V1` secret in each environment (the local/CI value is in `.env-dist`).
* Minor, intentional deviation from the reference implementations: `ff1.py` builds a fresh `cryptography` AES `Cipher` per PRF call and per S-block, whereas Capital One and str4d build the AES key schedule once and reuse it. This is a readability choice (obviously stateless), not correctness. Output is byte-identical (the NIST and Capital One vectors pass), and at our 15-char input size the cost is negligible. Switching to a reused cipher object is a safe optimization if ever needed.
* The optional timing-attack smoke test is deferred to a follow-up, per the spec.
* Runbook (per the spec's operational notes): watch Sentry for sustained spikes in the decode warnings. A `bad_format` spike suggests broken invitee UX or a scraper. A `version_not_in_keyring` spike suggests a keyring/rotation mismatch across pods.

## Context

* https://mozilla-hub.atlassian.net/browse/WT-1478 for spec


springfield/cms/tests/test_referral_pages.py
```diff
@@ -14,10 +15,16 @@ def test_hub_page_get_context_builds_invite_url_from_ref_key(rf):
...
 
-    # Placeholder algorithm reverses the ref_key and swaps a leading TSET for FAKE.
-    assert context["invite_url"] == "http://testserver/get-firefox/?invitation=FEDCBAFAKE"
+    # The hub encrypts the ref_key into an invite code and wraps it in the
+    # shareable download URL. Asserted against the crypto helpers rather than a
+    # hardcoded code so this tracks the wiring, not the cipher output (which is
+    # covered by the referral crypto tests).
+    expected_code = crypto.referral_id_to_invite_code("TESTABCDEFGHJKM")
+    assert context["invite_url"] == crypto.invite_url_for_code(expected_code)
+    assert context["invite_url"].startswith("https://www.firefox.com/get-firefox/?invitation=")
```
stevejalim: What do you think about skipping `startswith` and just doing a full comparison by concatenating the URL with `expected_code`?

springfield/cms/tests/test_referral_pages.py
```diff
@@ -14,10 +15,16 @@ def test_hub_page_get_context_builds_invite_url_from_ref_key(rf):
...
+    assert crypto.invite_code_to_referral_id(expected_code) == "TESTABCDEFGHJKM"
```
stevejalim: Is this reversal of the `invite` code back to a `ref_key` definitely needed here?

springfield/cms/tests/test_referral_pages.py
```diff
@@ -38,23 +45,12 @@ def test_hub_page_get_context_invite_url_empty_when_ref_key_blank(rf):
...
+def test_hub_page_get_context_invite_url_empty_when_ref_key_invalid(rf):
     site = Site.objects.get(is_default_site=True)
     hub_page = ReferralHubPageFactory(parent=site.root_page)
 
-    # Force the helper to return a value with characters that must be
-    # percent-encoded so we exercise the urlencode call without relying
-    # on the placeholder algorithm ever emitting them.
-    hub_page._referral_id_to_invite_code = lambda referral_id: "a b&c=d"
+    # A ref_key that is not a valid referral ID (here it has separators and is
+    # the wrong length) is treated like a missing one instead of raising.
```
stevejalim, line 57: Nice test

springfield/settings/base.py
```diff
@@ -546,6 +547,23 @@ def language_url_map_with_fallbacks():
...
+# Symmetric keyring for turning a Firefox referral ID into a public invite code
+# and back. Each key is a dedicated 256-bit AES key, hex-encoded (64 hex chars)
+# in its own per-environment secret. These keys must not be reused for cookies,
+# CSRF, sessions, or anything else. Add new versions to the dict as rotations
+# happen (a two-deploy operation: add the key first, flip the active version in
+# a later deploy). Version identifiers are single Crockford base32 characters.
+# The local/CI default below is a fixed non-production value (also in .env-dist)
+# so tests are deterministic. Staging and prod override it with distinct secrets.
+REFERRAL_INVITE_CODE_KEYS = {
+    "1": bytes.fromhex(config("REFERRAL_INVITE_CODE_KEY_V1", default="abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcd")),
```
stevejalim: When we have default keys, it's likely someone will raise a bug-bounty submission saying there's a risk of this publicly-disclosed key being used in production. In reality, that is not likely, but it is _possible_, so better to have no default and require a key in the .env at startup time.
```suggestion
    "1": bytes.fromhex(config("REFERRAL_INVITE_CODE_KEY_V1")),
```

The V1 keys (different for all envs) are already deployed to k8s, so there's no startup-failure risk there.

To avoid breaking out demo servers by dropping the default key here, we can add this to the bottom of `gcp/springfield-demos/cloudrun/cloudrun-demo.env.yaml`:

`REFERRAL_INVITE_CODE_KEYS: "abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcd"`

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+Everything here is a pure function with no I/O so it is safe to import from
+``settings.py`` (see :func:`validate_invite_code_keyring`, which runs at Django
+startup). It deliberately never logs, echoes, or embeds the referral ID or
+invite code in an exception message, because those are the plaintext and
+ciphertext of the FF1 mapping.
+"""
+
+from __future__ import annotations
+
```
stevejalim:
```suggestion

```

I don't think we need this for the kind of type hinting we're doing here, yeah?

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+# Crockford base32 alphabet, uppercase canonical: digits 0-9 then A-Z with
+# I, L, O and U removed (the visually ambiguous letters). Exactly 32 symbols,
+# so it maps one-to-one onto FF1 radix 32. The ORDER of this string is part of
+# the invite-code wire format: reordering it silently invalidates every
+# outstanding code, so treat it as frozen once shipped.
+CROCKFORD_ALPHABET = "0123456789ABCDEFGHJKMNPQRSTVWXYZ"
+CROCKFORD_SYMBOLS = frozenset(CROCKFORD_ALPHABET)
+
+# Assumed input format from Firefox: 15-char Crockford base32 referral IDs.
+REFERRAL_ID_LENGTH = 15
```
stevejalim: **blocking:** this is moving to 16 chars, which menas that invite code length moves to 17 as well
```suggestion
REFERRAL_ID_LENGTH = 16
```

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+# 1 version char + 15 ciphertext chars.
+INVITE_CODE_LENGTH = 16
```
stevejalim:
```suggestion
INVITE_CODE_LENGTH = 17
```

springfield/firefox/referral/utils.py
stevejalim: I won't mark it all up with suggestions, as there are lots of comments etc referencing the 15-char length

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+KEY_LENGTH_BYTES = 32
+
+
+def is_canonical_crockford(value: str) -> bool:
+    """True if every character of ``value`` is an uppercase Crockford symbol."""
+    return all(char in CROCKFORD_SYMBOLS for char in value)
+
+
+def validate_referral_id(referral_id: str) -> None:
+    """Validate a referral ID is 15-char, canonical (uppercase) Crockford base32.
```
stevejalim:
```suggestion
    """Validate a referral ID is 16-char, canonical (uppercase) Crockford base32.
```

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def validate_invite_code(invite_code: str) -> None:
+    """Validate a normalized invite code is 16-char Crockford base32.
+
+    Raises ``ValueError`` on wrong length or any non-Crockford character. This
+    catches typos and rejects separators (hyphens, dots, underscores) that
+    ``normalize_invite_code`` intentionally does not strip.
+    """
+    if len(invite_code) != INVITE_CODE_LENGTH:
```
stevejalim, line 92: For the referral code we validate it's a string - why do we not do that here too for the invite code?

springfield/firefox/referral/crypto.py
```diff
@@ -0,0 +1,109 @@
...
+Thin adapter over the vendored FF1 primitive (:mod:`ff1`). Callers use these
+three functions and never touch FF1 directly. The mapping is deterministic,
+reversible, and format-preserving. Different referral IDs can never collide to
+the same invite code because FF1 is a permutation.
+
+Nothing here does database access or logs secrets. The only thing ever logged
+is the single-character key version (safe to log), and only on a decode failure.
+"""
+
+from __future__ import annotations
```
stevejalim: Again I don't think we need this
```suggestion

```

springfield/firefox/referral/tests/test_crypto.py
```diff
@@ -0,0 +1,217 @@
...
+        "a7b9k2m4px",  # lowercase, not canonical
+        "A7B9K2M4PI",  # I not in Crockford alphabet
+    ],
+)
+def test_referral_id_to_invite_code_rejects_malformed(bad):
+    with pytest.raises(ValueError):
+        referral_id_to_invite_code(bad)
+
+
+@pytest.mark.parametrize("code", ["", "1", "1X3ZQ8R5M7NPQRS", "1X3ZQ8R5M7NPQRST9"])
```
stevejalim: This is fine but as a tip, it might be useful to add an `ids` list for these examples so it's easier to see what they're testing for could be neeat
```suggestion
@pytest.mark.parametrize("code", ["", "1", "1X3ZQ8R5M7NPQRS", "1X3ZQ8R5M7NPQRST9"], ids=["empty", "single digit", "too long", "also too long"])
```

springfield/firefox/referral/crypto.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def invite_url_for_code(invite_code: str) -> str:
+    """Build the shareable download URL for an invite code.
+
+    Kept as a function so the URL structure lives in exactly one place. The
+    Crockford alphabet is URL-safe, so no percent-encoding is needed. Does not
+    validate the code.
+    """
+    return f"https://www.firefox.com/get-firefox/?invitation={invite_code}"
```
stevejalim: Suggestion: we raise if invite_code is invalid - or are we 100% sure that'll be checked before we call this helper?

springfield/firefox/referral/tests/test_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    assert normalize_invite_code("  1x3zq8r5m7n  ") == "1X3ZQ8R5M7N"
+    assert normalize_invite_code("1x3zq8 r5m7n") == "1X3ZQ8R5M7N"
+
+
+def test_normalize_invite_code_rejects_non_string():
+    with pytest.raises(ValueError):
+        normalize_invite_code(None)
+
+
+def test_validate_invite_code_accepts_16_char_crockford():
```
stevejalim: I know you know this but this'll need ugprading to 17 chars, along with related code

stevejalim (commented):
Posting reviews of everyting except ff1.py to help you keep moving.

Overall looking good - the biggest change will be to 16-char ref_key and 17-chat invite code.

More comments to follow when I look at ff1.py, and i'll also unleash Copilot and Claude

springfield/firefox/referral/crypto.py
```diff
@@ -0,0 +1,109 @@
...
+
```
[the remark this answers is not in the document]
stevejalim: Yeah I wondered about this, too - will mean we can run on Staging cleanly if we need to test stuff, etc

stevejalim (commented):
CI is green. FF1 port is correct against NIST SP 800-38G and independently verified vectors. A few things worth addressing before merge - see inline comments. The biggest one is the default key concern on settings/base.py.

springfield/settings/base.py
stevejalim: The dev key is hardcoded as the default, and it's publicly visible in `.env-dist` and this repo. If `REFERRAL_INVITE_CODE_KEY_V1` is not set in a prod environment the active key is immediately known to any reader of the codebase, breaking the "code reveals nothing about the referrer" property entirely. Consider using `default=None` and raising `ImproperlyConfigured` when the value is absent (or is the sentinel) on non-DEBUG deployments. At minimum add a deployment runbook item so this is explicitly verified before go-live.

springfield/cms/models/pages.py
```diff
@@ -2397,30 +2398,27 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
...
-            params = urlencode({"invitation": invite_code})
-            context["invite_url"] = request.build_absolute_uri(f"/get-firefox/?{params}")
-        else:
-            # No referral-id code == no invite URL. Template needs to handle
-            # this case
-            context["invite_url"] = ""
+            try:
+                invite_code = crypto.referral_id_to_invite_code(referral_id)
+                context["invite_url"] = crypto.invite_url_for_code(invite_code)
+            except ValueError:
```
stevejalim, line 2419: Silent swallow is fine for UX, but a malformed `ref_key` should arguably be observable somewhere. If this parameter only ever comes from Firefox's referral flow, a `ValueError` here hints at an upstream bug. Consider an info-level Sentry breadcrumb (without echoing the `ref_key` value) so a spike is visible. Easy follow-up if you'd rather not block on it.

springfield/firefox/referral/utils.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    if not isinstance(referral_id, str):
+        raise ValueError("referral_id must be a string")
+    if len(referral_id) != REFERRAL_ID_LENGTH:
+        raise ValueError(f"referral_id must be exactly {REFERRAL_ID_LENGTH} characters")
+    if not is_canonical_crockford(referral_id):
+        raise ValueError("referral_id must use only uppercase Crockford base32 characters")
+
+
+def normalize_invite_code(raw: str) -> str:
```
stevejalim, line 65: `normalize_invite_code` strips whitespace and uppercases but does not apply Crockford's defined substitutions: `O→0`, `I→1`, `L→1`, `U→V`. Users who hand-type a code and misread an ambiguous glyph will get a `bad_format` Sentry event instead of a successful decode. Not a correctness bug (and Crockford substitution on encode-output is never needed), but likely to drive Sentry noise once real users exist. Easy follow-up.

springfield/settings/test.py
```diff
@@ -10,3 +10,13 @@
 PASSWORD_HASHERS = ("django.contrib.auth.hashers.MD5PasswordHasher",)
 
 logging.root.setLevel(logging.WARNING)
+
+# Fixed two-version referral invite-code keyring so the crypto tests are
+# deterministic and can exercise rotation and per-version regression fixtures.
+# Version "1" reuses the shared local/CI dev key from .env-dist. These are not
+# production values. Version identifiers are single Crockford base32 characters.
+REFERRAL_INVITE_CODE_KEYS = {
```
stevejalim, line 22: The test keyring overrides the value already validated at import time in `base.py`, so `validate_invite_code_keyring` is never called on the test keyring itself. Not a practical issue with the current hardcoded values, but a bad edit here (wrong length key, bad version identifier) would surface as an opaque crypto test failure rather than a clear `ImproperlyConfigured`. Consider calling `validate_invite_code_keyring(REFERRAL_INVITE_CODE_KEYS, REFERRAL_INVITE_CODE_ACTIVE_KEY_VERSION)` at the bottom of this file.

springfield/settings/base.py
```diff
@@ -25,6 +25,7 @@
 from sentry_sdk.integrations.rq import RqIntegration
 
 from springfield.base.config_manager import config
+from springfield.firefox.referral.utils import validate_invite_code_keyring
```
stevejalim, line 30: `utils.py` is currently safe to import here (only `django.core.exceptions` and `collections.abc` dependencies). Worth a one-line comment on the import or on the module docstring in `utils.py` saying this must remain settings-safe. The module docstring already says so, so no action required - flagging in case it ever gains a heavier import.

springfield/firefox/referral/ff1.py
```diff
@@ -0,0 +1,255 @@
...
+    where [j] is the 16-byte big-endian encoding of the block counter j. For our
+    parameters d <= 16 so only R itself is needed, but the full expansion is
+    implemented for conformance with larger domains. A fresh ECB encryptor per
+    block keeps the code obviously correct (ECB has no chaining).
+    """
+    S = bytearray(R)
+    j = 1
+    while len(S) < d:
+        block = bytes(r ^ counter for r, counter in zip(R, j.to_bytes(16, "big")))
+        encryptor = Cipher(algorithms.AES(key), modes.ECB()).encryptor()
```
stevejalim, line 100: The "fresh `Cipher` per block" note in the PR description and module docstring is clear and the NIST vectors confirm byte-identical output. Just confirming this is acknowledged as a readability choice, not a correctness concern. Safe to ship as-is; the switch to a reused cipher object is a straightforward optimisation if throughput ever matters.

springfield/firefox/referral/ff1.py
```diff
@@ -0,0 +1,255 @@
...
+
+Integer-only radix math: ``b`` is derived from ``(radix ** v - 1).bit_length()``
+which is exactly ceil(v * log2(radix)), avoiding floating-point rounding. This
+matches str4d's power-of-two handling and is exact for our radix 32 case.
+
+The tweak, alphabet ordering, and these byte layouts are part of the invite-code
+wire format. Do not change them once shipped.
+"""
+
+from __future__ import annotations
```
stevejalim: Again I don't think we need this line based on the type annotation we're doing

springfield/firefox/referral/ff1.py
```diff
@@ -0,0 +1,255 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
stevejalim: I suggest we replace this  with
```
# This code is licensed under the Apache License 2.0 (SPDX: Apache-2.0). 
# Text of the Apache License 2.0 can be found in the same directory as this file.
```

then add a new file with these contents https://github.com/capitalone/fpe/blob/v1.2.1/LICENSE

stevejalim (commented):
Looks good and ready for the tweaks listed in comments, then please re-request me for review and I can approve and we'll merge this

.github/workflows/post_deploy_asset_check.yml
```diff
@@ -56,6 +56,7 @@ jobs:
       - name: Verify deployed assets are available
         run: |
           docker run --rm \
+            -e REFERRAL_INVITE_CODE_KEY_V1=abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcd \
```
stevejalim:
```suggestion
            -e REFERRAL_INVITE_CODE_KEY_V1=0000000000000000000000000000000000000000000000000000000000000000 \
```

stevejalim: This can be a zeroed key as this check will never hit the referral views

stevejalim (approved):
r+ nice one @janriokrause!

PR #1651, opened 2026-07-27
Ff referral hub page

wen-2018:
## One-line summary

This PR modifies and adds components for referral hub page.

- adds new custom header
- modifies showcase component to have more content field and new tabs component
- adds tabs, tab, referral contrsls, impact dash, and badge components

Note:
- Figma matching styles will be in a separate PR

## Issue / Bugzilla link

[Figma design](https://www.figma.com/design/obTHe5nDCMGTlUXI8QZyIg/Firefox-Referrals?node-id=1386-18620&m=dev)
[Jira](https://mozilla-hub.atlassian.net/browse/WT-1476)

## Testing

1. Follow instructions on https://github.com/mozmeao/springfield/pull/1637
2. Add the following components, using Figma design as reference:
    - showcase
    - inside showcase, tabs, 
    - tab a: referral controls
    - tab b: impact dash
    - card list
    - kit banner
    - showcase
3. Go to one of the links from step 1 to review the page, should look like: http://localhost:8000/en-US/invite/?ref_key=TEST000000000001
4. All content in Figma should be fitted into the page with the existing and newly added components


media/js/cms/components/flare-tabs.es6.js
```diff
@@ -0,0 +1,122 @@
...
+    setSelectedTab(currentTab, setFocus) {
+        if (typeof setFocus !== 'boolean') {
+            setFocus = true;
+        }
+        for (let i = 0; i < this.tabs.length; i += 1) {
+            const tab = this.tabs[i];
+            const tabpanel = this.tabpanels[i];
+            const isCurrent = currentTab === tab;
+
+            tab.setAttribute('aria-selected', isCurrent ? 'true' : 'false');
```
knowler, line 47: Something we’ll need to do here is remove non-current tabs from the keyboard focus order since they should only be operable with arrow keys (and Home/End). We can do that by setting `tabindex="-1"` (then remove for the current or set current to `tabindex="0"`.

springfield/cms/templates/cms/blocks/tabs.html
```diff
@@ -0,0 +1,24 @@
...
+ This Source Code Form is subject to the terms of the Mozilla Public
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+<div class="fl-media-tabs">
+  {% if value.tabs %}
+    {% set section_id = value.section_id %}
+    <div class="fl-tabs">
+      <div class="fl-tabs-nav" role="tablist">
```
knowler: Is there either:

- A way to add an `aria-label` to this.
- A way to associate this with a visible element with `aria-labelledby`.

An accessible name is a requirement of `role=tablist`.

media/js/cms/components/flare-tabs.es6.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    setSelectedToNextTab(currentTab) {
+        if (currentTab === this.lastTab) {
+            this.setSelectedTab(this.firstTab);
+        } else {
+            const index = this.tabs.indexOf(currentTab);
+            this.setSelectedTab(this.tabs[index + 1]);
+        }
+    }
```
knowler, line 76: There’s a common function that could be useful here that usually gets used in implementations like need a next/prev button that support wrapping:

```javascript
/**
 * @param {number} n - number
 * @param {number} m - maximum
 * @returns {number}
 */
function mod(n, m) {
  return ((n % m) + m) % m;
}
```

Usage:

```javascript
// Get the next index, but wrap to the start if necessary (e.g. when current index is 4 and there are 5 tabs, the next value will be 0).
mod(currentIndex + 1, tabs.length);

// Get the previous index, but wrap to the end if necessary (e.g. when current index is 0 and there are 5 tabs, the previous value will be 5).
mod(currentIndex - 1, tabs.length);
```

See: https://nik.digital/posts/tab-roving

knowler (commented):
Just some early feedback on the tabs widget implementation.

springfield/cms/templates/cms/blocks/tabs.html
wen-2018: Yeah, I just haven't pushed up all local changes yet

media/js/cms/easter-egg-logo.es6.js
```diff
@@ -22,6 +22,8 @@ const setupEasterEggLogo = () => {
     if (document.body.classList.contains('flare26-kick-page')) return;
     // Enterprise pages use a wider (200x40) wordmark; our sizing tweaks don't fit.
     if (document.body.classList.contains('fl-theme-enterprise')) return;
+    // Fx referral page has a custom header
+    if (document.querySelector('header.fl-fx-share')) return;
```
stephaniehobson, line 26: Nice!

springfield/cms/templates/cms/blocks/impact-dash.html
```diff
@@ -0,0 +1,48 @@
...
+        number="{{ badge.number }}"
+        label="{{ badge.label }}"
+        badge_name="{{ badge.badge_name }}"
+        is_achieved="{{ badge.is_achieved }}"
+      >
+        {% if badge.image %}
+          <content:media>
+            {{ srcset_image(
+                badge.image,
+                "width-{80,160,240}",
```
stephaniehobson: **Issue:**  Images need to be multiples of 100 - more here: [CMS image filter specs.](https://mozmeao.github.io/platform-docs/cms/images/#allowed-filter-specs)

springfield/cms/blocks.py
```diff
@@ -2642,9 +2911,10 @@ class Meta:
 class ShowcaseBlock(blocks.StructBlock):
     settings = ShowcaseSettings()
     headline = RichTextBlock(features=HEADING_TEXT_FEATURES)
+    description = RichTextBlock(features=HEADING_TEXT_FEATURES, required=False)
     media = MediaBlock(max_num=1)
     caption_title = RichTextBlock(features=HEADING_TEXT_FEATURES, required=False)
-    caption_description = RichTextBlock(features=HEADING_TEXT_FEATURES)
+    caption_description = RichTextBlock(features=HEADING_TEXT_FEATURES, required=False)
```
stephaniehobson, line 2940: **issue:** I think it was smart to try to re-use the ShowcaseBlock for this. However, the tabs are being output inside a `<figure>` HTML element and that's going to be confusing to a user who relies on assistive tech. The caption is required because the showcase is a `<figure>` element. 

Potential fixes: 
- Edit showcase to not use a `<figure>` element when a tab block is the contents of the showcase's media section.
- Instead of re-using an existing component create a model that is only on the referral hub.
- Maybe an intro block could work?

springfield/cms/models/pages.py
```diff
@@ -2393,53 +2395,163 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
     parent_page_types = ["cms.HomePage"]
     template = "cms/referral_hub_page.html"
 
+    upper_content = StreamField(
+        [
+            ("showcase", ShowcaseBlock()),
```
stephaniehobson, line 2404: **todo:** Whatever ends up here should be limited to a max_num of 1

springfield/cms/models/pages.py
```diff
@@ -2393,53 +2395,163 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
     parent_page_types = ["cms.HomePage"]
     template = "cms/referral_hub_page.html"
 
+    upper_content = StreamField(
```
stephaniehobson, line 2402: **issue:** I think it was smart to try to re-use the ShowcaseBlock for this. However, the tabs are being output inside a `<figure>` HTML element and that's going to be confusing to a user who relies on assistive tech. 

Potential fixes: 
- Don't use a `<figure>` element when a tabblock is the contents of the showcase's media section.
- Create a separate block which is only valid on this page.
- Use an intro block.

media/css/cms/components/flare-impact-dash.css
```diff
@@ -0,0 +1,91 @@
...
+        grid-auto-flow: column;
+        grid-template-columns: none;
+    }
+}
+
+/*
+ * Locked is the default state; .is-achieved is added by ImpactDashBlock once the
+ * referrer's install count reaches the badge's number. Only the artwork is
+ * dimmed: the number and label stay at full contrast in both states so the
+ * distinction is never carried by appearance alone.
```
stephaniehobson: **question:** I'm confused by the comment, it says "distinction is never carried by appearance alone" but the only difference is the artwork dimming - no?

media/css/cms/components/flare-navigation.css
```diff
@@ -85,6 +85,17 @@ html {
     z-index: 1;
 }
 
+.fl-logo-fx.fl-logo-share-fx {
```
stephaniehobson: **todo:** Please create a referral/invitee specific CSS file for styles specific to these pages.

media/css/cms/flare-components.css
```diff
@@ -34,6 +34,8 @@
 @import 'components/flare-two-column-cards.css';
 @import 'components/flare-notification.css';
 @import 'components/flare-media.css';
+@import 'components/flare-tabs.css';
+@import 'components/flare-referral-controls.css';
```
stephaniehobson: **todo:** The referral-controls CSS should go in a separate CSS file only attached to these two pages.

springfield/cms/templates/cms/includes/referral-hub-custom-header.html
```diff
@@ -0,0 +1,18 @@
+{#
+ This Source Code Form is subject to the terms of the Mozilla Public
+ License, v. 2.0. If a copy of the MPL was not distributed with this
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+<header class="fl-header fl-force-dark-theme fl-fx-share">
+  <div class="fl-header-inner">
+      <a href="/{{LANG}}/" class="fl-logo-fx fl-logo-share-fx" aria-label="Share Firefox">
```
stephaniehobson: **nitpick:** The logo is still a link to the homepage the ALT/ARIA text should indicate that. I have mixed feelings about changing the logo as well but if Tess okayed it it's okay.

springfield/cms/blocks.py
```diff
@@ -1244,11 +1244,280 @@ class Meta:
...
+        label_format = "Tabs"
+        template = "cms/blocks/tabs.html"
+
+
 class MediaBlock(blocks.StreamBlock):
     image = ImageVariantsBlock(required=False)
     video = VideoBlock(required=False)
     animation = AnimationBlock(required=False)
     qr_code = QRCodeBlock(required=False)
+    tabs = TabsBlock(required=False)
```
stephaniehobson, line 1541: **issue:** We should remove the TabsBlock from the MediaBlock until we're sure it can go everywhere. All components which contain MediaBlocks would have to be checked and then we'd need to update the fixtures for all components to include examples of its use.

springfield/cms/middleware.py
```diff
@@ -48,6 +60,13 @@ def __call__(self, request):
                 # are fake/vuln scan requests
                 return response
 
+            if getattr(request, LOCALE_FALLBACK_EXEMPT_ATTR, False):
```
stephaniehobson, line 63: We will have to consider this case before production, it can't stay a 404.

media/css/cms/flare-components.css
```diff
@@ -52,3 +54,4 @@
 @import 'components/flare-form-fields.css';
 @import 'components/flare-dropdown.css';
 @import 'components/flare-enterprise-download.css';
+@import 'components/flare-impact-dash.css';
```
stephaniehobson: **todo:** The impact-dash CSS should go in a separate CSS file only attached to these two pages.

media/js/cms/components/flare-tabs.es6.js
```diff
@@ -0,0 +1,125 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+class TabsAutomatic {
```
stephaniehobson, line 7: I haven't had a chance to review the tabs implementation yet but I think Nathan probably steered you right.

springfield/cms/models/pages.py
```diff
@@ -2393,53 +2395,163 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
...
+        else:
+            # No referral-id code == no invite URL. Template needs to handle
+            # this case
+            context["invite_url"] = ""
+
+        context["install_count"] = self._get_install_count(referral_id)
 
         return context
 
+    def _get_install_count(self, referral_id: str) -> int:
```
stephaniehobson, line 2479: @stevejalim Can you have a look at this an other changes to this file please?

springfield/cms/templates/cms/blocks/sections/showcase.html
```diff
@@ -14,6 +14,9 @@
       heading_size="{{ 'fl-heading-size-' ~ block_level if block_level else 'fl-heading-size-2' }}"
       heading_text="{{ value.headline|richtext|remove_p_tag }}"
     />
+    {% if value.description %}
+      <div class="fl-showcase-description">{{ value.description|richtext }}</div>
+    {% endif %}
   </content:heading>
   <content:image>
```
stephaniehobson, line 27: **issue:** As mentioned before, this component renders the image in a `<figure>` which isn't the right element for tabs.

springfield/cms/templates/cms/referral_hub_page.html
```diff
@@ -5,3 +5,63 @@
...
+
+        {% include_block block %}
+      {% endfor %}
+    </div>
+  </div>
+</div>
+{% endblock %}
+
+{% block pre_footer %}
+  {% set ns = namespace(headings=0) %}
```
stephaniehobson, line 56: **note to self:** Come back to understand what's going on here.

springfield/cms/tests/test_middleware.py
```diff
@@ -614,3 +614,21 @@ def test_CMSLocaleFallbackMiddleware_alias_locale_view_restricted_page_redirects
     # Redirects to canonical URL so Wagtail's restriction enforcement fires there
     assert response.status_code == 302
     assert response.headers["Location"] == pt_br_child.url
+
+
+def test_CMSLocaleFallbackMiddleware_respects_the_locale_fallback_exemption(rf):
```
stephaniehobson, line 619: @stevejalim can you review this file please?

stephaniehobson (dismissed):
I'm not done my review but I'll submit what I have since you'll be back at work before me.

The referral dashboard pieces of this are only ever going to be on this one page. There's a few architecture I'd like to see because of that: the CSS should be in its own file instead of the shared bundle, only one of each of these page types should be allowed (max_num=1), and the data model should be coded right into the page model - you can copy the smart window page's approach.

The tab block is a good addition to the system. I'd create a more generic component for it in springfield/cms/templates/cms/blocks/sections/ and keep the impact-dash and referral-controls content somewhere else (probably just on the referral page model) then slot it into the more generic pattern. I'd also remove it from MediaBlock for now until we can do a thorough assessment of where it is and isn't appropriate, and update all the fixtures to include tab variants.

springfield/cms/models/pages.py
```diff
@@ -2393,53 +2395,163 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
...
+        to fail the whole hub render.
+
+        Note this collapses "we don't know" and "you genuinely have 0 installs"
+        into the same value. If the design ever needs to distinguish them (e.g.
+        "we couldn't load your progress"), this should return None instead.
+        """
+        # referral_id is stored in a bounded CharField, so anything longer cannot
+        # match a row. Skip the pointless round trip rather than query for it.
+        max_length = FirefoxReferralData._meta.get_field("referral_id").max_length
+        if not referral_id or len(referral_id) > max_length:
```
stevejalim: `len(referral_id) > max_length` will never be true becuse the import will fail, so this can just be `if not referral_id:`

springfield/cms/models/pages.py
```diff
@@ -2393,53 +2395,163 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
...
+            return 0
+
+        try:
+            # values_list().first() is a single SELECT ... LIMIT 1 with no model
+            # instantiation, and yields None for a missing row.
+            count = FirefoxReferralData.objects.filter(referral_id=referral_id).values_list("install_count", flat=True).first()
```
stevejalim: This is very heavily optimised, but at the costs of readability.

`FirefoxReferralData.referral_id` is set to `unique=True` so the main realistic risk here is `FirefoxReferralData.DoesNotExist` rather than the broader `DatabaseError`

media/css/cms/flare-components.css
wen-2018: Tabs will be used on /get-firefox page too. Will move referral-controls css to referral hub page only

media/css/cms/components/flare-impact-dash.css
wen-2018: Removed "never" to reflect reality.

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+            # No matching row is an ordinary outcome here -- this is a public page
+            # and any string can arrive in `ref_key` -- so it resolves to 0 rather
+            # than being reported. Only the table itself being unreadable is worth
+            # a Sentry event.
+            count = FirefoxReferralData.objects.filter(referral_id=referral_id).values_list("install_count", flat=True).first()
+        except DatabaseError as exc:
+            with new_scope() as scope:
+                scope.set_extra("exception", str(exc))
+                capture_message("Failed to read FirefoxReferralData install count", level="error")
+            return 0
```
stevejalim: This is more idiomatic and easier to reason about

```suggestion
            count = FirefoxReferralData.objects.get(referral_id=referral_id).install_count
        except FirefoxReferralData.DoesNotExist as exc:
            with new_scope() as scope:
                scope.set_extra("exception", str(exc))
                capture_message("Failed to read FirefoxReferralData install count", level="error")
            return 0
```

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+            # uppercase Crockford base32, with no case- or glyph-folding.
+            validate_referral_id(request.GET.get("ref_key"))
+        except ValueError:
+            # Without this, CMSLocaleFallbackMiddleware would see the 404, find
+            # this very page live at the same path, and redirect to it forever.
+            mark_locale_fallback_exempt(request)
+            raise Http404("Referral Hub URL is missing a well-formed ref_key") from None
+
+        response = super().serve(request, *args, **kwargs)
+        add_never_cache_headers(response)
```
stevejalim: We DO want to cache the result (for 10 mins, at the CDN, so we don't want this behaviour

```suggestion
```

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+        except DatabaseError as exc:
+            with new_scope() as scope:
+                scope.set_extra("exception", str(exc))
+                capture_message("Failed to read FirefoxReferralData install count", level="error")
+            return 0
+
+        return count or 0
+
+    def serve(self, request, *args, **kwargs):
+        """Require a well-formed ref_key, and never cache the result.
```
stevejalim:
```suggestion
        """Require a well-formed ref_key
```

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+            # than admitting some traffic we could not verify.
+            with new_scope() as scope:
+                scope.set_extra("exception", str(exc))
+                capture_message("Failed to verify referral invite code; allowing through", level="error")
+            return True
+
+    def serve(self, request, *args, **kwargs):
+        if not self._is_valid_invite_code(request.GET.get("invitation")):
+            # Locale-aware so a visitor is not forced into English.
+            return redirect(f"/{l10n_utils.get_locale(request)}/")
```
stevejalim, line 2576: nit: this will save an extra redirect hop but we can just send the user to `/` and Springfield will send them to the most appropriate version of the homepage for their locale that it can. We could simplify this here

springfield/cms/tests/test_middleware.py
stevejalim: This is legit, yeah. r+

stevejalim (approved):
Adding some comments - looks fine to me for now but will do another pass with Jan when it's merged to main. Good to keep this moving forward and off a long-running branch 

r+wc

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+        impact dashboard is one optional part of this page and must not be able
+        to fail the whole hub render.
+
+        Note this collapses "we don't know" and "you genuinely have 0 installs"
+        into the same value. If the design ever needs to distinguish them (e.g.
+        "we couldn't load your progress"), this should return None instead.
+        """
+        # referral_id is stored in a bounded CharField, so anything longer cannot
+        # match a row. Skip the pointless round trip rather than query for it.
+        if not referral_id or len(referral_id) > REFERRAL_ID_LENGTH:
```
stevejalim: [Claude] The length guard only catches IDs longer than `REFERRAL_ID_LENGTH` (16); shorter IDs (e.g. a 10-char `ref_key`) still pass through and hit the DB for a guaranteed miss. Consider `len(referral_id) != REFERRAL_ID_LENGTH` to match what `validate_referral_id` checks and skip short keys too.

springfield/cms/blocks.py
```diff
@@ -1291,7 +1560,7 @@ class _BaseContentBlock(blocks.StreamBlock):
         buttons = MixedButtonsBlock(
             button_types=get_button_types(allow_uitour),
             min_num=0,
-            max_num=3,
+            max_num=5,
```
stevejalim, line 1569: [Claude] This changes the button cap from 3 → 5 for **all** pages using `_BaseContentBlock` (intro blocks, section blocks, etc.), not just the referral hub. The PR description does not mention it — intentional for a CTA-heavy referral hub layout, or accidental scope creep? If intentional, worth a note; if accidental, revert.

springfield/cms/blocks.py
```diff
@@ -1259,11 +1259,280 @@ class Meta:
...
+    def _coerce_count(raw) -> int:
+        """Never let a missing, empty or non-numeric context value raise."""
+        try:
+            return max(int(raw), 0)
+        except (TypeError, ValueError):
+            return 0
+
+    @staticmethod
+    def _badge_context(badge, install_count: int) -> dict:
+        number = badge.get("number") or 0
```
stevejalim: [Claude] If `number` is 0 (only reachable via direct-JSON import or a migration without the `min_value=1` constraint), `install_count >= 0` is always `True` and that badge shows as achieved for every visitor including those with zero installs. Low-risk given the UI enforces `min_value=1`, but a defensive `max(..., 1)` or an early return would prevent a silent always-achieved state from bad data.

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,114 @@ def get_context(self, request, *args, **kwargs):
...
+            # visitor who is not an invitee. There is nothing to decode, and
+            # nothing worth the Sentry warning a decode attempt would raise.
+            return False
+
+        try:
+            # Case-insensitive and whitespace-tolerant, so a code retyped or
+            # copied out of a messenger still works. Raises for a malformed
+            # code or a key version we no longer hold -- either way there is no
+            # referral to credit, and crypto has already reported it to Sentry.
+            referral_id = crypto.invite_code_to_referral_id(invite_code)
```
stevejalim, line 2558: [Claude] The comment notes that `crypto.invite_code_to_referral_id` already reports to Sentry on a malformed code. Any non-empty `?invitation=` value that fails crypto (scanner traffic, misconfigured clients, typos) will fire a Sentry event on every hit. Worth confirming this is intentional for security monitoring and the alert volume is acceptable in prod, or sampling/suppressing if it is expected to be noisy.

springfield/cms/blocks.py
wen-2018: This needs to be increased to account for the usage on the referral page. Will add button limit to all other usages.

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,105 @@ def get_context(self, request, *args, **kwargs):
...
+        unknown ref_key, or the referral table being unavailable -- because the
+        impact dashboard is one optional part of this page and must not be able
+        to fail the whole hub render.
+
+        Note this collapses "we don't know" and "you genuinely have 0 installs"
+        into the same value. If the design ever needs to distinguish them (e.g.
+        "we couldn't load your progress"), this should return None instead.
+        """
+        if not referral_id:
+            return 0
```
janriokrause, line 2492: Not a big fan of this behavior :-( is this intentional right now?

media/js/cms/components/flare-tabs.es6.js
```diff
@@ -0,0 +1,125 @@
...
+            this.lastTab = tab;
+        }
+
+        this.setSelectedTab(this.firstTab, false);
+    }
+
+    setSelectedTab(currentTab, setFocus) {
+        if (typeof setFocus !== 'boolean') {
+            setFocus = true;
+        }
```
janriokrause, line 41:
```suggestion
    setSelectedTab(currentTab, setFocus = true) {
```

Is this intended to provide a default value? Since the callers either omit `setFocus` or explicitly pass `false`, could this be expressed as `setSelectedTab(currentTab, setFocus = true)` instead?

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,105 @@ def get_context(self, request, *args, **kwargs):
...
+            # uppercase Crockford base32, with no case- or glyph-folding.
+            validate_referral_id(request.GET.get("ref_key"))
+        except ValueError:
+            # Without this, CMSLocaleFallbackMiddleware would see the 404, find
+            # this very page live at the same path, and redirect to it forever.
+            mark_locale_fallback_exempt(request)
+            raise Http404("Referral Hub URL is missing a well-formed ref_key") from None
+
+        response = super().serve(request, *args, **kwargs)
+        return response
```
janriokrause, line 2525:
```suggestion
        response = super().serve(request, *args, **kwargs)
        add_never_cache_headers(response)
        return response
```

**Blocking:** The new test (`test_hub_page_serve_sets_never_cache_headers`) expects this response to be non-cacheable, but `serve()` currently returns it without adding the never-cache headers. Should we call `add_never_cache_headers(response)` before returning?

springfield/cms/models/pages.py
```diff
@@ -2431,18 +2472,105 @@ def get_context(self, request, *args, **kwargs):
...
+        to fail the whole hub render.
+
+        Note this collapses "we don't know" and "you genuinely have 0 installs"
+        into the same value. If the design ever needs to distinguish them (e.g.
+        "we couldn't load your progress"), this should return None instead.
+        """
+        if not referral_id:
+            return 0
+
+        try:
```
janriokrause, line 2494: **Blocking:** The docstring says an unavailable referral table returns 0, but the method does not catch `DatabaseError`. An actual database failure will still propagate and potentially fail the whole page.

The test at `test_referral_pages.py:193` raises exactly that exception and expects 0, so it cannot pass against the current implementation. `DatabaseError` is already imported in this module.

Maybe something like this:

```python
try:
    return FirefoxReferralData.objects.get(referral_id=referral_id).install_count
except FirefoxReferralData.DoesNotExist:
    return 0
except DatabaseError as exc:
    with new_scope() as scope:
        scope.set_extra("exception", str(exc))
        capture_message("Failed to read FirefoxReferralData install count", level="error")
    return 0
```

Also, `DoesNotExist` does not mean the database read failed. It means the query succeeded but no matching record exists.

Possibly at warning rather than error, depending on whether a valid-looking but unknown `ref_key` indicates a broken referral flow or is expected public-input noise. But also depending on whether we want to capture `DoesNotExist` or `DatabaseError`.

janriokrause (dismissed):
Some first comments, but will dig deeper.

springfield/cms/models/pages.py
stevejalim: @janriokrause @wen-2018 On a branch that I haven't PRed fully yet because of clashes, I implement a version of this where we deliberatley 404 if there is no referral_id or an invalid one.

I say we get Wen's FE code merged to main and then we do a pass, Jan, to check we're hitting the backend specs.

springfield/cms/models/pages.py
stevejalim: No - as mentioned in [my comment](https://github.com/mozmeao/springfield/pull/1651#discussion_r3706244796), we _do_ want to cache this page at the CDN (for 10 mins), so we do not want to set never-cache headers

springfield/cms/models/pages.py
stevejalim: I think this is unlikely to happen in reality, but there's a tiny case that the DB is up enough to serve the CMS page but the referral table is unhappy, so I'll fold this in now

springfield/cms/middleware.py
stevejalim: Can you elaborate? If it's 404ing because the URI (path+ query params) don't map to something we can render, then I think it's ok?

PR #1653, opened 2026-07-27
Use Django Form Fields to render field block templates

maribedran:
## One-line summary

Use Django Form Field's template rendering instead of manually building the inputs on all field block templates.

- Pass the form instance to the Contact Page's context so block templates have access to the form to render inputs
- Add a new max length option to the Text Area block

## Issue / Bugzilla link

The max length option is now required to match the new message field introduced by https://github.com/mozmeao/basket/pull/1765 

## Testing

Run `./manage.py load_page_fixtures` and check the contact page at http://localhost:8000/en-US/flare-docs/sample-pages/enterprise/contact/


stevejalim (approved):
r+ from human review, plus comments for you to decide what to do with from Claude

springfield/cms/tests/test_contact_page.py
```diff
@@ -429,7 +433,104 @@ def test_contact_page_country_select_field_renders_localized_countries(
...
+        to_email_address="test@example.com",
+        redirect_to=thank_you_page,
+    )
+    index_page.add_child(instance=page)
+    page.save_revision().publish()
+
+    content = page.serve(rf.post(page.relative_url(minimal_site), {})).content.decode()
+
+    assert 'aria-invalid="true"' in content
+    assert 'aria-describedby="first_name_error"' in content
```
stevejalim, line 532: This assertion may not be backed by anything in the code. `field.html` gives the error paragraph `id="{{ field.auto_id }}_error"`, but nothing adds `aria-describedby` to the widget itself — Django doesn't inject it automatically via `{{ field }}`. Either the attribute needs to be wired up explicitly (e.g. by mutating `form_field.widget.attrs` after validation, or via a template addition), or this test will fail. Worth verifying the test actually passes in CI before merge.

springfield/cms/templates/cms/blocks/form_fields/checkbox_group_field.html
```diff
@@ -4,15 +4,17 @@
   file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #}
 
-<fieldset class="fl-field-wrap{% if form_errors.get(value.internal_identifier) %} fl-field-error{% endif %}"{% if value.required %} aria-required="true"{% endif %}{% if form_errors.get(value.internal_identifier) %} aria-invalid="true"{% endif %}>
+{% set field = form[value.internal_identifier] %}
+
+<fieldset class="fl-field-wrap{% if field.errors %} fl-field-error{% endif %}">
```
stevejalim: The old `<fieldset>` had `aria-required="true"` and `aria-invalid="true"` when applicable. Both are dropped here. Django's `MultipleChoiceField` / `CheckboxSelectMultiple` doesn't add these to the fieldset wrapper. If they're intentionally dropped (e.g. because `required` on individual checkboxes covers it), that's fine — but it should be a deliberate decision rather than a side-effect of the refactor. Same applies to `checkbox_field.html`.

springfield/cms/models/pages.py
```diff
@@ -2178,11 +2178,9 @@ def clean(self):
 
     def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
-        form = getattr(request, "form", None)
-        context["form_errors"] = getattr(form, "errors", {})
+        context["form"] = getattr(request, "form", None)
```
stevejalim, line 2181: `context["form"]` can be `None` here if `request.form` was never set (i.e. any caller that invokes `get_context()` directly without going through `serve()`). All block templates now do `form[value.internal_identifier]` or `form.non_field_errors()`, both of which raise `TypeError` on `None`. The old code was safe: `getattr(form, "errors", {})`. The added `serve_preview()` override covers Wagtail preview, but other paths (e.g. Wagtail search indexing, unit tests that call `get_context()` directly) would crash. Consider raising `ValueError` early or asserting `form is not None` inside the templates with a clear error.

springfield/cms/templates/cms/blocks/form_fields/field.html
```diff
@@ -0,0 +1,17 @@
...
+ file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#}
+
+{% set field = form[value.internal_identifier] %}
+
+<div class="fl-field-wrap{% if field.errors %} fl-field-error{% endif %}">
+  <label class="fl-label" for="{{ field.id_for_label }}">
+    {{ value.label }}{% if value.required %} <span aria-hidden="true">*</span>{% endif %}
+  </label>
+  {{ field }}
```
stevejalim, line 13: The old per-field templates set both `required` and `aria-required="true"` on inputs for required fields. Django's widget rendering adds `required` automatically but not `aria-required`. Per current HTML spec this is technically fine (native `required` is accessible), but it's a behaviour change — anything in CSS or AT configuration that keys off `aria-required` would silently stop working. Worth a deliberate sign-off that `aria-required` is intentionally dropped.

springfield/cms/tests/test_contact_page.py
maribedran: I didn't know that Django added that to the input tag. `<input type="email" name="business_email" maxlength="320" class="fl-field" required="" aria-invalid="true" aria-describedby="business_email_error" id="business_email">`

PR #1663, opened 2026-07-30
WT-1475: Comparison table block

lucianoratamero:
## One-line summary

Add a comparison table CMS block with configurable column highlighting and mobile layout behavior.

- **new `ComparisonTableBlock`**, with fixtures, tests, and pattern lib/visual regression tests.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1475

## Testing

- run the page fixtures script;
- open http://localhost:8000/en-US/flare-docs/#test-comparison-table and check that it works as intended.


media/css/cms/components/flare-comparison-table.css
```diff
@@ -0,0 +1,124 @@
...
+}
+
+.fl-comparison-table {
+    inline-size: 100%;
+}
+
+.fl-comparison-table th {
+    font-weight: var(--fl-theme-font-weight-headline);
+    min-inline-size: 200px;
+    text-transform: uppercase;
```
maribedran, line 13: We should check if the language suppoert uppercase:
```
    @container style(--lang-supports-uppercase: true) {
        text-transform: uppercase;
    }
```

springfield/cms/tests/test_blocks.py
```diff
@@ -3576,6 +3577,84 @@ def test_uuid_block_is_not_translatable():
...
+    lower = soup.find("div", class_="fl-split-page-lower")
+    assert upper and lower
+
+    for region in (upper, lower):
+        tables = region.find_all("div", class_="fl-comparison-table-wrapper")
+        scroll_table = next(t for t in tables if "scroll" in t.get("class", []))
+        assert_comparison_table(scroll_table, scroll_variant)
+
+
+def test_comparison_table_stacked_highlighted(index_page, rf):
```
maribedran: There's no need to split those two tests since they both do the same thing. A single test can call the same assert method and check for the variant class according to the block's settings.

springfield/cms/blocks.py
```diff
@@ -2905,6 +2906,83 @@ class Meta:
         label = "Enterprise Download"
 
 
+# Comparison Table
+
+
+class ComparisonTableCellSettingsBlock(blocks.StructBlock):
```
maribedran: We should use Wagtail's block settings standard. We'll move all blocks to that at some point.

media/css/cms/components/flare-comparison-table.css
lucianoratamero: nah, it's VERY old, no need to worry about compatibility on this one.

maribedran: It's not a compatibility issue. There are languages where uppercasing changes the meaning of words. If you look through the code, you'll see that pattern all over. I'll try to find the Slack thread where that was discussed.

springfield/cms/tests/test_blocks.py
```diff
@@ -3576,6 +3577,65 @@ def test_uuid_block_is_not_translatable():
...
+    soup = BeautifulSoup(response.content, "html.parser")
+    upper = soup.find("div", class_="fl-split-page-upper")
+    lower = soup.find("div", class_="fl-split-page-lower")
+    assert upper and lower
+
+    for region in (upper, lower):
+        tables = region.find_all("div", class_="fl-comparison-table-wrapper")
+        for variant in variants:
+            mobile_behavior = variant["value"]["mobile_behavior"]
+            table = next(t for t in tables if mobile_behavior in t.get("class", []))
```
maribedran: This is only testing one of the variants. What I meant on my last comment is that all variants can be tested by the same test. You just need to fetch them by the index.

maribedran (approved):

PR #1686, opened 2026-08-04
feat: ensure users outside of the target territory cannot access referral hub or invitation page

stevejalim:
For MVP this is limited only to users in the US, and that includes invitees - they have to be in the US too. If they are not, we redirect to the homepage, from where they can at least download Firefox.

Note that we don't show a message to explain things: we'd need a special version of the destination page because we can't do session-based messages and parking the message in the querystring would be a reputational vulnerability.

Unit tests passing should be enough for now. We'll do full QA one the get-firefox page is built


stephaniehobson (approved):
R+?

Requesting a page outside of the en-US local resulted me being forwarded to firefox.com and not localhost if this behaviour is expected then R+

Does let me request the pages with valid codes in the US. Does not let me request the pages in languages other than English (en-CA thanks you). Does not let me request pages from outside of the US.

PR #1688, opened 2026-08-05
WT-1476 Ff referral hug page styling

wen-2018:
## One-line summary

This PR styles the referral hub page.

Page styles and functions, with exception of the badge connection lines

## Issue / Bugzilla link

[Figma](https://www.figma.com/design/obTHe5nDCMGTlUXI8QZyIg/Firefox-Referrals?node-id=1415-20429&m=dev
)
[Jira](https://mozilla-hub.atlassian.net/browse/WT-1476)

## Testing

- Follow instructions on https://github.com/mozmeao/springfield/pull/1637
- Go to one of the links from step 1 to review the page, should look like: http://localhost:8000/en-US/invite/?ref_key=TEST000000000001
- Update the page in Wagtail admin to match the Figma


springfield/cms/templates/components/referral-controls.html
```diff
@@ -4,7 +4,7 @@
  file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #}
 
-{# props invite_url="", copy_label="", copy_success_label="", qr_label="", email_label="", email_subject="", email_body="" #}
+{# props invite_url="", copy_label="", copy_success_label="", qr_heading="", qr_label="", email_label="", email_subject="", email_body="", qr_dialog_id="fl-referral-controls-qr-dialog" #}
```
janriokrause: Small thing that might be worth future-proofing: `qr_dialog_id` defaults to the static value `fl-referral-controls-qr-dialog`.

I think this is fine as long as there's only one referral-controls block on the page, but since referral controls live inside CMS tabs, adding them to more than one tab later could create duplicate dialog IDs. The dialog trigger looks up the target by ID, so that could make a later QR button open the wrong dialog. Maybe we could make the ID unique from the tab context, like `section_id` + `tab_index`?

media/css/cms/components/flare-tabs.css
```diff
@@ -77,3 +76,9 @@
         display: block;
     }
 }
+
+@media (--viewport-md-up) {
+    button[role='tab'] {
```
janriokrause:
```suggestion
    .fl-tabs button[role='tab'] {
```

Before this PR, `min-inline-size: 122px` belonged to the nested `.fl-tabs button[role='tab']` rule. The PR moves it into a top-level media query but leaves the selector as `button[role='tab']`. At the medium breakpoint and above, this now changes every ARIA tab button on any page that loads this stylesheet, including tab implementations unrelated to Flare.

media/css/cms/components/flare-referral-controls.css
```diff
@@ -8,29 +8,129 @@
...
+    gap: var(--token-spacing-sm);
 }
 
 .fl-referral-controls-invite {
+    background-color: var(--token-color-medium-purple);
     border-radius: var(--token-border-radius-xl);
+    margin: var(--token-spacing-xl) auto;
+    max-inline-size: 100%;
+    padding: var(--token-spacing-sm) 0 var(--token-spacing-sm)
+        var(--token-spacing-lg);
```
janriokrause: The four-value `padding` shorthand always places the large padding on the physical left. The child paragraph also has `padding-inline-end`, which becomes left padding in RTL. As a result, RTL pages get both padding values on the left and no start padding on the right, even though the mask direction is correctly reversed. 

Use logical properties so the two sides mirror correctly:

```css
.fl-referral-controls-invite {
    padding-block: var(--token-spacing-sm);
    padding-inline-start: var(--token-spacing-lg);
    /* or maybe */
    padding-inline: var(--token-spacing-lg) 0;
}
```

wen-2018: Good catch!

janriokrause (approved):
The implementation fix looks good. It would be useful to add a test that renders referral controls in two tabs, since the current single-instance test would not catch duplicate dialog IDs. I would not consider this blocking, but it would cover the exact regression.

PR #1693, opened 2026-08-06
Remove BOM in complied Sass for protocol.css

stephaniehobson:
## One-line summary

Prevents a UTF-8 BOM appearing mid-file between concatenated CSS chunks.


slightlyoffbeat (approved):
r+

Was able to reproduce and confirm this locally.

PR #1697, opened 2026-08-06
Support delayed auto-open for the floating QR code snippet

slightlyoffbeat:
## One-line summary

Add a third "closed, then auto-opens after a delay" starting state to the floating QR code snippet (Thanks page + WNP).

Adds a third start state to the floating QR snippet: render collapsed, then auto-open after a configurable delay (default 3s). Previously open-or-closed only.

- **Model (add-only).** New `open_behavior` + `open_delay_ms` on `QRCodeFloatingSnippet`, mirrored as per-page overrides. Legacy `default_open` booleans are kept as a fallback, so the one migration is **8 `AddField`s — no drops/renames**, and no data or revision is touched. Dropping the legacy fields is a follow-up.
- **Resolution logic** — `resolve_open_behavior` / `resolve_qr_source`: page override → snippet → legacy booleans; dismissed cookie forces closed. `QROpenBehavior` lives in `models/base.py` to dodge a circular import.
- **Template / JS.** `delayed` renders collapsed; JS auto-opens from `data-open-delay` and cancels the timer on user interaction. Also fixes a pre-existing bug where the toggle's `aria-label` wasn't updated on click.
- **CSS.** Content fades in as the panel grows; adds a `prefers-reduced-motion` guard and repairs the inert `@starting-style`. `.fl-qr-code-floating-*` classes are shared with the QR modal — verified unaffected.

## Issue / Bugzilla link

_<add ticket>_

## Testing

- Migration is add-only and reversible; `makemigrations --check` clean.
- Manual QA: set a Thanks/WNP page to **Closed, then opens automatically**, clear the `moz-qr-snippet-dismissed` cookie, reload → renders collapsed, auto-opens at the delay with content fading in, icon + `aria-label` flip. Confirm `open`/`closed` unchanged, closing during the delay window keeps it closed, and reduced-motion shows it appear without animating.


media/js/cms/components/flare-qr-code-snippet.es6.js
```diff
@@ -6,8 +6,85 @@
...
+        icon.classList.add(isOpen ? 'fl-icon-subtract' : 'fl-icon-add');
+
+        const button = snippetEl.querySelector('.fl-qr-code-snippet-close');
+        const label =
+            button &&
+            button.getAttribute(isOpen ? 'data-label-hide' : 'data-label-show');
+
+        if (label) {
+            icon.setAttribute('aria-label', label);
+        }
```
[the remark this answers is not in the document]
slightlyoffbeat: good catch. This has been addressed

media/css/cms/components/flare-floating-qr-overlay.css
```diff
@@ -103,6 +114,16 @@
     flex: 1;
     flex-direction: column;
     justify-content: center;
+    opacity: 1;
+
+    /*
+     * Fade the content in as the panel grows, so a delayed auto-open reveals
+     * the QR code smoothly rather than snapping it in at the end of the size
+     * transition. `allow-discrete` lets the `display` change below animate.
```
stephaniehobson:
```suggestion

```

stephaniehobson: Technically wrong and also to much information.

(the code is fine, it's the the comment that's wrong)

stephaniehobson (approved):
I would prefer to migrate the value of `default_open` into a corresponding value in `open_behavior` over maintaining two separate code paths and returning to clean them up later.

If you would prefer to roll forward and clean it up later please create a follow up ticket and link to it here.

I found no launch blocking issues in your changes.

Pre existing problems:

- It is possible to create multiple "QR Code Floating Snippets" and there is no mechanism to pick between them.
  - A more informative label like "Mobile Firefox Floating QR Code Button" could be used in the settings since there is no way to pick between different snippets.
- Clicking anywhere on the heading opens it but dismissing it requires clicking on the icon.
- Panel tab placement differs between page types (thanks vs freeform/whatsnew)
- `override_translatable_fields` is defined twice in both FreeFormPage2026 and WhatsNewPage2026 — and the second definition silently overwrites the first.
- I would be nice to group the snippet over-rides together visually some how and the "heading" for the section is right up against the label for the first over-ride.
- The use of "updated" and "legacy" language is ambiguous since it could refer to two different things.

slightlyoffbeat:
Decided to update this one a bit more after reviewing Stephanie's comments. I'm calling the old one "gen 1" and this one "gen 2". I know there's a case for calling this 'gen 3', but i view this as an update to the second version. Updated:

- Backfill `default_open` / `floating_qr_default_open` into `open_behavior` and collapse resolution to a single code path. The old columns are kept but unread — dropped in a follow-up (deploy safety on live-traffic tables).
- Fix a duplicate `override_translatable_fields` on FreeForm / What's New 2026 that had dropped `version` + `pre_footer_image` from sync.
- Move `transition-behavior` out of the QR overlay's transition shorthand so the minified bundle keeps the close animation.
- Clarify snippet help_text (first-/second-generation labels).

A follow-up PR will drop the now-unread `default_open` / `floating_qr_default_open` columns once this is deployed, keeping the `DROP COLUMN` on its own migration.

springfield/cms/templates/cms/snippets/qr-code-floating-snippet.html
```diff
@@ -7,20 +7,20 @@
...
-      <span class="fl-icon fl-icon-device-mobile"></span>
+    <div class="fl-qr-code-prompt" id="fl-qr-code-heading">
+      <span class="fl-icon fl-icon-device-mobile" aria-hidden="true"></span>
       {{ value.heading|richtext }}
     </div>
-    <button class="fl-qr-code-snippet-close fl-qr-code-floating-button" data-label-show="{{ ftl('ui-show-more') }} {{ value.heading|richtext|remove_p_tag }}" data-label-hide="{{ ftl('ui-show-less') }} {{ value.heading|richtext|remove_p_tag }}">
+    <button class="fl-qr-code-snippet-close fl-qr-code-floating-button" aria-labelledby="fl-qr-code-heading" aria-controls="fl-qr-code-content" aria-expanded="{% if value.qr.open %}true{% else %}false{% endif %}">
       {% if value.qr.open %}
-        <span class="fl-icon fl-icon-subtract" aria-label="{{ ftl('ui-show-less') }} {{ value.heading|richtext|remove_p_tag }}"></span>
+        <span class="fl-icon fl-icon-subtract" aria-hidden="true"></span>
```
stephaniehobson: hidden when open?

stephaniehobson (approved):
Updates look great. Thanks for adding the extra polish and addressing some old bugs!

PR #1700, opened 2026-08-07
WT-1493 Blog Topic page

maribedran:
## One-line summary

List articles from a topic under `/topics/{topic_slug}/` and create a new Blog Topic Page to allow editors to customize it.

- Add a new `/topics/{slug}/` route to the Blog Index Page listing articles from a topic
- Create a new Blog Topic Page with custom header and featured articles blocks to override the basic listing view
  - The page's URL and route method are overridden so that the URL matches the `/topics/{slug}/` path and that serving is delegated to the parent page
- Some index page methods were refactored to be used by both routes
- Some template sections were extracted into reusable includes
  - The featured articles template drops the published date since we have another ticket requesting that listing pages don't show the dates

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1493

## Testing

Run `./manage.py load_page_fixtures` and check the following pages:
- Index: http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/ 
- Basic topic listing: http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/topics/tips/
- Curated topic page: http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/topics/privacy/


springfield/cms/templates/cms/includes/blog/blog-featured-articles.html
```diff
@@ -0,0 +1,116 @@
...
+        </div>
+      </div>
+
+      {% if featured_articles|length > 1 %}
+        <div class="fl-blog-article-list">
+          {% for block in featured_articles[1:4] %}
+            {% set description = block.value.get_description() %}
+            <include:blog-article-list-item>
+              <content:heading>
+                <hgroup class="fl-heading-group">
```
stevejalim, line 62: Is removing the date display here intentional? The original `blog_index_page.html` rendered `first_published_at` for items 2–4 of the featured row (via `<content:date>`), and the corresponding test assertion was dropped. If this is a deliberate design decision it should be noted in the PR description; if accidental it is a visual regression on the index page.

springfield/cms/models/pages.py
```diff
@@ -1733,75 +1786,55 @@ class Meta:
...
 
-    def _prefetch_streamfield_articles(self):
-        """Bulk-fetch all BlogArticlePages referenced in featured_articles and cards_lists,
-        and populate _article_cache on each block value to avoid per-block DB queries."""
-        from springfield.cms.models.snippets import BlogTopic, Tag  # circular import
-
+    def prefetch_streamfield_articles(self):
+        """Prefetch every article referenced by this page's featured_articles and cards_lists."""
         # StreamField iteration yields BoundBlocks; their .value is BlockArticleValue.
         # ListBlock iteration yields StructValues (BlockArticleValue) directly.
```
stevejalim, line 1793: Worth a quick manual check: does the Wagtail sitemap generator pick up the overridden URL here? `get_sitemap_urls()` delegates to `get_full_url()` → `get_url_parts()`, so in theory the sitemap should list `/topics/<slug>/` rather than the tree path. But if any sitemap plugin iterates the page tree and calls `route()` instead, every `BlogTopicPage` would 404. Confirming the sitemap output contains the correct URL for the curated Privacy page (from the fixture) would close this off.

stevejalim (commented):
Two things to confirm before merge (see inline comments).

springfield/cms/models/pages.py
```diff
@@ -1843,30 +1876,27 @@ def index_route(self, request):
     def topics_route(self, request):
         return self._render_route(request, "cms/blog_topics_page.html")
 
+    @path("topics/<slug:topic_slug>/")
```
stevejalim, line 1880: Question, non-blocking: Do you think there's a risk of confusion that CMS editors might not find the topics list in the page tree?

stevejalim (approved):
r+ with some questions, comments

springfield/cms/models/pages.py
maribedran: They probably won't unless we train them. CMS training is on our priority list (who knows when we get to that, though).

springfield/cms/templates/cms/includes/blog/blog-featured-articles.html
maribedran: It's intentional. There's another ticket dropping it. I'll add it to the description.

springfield/cms/models/pages.py
maribedran: @stevejalim thanks for flagging that. It revealed a bug in our sitemap URLs. Fixed here: https://github.com/mozmeao/springfield/pull/1700/commits/b3b9b2efb695bb647cfd276dcb6a9ac4be4940c9

PR #1702, opened 2026-08-08
User Routing 1/4: schema, signal registry and validation

slightlyoffbeat:
## One-line summary

Add the User Routing schema, signal registry and save-time validation — layer 1 of 4, behind the `user_routing` waffle switch.

## The stack

This is [#1681](https://github.com/mozmeao/springfield/pull/1681) re-cut as four stacked PRs. The feature
ships dark behind the `user_routing` waffle switch, which is what makes splitting it safe: every layer can
merge without exposing anything to visitors. Nothing is served until layer 4 lands **and** someone flips
the switch.

| PR | Layer | What it adds |
|---|---|---|
| **this one** | Schema | Models, migration, signal registry, validation |
| 2 | Runtime | The resolver page and client-side evaluation |
| 3 | Admin | The authoring surface |
| 4 | Adoption | `WhatsNewPage2026` adopts it |

**This layer is the data model, and can be judged as one.** Three models, one migration and a registry —
no view, no URL, no template, and no consumer until layer 4. Reviewing it alone means assessing the schema
without the runtime and admin code sitting on top of it. Where a decision here looks unmotivated, layer 2
or 3 is usually what it was made for.

- **The three models (`springfield/cms/routing/models.py`) — the most important thing to review.** All
  three are keyed to `wagtailcore.Page` rather than to a consumer page type, so one generic table set
  serves every future consumer and adopting the framework later adds **no migration**. `RoutingRule` is an
  ordered rule hosted by a canonical page; `RoutingCondition` is one `<signal> <operator> <expected-value>`
  clause, with a rule's conditions forming an ordered conjunction (AND — there is no rule-level OR or NOT,
  negation is the operator's paired form); `RoutingConfig` is a per-page 0-or-1 record holding the
  `routing_paused` kill switch, with room for future per-page settings.

- **`clean()` validation is server-side, not just admin JS.** The operator must be legal for the signal's
  value type, an enum expected value must be a member of its set, and a rule's target must be a strict
  descendant of the canonical it attaches to. The descendant guard judges the **localized** target
  (`localized_target`), not the raw foreign key: translating a page copies its rules with the *source*
  locale's target still stored, so checking the raw target would reject every translated page's own rules.

- **The signal registry (`signals.py`, `v1_signals.py`).** Signals are declared once with a source, value
  type and legal operators; everything else — admin choices, the reference page, client metadata — is
  derived from the registry so it cannot drift. `CmsConfig.ready()` imports `v1_signals` to populate it;
  without that import the registry is empty everywhere.

- **Mechanical, skim these:** `value_lists.py` (the closed-set value lists), and the registration of the
  three models under the `cms` app in `models/__init__.py` so the migration is discoverable.

## Testing

```bash
python -m pytest springfield/cms/routing/tests/ -q
python -m pytest springfield/cms/ -q
python manage.py makemigrations --check --dry-run
```

Expected: **99 routing tests** pass on this layer (the suite grows to 383 by layer 4), and
`makemigrations --check` reports no changes — the schema must stay at exactly one migration.

There is nothing to click through: this layer has no admin surface and serves no page. The kill switch,
the routing tab and the resolver all arrive in later layers.

### If those commands fail with an `OperationalError`

Only affects a database that applied a **previous version of this migration** — from the old `#1681`
branch, where it was numbered `0125` and amended in place several times. Django never re-runs an applied
migration, so such a database keeps the older schema. It is now `0129`, renumbered because main gained
`0125`–`0128` while this work was in flight.

```bash
python manage.py migrate cms 0124 && python manage.py migrate cms
```

Safe to run — the feature is dark and holds no data. Note `makemigrations --check` stays green regardless,
because it compares models to migration *files* and not to your database, so CI will not catch this for
you. Fresh databases are unaffected.


springfield/cms/migrations/0129_routingconfig_routingrule_routingcondition.py
stevejalim, line 1: The migrations in this chain of PRs will need renumbering

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Framework-owned, Page-keyed routing schema.
```
stevejalim: "Framework" in a Django/Wagtail project tends to refer to Django or Wagtail itself - if this is referring to the routing logic, Claude needs to rephrase

stevejalim: Or we just lose that line - it doens't add anything compared to what follows

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+
+Three models, all keyed to ``wagtailcore.Page`` so a single generic table set is
+shared by every consumer page type — there is no per-consumer model, and adopting
+the framework adds no migration:
```
stevejalim: > the framework adds no migration:

💯 100: I really like the idea here

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+
+* ``RoutingRule`` — an ordered rule hosted by a canonical page, resolving to a
+  target page that must be a descendant of that canonical.
+* ``RoutingCondition`` — one ``<signal> <operator> <expected-value>`` clause; a
+  rule's conditions form an ordered conjunction (AND).
+* ``RoutingConfig`` — a per-page 0-or-1 record carrying the ``routing_paused`` kill
```
stevejalim: Calm down, Claude ;) 

```suggestion
* ``RoutingConfig`` — a per-page record carrying the ``routing_paused`` kill
```

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+    return [(SOURCE_LABELS[source], grouped[source]) for source in Source if source in grouped]
+
+
+class ExcludingSelect(forms.Select):
+    """A ``Select`` that always withholds one option value.
+
+    Filtering on *assignment* rather than once at construction is deliberate:
+    ``ChoiceField`` pushes its own choices onto its widget whenever they are set, so
+    choices passed to the constructor are overwritten before the field ever renders.
+    Optgroups left empty are dropped rather than rendered as bare headings.
```
stevejalim: Less is more in the docstring?

```suggestion
    """A ``Select`` that always withholds one option value, enabling a form that narrows down options
```

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+    rules without its own model or migration. Priority is the ``sort_order``
+    position; ties break by ascending id (older rule wins) so ordering is
+    deterministic across database engines. There is no draft/status field — a rule
+    exists iff its parent page is published.
+    """
+
+    page = ParentalKey("wagtailcore.Page", on_delete=models.CASCADE, related_name="routing_rules")
+    target = models.ForeignKey(
+        "wagtailcore.Page",
+        # SET_NULL (not CASCADE, and not PROTECT): deleting the target clears it from this rule
```
stevejalim: Why not CASCADE here? If a Page is deleted and leaves this Rule in place, is there a clean way to reach/view it? Is there a way to reattach it?  If not, I say we let it be zapped with the page it was attached to.

stevejalim (approved):
r+wc 

Side point: IMO Claude's overly verbose docstrings distract from reading the actual code

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+            raise ValidationError({"signal": _("Unknown signal “%(name)s”.") % {"name": self.signal}})
+        signal = registry.get(self.signal)
+
+        # The operator must be legal for the signal's value type.
+        if not signal.allows_operator(self.operator):
+            raise ValidationError(
+                {"operator": _("Operator “%(operator)s” is not valid for the “%(name)s” signal.") % {"operator": self.operator, "name": self.signal}}
+            )
+
+        # An enum condition's expected value(s) must be members of the enum set.
```
stevejalim, line 325: [Codex] `RoutingCondition.clean()` validates enum and closed-set membership, but it does not reject malformed boolean, integer, or version values. Because the admin JavaScript intentionally lets hidden-tab rows fall through to server validation, values such as `yes` can be persisted; the evaluator then coerces `yes` to false, which can route the wrong audience. Please add type-aware server-side validation and tests for these scalar value types.

springfield/cms/routing/v1_signals.py
```diff
@@ -0,0 +1,210 @@
...
+        value_type=ValueType.BOOLEAN,
+        browser_state_key="sync",
+    )
+)
+
+registry.register(
+    RoutingSignal(
+        name="ai_controls",
+        description=_("The state of the browser's AI controls."),
+        source=Source.UITOUR,
```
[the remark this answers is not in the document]
slightlyoffbeat: Confirmed: config.state was always undefined and the signal could never match.

Rather than expose seven signals, ai_controls is now a single enum summarizing what the user has actually chosen: neutral, enabled_some, blocked_some, blocked_all, mixed. That keeps rules readable and means a new Firefox AI feature doesn't need a new signal here.

The declaration is fixed in this PR; the client-side derivation lands in [#1704]

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+        if signal.value_type is ValueType.ENUM:
+            members = {enum_value.value for enum_value in signal.enum_values}
+        else:
+            # A STRING signal whose domain is fully known (locale / language / country) is an
```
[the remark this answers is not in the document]
slightlyoffbeat: Fixed. RoutingCondition.clean() now validates booleans, integers and versions against the grammar the client evaluator actually parses.

springfield/cms/routing/models.py
```diff
@@ -0,0 +1,404 @@
...
+    panels = rule_panels()
+
+    def __str__(self):
+        if self.name:
+            return self.name
+        if self.match_all:
+            summary = _("all triggered visitors")
+        else:
+            summary = ", ".join(str(condition) for condition in self.conditions.all()) or _("no conditions")
+        return f"{summary} → target {self.target_id}"
```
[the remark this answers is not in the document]
slightlyoffbeat: Fixed. The whole string is now interpolated through a single gettext_lazy call rather than f-stringing a translated summary into untranslated text.

springfield/cms/routing/tests/test_signals.py
```diff
@@ -0,0 +1,157 @@
...
+
+def test_registry_only_accepts_routing_signals():
+    reg = RoutingSignalRegistry()
+    with pytest.raises(RoutingSignalError):
+        reg.register(object())
+
+
+def test_naming_avoids_django_dispatch_signal_collision():
+    # The metadata class is named RoutingSignal, never bare Signal.
+    import springfield.cms.routing.signals as module
```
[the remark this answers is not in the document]
slightlyoffbeat: Fixed. The comment now explains the placement (the test inspects the module object for an absent attribute, so it needs the module bound to a name) rather than restating the assertion.

springfield/cms/routing/models.py
slightlyoffbeat: fixed.

slightlyoffbeat: fixed

springfield/cms/routing/models.py
slightlyoffbeat: updated with your suggestion

springfield/cms/routing/models.py
slightlyoffbeat: agree. Took your suggestion and edited down a few others as well

springfield/cms/routing/models.py
slightlyoffbeat: Same fix as the thread above: type-aware validation for boolean, integer and version, matching the client evaluator's grammar, with tests.

springfield/cms/migrations/0129_routingconfig_routingrule_routingcondition.py
slightlyoffbeat: ya, lets leave till just before we are ready for a merge. Can update once, monitor CI, and move forward

stevejalim (approved):

PR #1710, opened 2026-08-10
[WT-1491] Blog Tags

maribedran:
## One-line summary

Create the Blog Tag model based on taggit and use it on Blog Article pages.

- Create the new `BlogTag` model using the `taggit` library, adding revision and translation features
- Replace taggit's autocomplete field with a custom implementation that only returns published tags in the default locale
  - The Blog Article page links to the default locale's tags, and translations render the localized tags queried by the pages' contexts
  
## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1491

## Testing

- Check http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/
- Go to http://localhost:8000/cms-admin/snippets/cms/blogtag/?locale=en-US and pick a tag to translate
- Publish the translated tag
- Translate the index page to the same language and check that it displays the translated tag


springfield/cms/management/commands/link_translations_after_export.py
```diff
@@ -55,6 +56,7 @@
     BannerSnippet,
     Tag,
     BlogTopic,
+    BlogTag,
```
stevejalim, line 59: Why do these need to be imported if they're not directly referenced later on? Won't this cause a linting fail?

springfield/cms/migrations/0133_alter_blogarticlepage_tags.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0132_blogtag_taggedblogarticle"),
+    ]
+
+    operations = [
+        # The old tags M2M and the new taggit relation live in different tables, so the
+        # column-less AddField is safe: RemoveField drops cms_blogarticlepage_tags and with
+        # it every existing tag assignment, which is intended — see the design doc.
```
stevejalim: > see the design doc

if the doc exists, we should link to it. If not, let's lose the comment

springfield/cms/fields.py
```diff
@@ -368,3 +369,16 @@ def to_python(self, data):
                 raise validation_error
 
         return f
+
+
+class LocalizedClusterTaggableManager(ClusterTaggableManager):
```
stevejalim, line 374: Something about this feels like a wrinkle: I think that it's because the concept (a manager whose form filters tags to one locale) is locked in implementaton to a very Blog-specific field, making it potentially risky to reuse elsewhere?

stevejalim: Maybe generalizing `BlogTagField` to something more generic `LocaleTagField` / `SpringfieldLocaleTagField`? might sidestep this?

springfield/cms/migrations/0133_alter_blogarticlepage_tags.py
maribedran: Thanks! I always clean up those Claude comments, but clearly not well enough :/

springfield/cms/management/commands/link_translations_after_export.py
maribedran: I think you just got misled by the collapsed lines on the UI. 
```
from springfield.cms.models.snippets import (
    BannerSnippet,
    BlogTag,
    BlogTopic,
    PreFooterCTAFormSnippet,
    PreFooterCTASnippet,
    Tag,
)

...

TRANSLATABLE_SNIPPET_MODELS = [
    PreFooterCTASnippet,
    PreFooterCTAFormSnippet,
    BannerSnippet,
    Tag,
    BlogTopic,
    BlogTag,
]
```

maribedran: I also spent a while looking for the where this import was used.

springfield/cms/fields.py
maribedran: @stevejalim What do you think? https://github.com/mozmeao/springfield/pull/1710/commits/f1ce6841019a5a510d369e52cc01c1d882417194

stevejalim (approved):

PR #1711, opened 2026-08-10
Custom attributon for the invite-referral page

stephaniehobson:
## One-line summary

Custom attribution for the invite-referral page

Referral attribution should only happen on this page (because this is the only place the user can give consent).

To confine it to this page we are downloading directly from this page, rather than forwarding to /thanks.

Linux and iOS do not support the referral codes (at the moment).

- Adds a Wagtail block to support the custom referral-attribution download buttons
- Displays a checkbox for consent (checked by default)
- Adds a separate referral-attribution JS flow for Desktop
  - Reads the invitation code from the HTML template (Django does the verification)
  - Constructs the referral information and gets it signed
  - Attaches it to the appropriate links
  - Adds `referralActive` to stubAttribution which prevents it from clobbering the custom referral attribution
- Over-rides mobile links with attribution 
  - Returns original attribution when check-box un-checked

## Issue / Bugzilla link

[https://mozilla-hub.atlassian.net/browse/WT-1455](https://mozilla-hub.atlassian.net/browse/WT-1455)

## Testing

Setup:

- [Generate dummy data according to the instructions in #1637](https://github.com/mozmeao/springfield/pull/1637)

In Wagtail:

1. Create a page from the "Referral Program: Invitee / Get Firefox Page" template
2. In the "Upper Content" section add a "Kit Intro"
3. In the "Button" section pick "Referral Download CTA"
4. Publish the page.
5. [Load the page with a valid dummy code.](http://localhost:8000/en-US/get-firefox/?invitation=1HR4FZ672Z8Y0E4HW)

In a supported desktop browser (ideally, Chrome):

1. [Load the page with a valid dummy code.](http://localhost:8000/en-US/get-firefox/?invitation=1HR4FZ672Z8Y0E4HW)
2. Make sure only one button appears
3. Right click the download button and copy the link
4. Check it contains both a `attribution_code` and `attribution_sig` attribute
5. [Decode](https://www.base64decode.org/) the value of the `attribution_code` and make sure it includes `content=fxrefer**1HR4FZ672Z8Y0E4HW**` where the bold part matches the referral code you arrive on the page with.
6. Un-check the checkbox and see that the `attribution_code` no longer contains a `fxrefer`. (It may still contain other stuff, like `client_id_ga4`, technically it should be re-set to the regular stubAttribution first-touch values).
7. Re-check the download button
8. Click the download button (did download?)
9. Log `dataLayer` to the console and make sure it includes `product_download` and `firefox_download` events

In Android (you can spoof the UA string for testing):

- Inspect the Play Store link element in the DOM, the correct utm params should be visible in the href.
- Toggle the check box a few times to make sure it's removing and adding it back appropriately.
  - We don't have first-touch attribution for mobile so there isn't any expectation it will restore some other atttibution values.

In an un-supported browser (iOS or Linux):

- No checkbox appears.
- You can still download Firefox.


wen-2018 (changes requested):
Still seeing checkbox with iOS or Linux, but everything else is working as described.
ios

stevejalim (commented):
Sorry if some of these comments about encoding `:` to `%3A` are off-target - it looks like it's already done for Android but needs doing for Desktop?

(This is based on the string checks in https://searchfox.org/firefox-main/source/browser/components/attribution/AttributionCode.sys.mjs#48)

stevejalim (approved):

wen-2018 (approved):

PR #1713, opened 2026-08-11
Move the invalid `ref_key` Sentry probe into `serve`

janriokrause:
`serve` 404s a malformed `ref_key` before `get_context` runs, so the probe could only ever fire for CMS preview and other callers that skip `serve`. It was watching editors, not the referral flow.

`get_context` still tolerates a bad key for those callers, and now returns early instead, which drops a guaranteed-miss install lookup.

### DIFF
--- a/springfield/cms/models/pages.py
+++ b/springfield/cms/models/pages.py
@@ -2529,25 +2529,27 @@ class ReferralHubPage(AbstractSpringfieldCMSPage):
 
         context = super().get_context(request, *args, **kwargs)
 
+        # The defaults stand for every case where there is no usable `ref_key`:
+        # no invite link to offer and no progress to report.
         context["invite_url"] = ""
-        if referral_id := request.GET.get("ref_key"):
-            try:
-                invite_code = crypto.referral_id_to_invite_code(referral_id)
-                context["invite_url"] = crypto.invite_url_for_code(invite_code)
-            except ValueError:
-                # Treated like a missing `ref_key` rather than raising. Only a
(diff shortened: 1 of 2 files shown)


springfield/cms/models/pages.py
janriokrause, line 2571: Unreachable now. Remove or leave for future callers?

springfield/cms/models/pages.py
```diff
@@ -2584,16 +2586,31 @@ def serve(self, request, *args, **kwargs):
...
+            # Only a correctly sized `ref_key` is reported, because this is a
+            # public page and anything can land in the query string. At the right
+            # length it plausibly came from the referral flow, so a spike is worth
+            # seeing. The value itself is masked out of the event by
+            # `SENSITIVE_FIELDS_TO_MASK_ENTIRELY`.
+            if len(ref_key) == REFERRAL_ID_LENGTH:
+                with new_scope() as scope:
+                    scope.fingerprint = ["referral-hub-invalid-ref-key"]
+                    capture_message(
+                        "ReferralHubPage received a ref_key that is not a valid referral ID",
```
stevejalim, line 2611: Given how broad a referral key can be I guess what we'll see here, if anything, is noise from 16-char tokens that feature non-Crockford chars. Will be interesting, but if we start getting a lot of noise can tune this out.

stevejalim (approved):
r+ but happy to chat about this later

stevejalim (approved):

PR #1730, opened 2026-08-13
allowing testimonial cards to have no attribution

slightlyoffbeat:
## One-line summary
Make the Testimonial block's Attribution field optional.

- `CardTestimonialBlock.attribution` is now `required=False` (`springfield/cms/blocks.py`).
- Both testimonial templates now render the whole footer (image + attribution + role) only when attribution is present, so a blank attribution shows just the quote — no empty footer gap. No CSS changes needed; footer spacing was already flex `gap`, not a border/margin on a sibling.
- Updated the shared `assert_card_block` test helper (`test_blocks.py`) to handle the no-attribution case instead of assuming a `<cite>` always exists.

## Testing
- `pytest springfield/cms/tests/test_blocks.py -k "card or testimonial"` — 32 passed.
- Manually confirmed in admin: clearing Attribution saves fine, and the frontend renders the quote with no leftover footer space.

### DIFF
--- a/springfield/cms/blocks.py
+++ b/springfield/cms/blocks.py
@@ -2187,7 +2187,11 @@ def StepCardListBlock(allow_uitour=False, *args, **kwargs):
 
 class CardTestimonialBlock(blocks.StructBlock):
     content = RichTextBlock(features=HEADING_TEXT_FEATURES)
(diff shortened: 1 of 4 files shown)


janriokrause (approved):
Nice!

PR #1731, opened 2026-08-14
WT-1281: add waffle switch to force Nightly on the referral get-firefox page (desktop + Android)

stevejalim:
## Summary
- Adds a `REFERRAL_FORCE_NIGHTLY_QA` waffle switch so QA can force the Firefox referral program's `/get-firefox/?invitation=<code>` page to download Nightly instead of Release, without waiting for a Release ship.
- **Desktop (Windows/Mac):** read once server-side in `ReferralGetFirefoxPage.get_context()` and threaded down through the existing referral CTA render chain. `download_firefox_thanks_link()` gained a `channel="release"` kwarg (previously hardcoded), and the `download-firefox-button` component now accepts an optional `channel` prop defaulting to `"release"` so its other existing call sites are unchanged.
- **Android:** when the switch is active, the Play Store badge links to `org.mozilla.fenix` (Firefox Nightly) instead of `org.mozilla.firefox` (Release). The JS-side referral attribution now preserves whichever package id the server rendered instead of unconditionally overwriting it back to Release: `MobileAttribution.getStoreUrl` gained an optional `packageId` parameter, and `ReferralAttribution.rewriteAndroidLinks` reads each badge's existing href to extract the package id before rebuilding the URL.
- **iOS:** stays on Release since there is no public Nightly listing on the App Store.
- The page carries its normal ~10 minute CDN cache TTL; baking `switch()` straight into a cached template like this is an existing, accepted convention elsewhere in the codebase (e.g. `download_as_default`, `sentry-js`), so no new cache-bypass or live-flag-delivery infrastructure was introduced.

## Test plan
- [x] Python: `test_helpers.py::TestDownloadFirefoxThanksLink` -- default release channel, and `channel="nightly"` override for both a stub-installer platform (win) and a non-stub platform (osx)
- [x] Python: `test_referral_pages.py` -- `get_context()` channel defaults to release / forced to nightly by the switch, plus end-to-end render checks that both the desktop `.download-link` and Android `.fl-store-button-android` badge switch to Nightly when the switch is active
- [x] JS: `mobile-attribution.js` -- `getStoreUrl` uses a custom `packageId` when given, defaults to Release when omitted, ignores it for iOS
- [x] JS: `referral-attribution.js` -- `getPackageIdFromHref` parses the id param from Play Store URLs, and `rewriteAndroidLinks` preserves a server-rendered Nightly package id
- [x] Full Python suite (2131 tests) and JS suite (619 specs) pass
- [x] `ruff check` / `ruff format` / `eslint` / `prettier` pass
- [ ] Manually verify in a deployed environment: flip `REFERRAL_FORCE_NIGHTLY_QA` on via Wagtail/Django admin, load `/en-US/get-firefox/?invitation=<code>`, confirm desktop buttons point at a Nightly Bouncer URL and Android badge points at `org.mozilla.fenix`; flip off and confirm reversion to Release

### DIFF
--- a/media/js/base/mobile-attribution.js
+++ b/media/js/base/mobile-attribution.js
@@ -60,16 +60,21 @@ if (typeof window.Mozilla === 'undefined') {
      * @param {String} [referralContent] - Optional utm_content for referral attribution
      *   (e.g. "fxrefer1HR4FZ672Z8Y0E4HW"). When provided, appended to the Android
      *   Play Store referrer string. Has no effect on iOS.
+     * @param {String} [packageId] - Android Play Store package id, e.g.
+     *   "org.mozilla.firefox" (default) or "org.mozilla.fenix" (Nightly).
+     *   Has no effect on iOS.
      */
     MobileAttribution.getStoreUrl = function (
         campaign,
         isAndroid,
-        referralContent
+        referralContent,
+        packageId
     ) {
         var encoded = encodeURIComponent(campaign);
         if (isAndroid) {
             var url =
-                'https://play.google.com/store/apps/details?id=org.mozilla.firefox' +
+                'https://play.google.com/store/apps/details?id=' +
+                (packageId || 'org.mozilla.firefox') +
                 '&referrer=utm_source%3Dwww.firefox.com%26utm_medium%3Dreferral' +
                 '%26utm_campaign%3D' +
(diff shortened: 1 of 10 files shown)


stephaniehobson (approved):
R+

Issue #1726, opened 2026-08-12
ESR–next downloads are offered for NLA linux32 builds


janbrasna:
## Description

Since #466 the i686 builds are no longer provided. That means the ESR download page for linux32 can only provide 140esr at this point, not 153esr, as i686 builds stopped at around 145 cycle.

### Steps to reproduce

https://www.firefox.com/download/all/desktop-esr/linux/en-US/ — click "Download 153.0"

### Expected result

No "Download 153" button since that arch's no longer built.

### Actual result

404

### Environment

prod

This is also the reason for failing download tests, which are collected from the actually rendered buttons, i.e. the same 404s as end users see above: [/actions/runs/29841568361/job/88671526561](https://github.com/mozmeao/springfield/actions/runs/29841568361/job/88671526561) ❌

PR #1732, opened 2026-08-14
Fix download/all linux32 esr-next 404s

janbrasna:
## One-line summary

Skips ESR–next condition if linux i686, which no longer ships.

There are generally version constraints for product details, however this ESR branch overlap seems to be bolted on top. There might be more systemic solution somewhere, however once 140esr is over in a few months this while arch code can be removed from all places, so it doesn't have to be pretty or smart for now.

## Issue / Bugzilla link

#1726

## Testing

http://localhost:8000/en-US/download/all/desktop-esr/linux/af/ (only single ESR, no 404)
http://localhost:8000/en-US/download/all/desktop-esr/linux64-aarch64/af/ (control, 2 ESRs)
http://localhost:8000/en-US/download/all/desktop-esr/win/af/ (control, 3 ESRs)

### DIFF
--- a/springfield/firefox/templates/firefox/all/download-flare.html
+++ b/springfield/firefox/templates/firefox/all/download-flare.html
@@ -39,7 +39,7 @@
       {# ESR #}
       {% set is_esr_next_version = desktop_esr_next_version %}
       {# If there are multiple ESR's available, show next as well. #}
-      {% if is_esr_next_version %}
+      {% if is_esr_next_version and download_esr_next_url %}
         <p class="c-step-download">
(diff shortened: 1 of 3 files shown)


stevejalim (approved):

janbrasna:
@stevejalim Would you want me to submit this to bedrock first, so it's deployed to prod there ahead of this, as the lower traffic download source? Or you have a convenient time frame in mind for this?

stevejalim:
@janbrasna shipped. Feel free to PR for Bedrock too. Thanks, dude!

PR #1738, opened 2026-08-14
Stop conditional variants getting different heading levels

slightlyoffbeat:
## Summary

Assign heading levels in Python instead of a Jinja counter, so conditional variants of the same block share a level rather than the second one getting an `<h2>`.

## Key changes

- New `springfield/cms/block_slots.py` + `BlockSlotsMixin` replace the `namespace(headings=0)` counter that was copy-pasted across 14 loops in 11 templates. Per-page config (`slot_region_groups`, `slot_flat_levels`, `slot_start_level`) reproduces each page's existing levels — the download page's blocks start at h2, the roadmap's intro owns the h1.
- Conditional blocks that genuinely exclude each other now share one logical position, and therefore one heading level.
- Exclusivity is inferred conservatively: blocks must differ on **exactly one** condition with values that can't both match. `platforms=[windows]` next to `firefox=is-firefox` are both visible to a Windows Firefox user, so they stay sequential.
- Headingless blocks no longer inherit the page's level — a `line_cards` block with no heading of its own was rendering `<h1 class="fl-heading-size-3">` per card.
- No migrations; the custom `StreamField` keeps block changes out of them.

## Behaviour matrix

| Blocks in a region | Levels before | Levels after |
|---|---|---|
| Three plain blocks | h1, h2, h2 | unchanged |
| `is-firefox` + `not-firefox` + plain | h1, **h2**, h2 | h1, **h1**, h2 |
| `platforms=[windows]` + `is-firefox` (both can show) | h1, h2 | unchanged |
| Headingless block first | inherits h1 | block template's own default |
| Five platform sections (Thanks page) | h1, h2, h2, h2, h2 | h1, h1, h1, h1, h1 |

## Testing

- `springfield/cms/tests/test_block_slots.py` — 23 unit tests (heading detection, level assignment, per-page config, the exclusivity rule) plus 3 rendering tests.
- The ~25 existing tests in `test_blocks.py` that re-implement the old rule inline pass **untouched** — the behaviour-preservation check.
- Replayed the old rule against every page in a local DB snapshot and diffed: the only real level changes are 17 Thanks pages and one What's New page, all conditional variants collapsing as intended.
- Suite: 1632 passing. Five failures pre-existing on this branch.

## Notes for reviewer

- **The Thanks page changes visibly.** Its five platform sections are exclusive, so only one renders per visitor — Windows users saw `<h1>` at 80px, everyone else `<h2>` at 64px. All platforms now get 80px. That's the platforms being made consistent, not a redesign; `section.html` doesn't override `heading_size`, so the size class follows the level.
- `__str__` added to `ReferralHubPage` and `ReferralGetFirefoxPage` — ruff flagged them once the new base made them resolvable as models.
- Follow-up worth having: an explicit heading-level override in page Settings, per Kasey's suggestion, for variants the rule can't infer (e.g. separated by another block).

### DIFF
--- a/springfield/cms/block_slots.py
+++ b/springfield/cms/block_slots.py
@@ -1,0 +1,168 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Heading levels for the top-level blocks of a page region.
+
+Conditional blocks are mutually exclusive alternatives that occupy one position, but
+they are authored as consecutive siblings. Assigning heading levels by document order
+therefore hands the second variant an <h2>, so a visitor who sees that one gets a page
+whose top heading is an <h2> and whose <h1> is hidden.
+
+Templates used to count headings inline with a Jinja namespace, once per region and
+fourteen times over. Doing it here means one implementation and one place to fix.
+"""
+
+from dataclasses import dataclass
+from typing import Any
+
+
+def has_heading(block) -> bool:
+    """Whether a block contributes a heading, mirroring what the templates rendered.
+
+    Blocks nest a HeadingBlock under `heading`, or carry a flat `headline`. Values that
+    are neither (rich text, lists) have no heading and no `.get`.
+    """
+    value = getattr(block, "value", None)
+    if not hasattr(value, "get"):
+        return False
+
+    heading = value.get("heading")
+    if hasattr(heading, "get") and heading.get("heading_text"):
+        return True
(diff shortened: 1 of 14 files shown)


lucianoratamero:
hey 👋 
I've talked with Mari about this, and I think we have a simpler solution to propose for the issue.
can we keep this PR on hold while I work on a separate solution, or do you need this merged now?

slightlyoffbeat:
@lucianoratamero this was only a proposal. If you have something else that will be better, feel free to close this one in favor of that one

maribedran:
I'll go ahead and close this, since we're going with the simpler solution.

PR #1745, opened 2026-08-17
Add analytics tracking to share buttons [WT-1644]

stephaniehobson:
## One-line summary

Track referral share buttons as `widget_action` in analytics.

## Testing

- Open the referral share page (steps below if you don't already have one locally)
    - Pull down the production database locally
    - Open nightly
    - Turn your VPN to US
    - Pick "Share Nightly" to open the referral page.
    - Change the domain to localhost:8000
- Click each button
- Log the dataLayer in your console and see that there is an event for each one

### DIFF
--- a/media/js/cms/components/flare-referral-controls.es6.js
+++ b/media/js/cms/components/flare-referral-controls.es6.js
@@ -1,0 +1,58 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+function initReferralControls(containerEl) {
+    const emailEl = containerEl.querySelector(
+        '.fl-referral-controls-share-email'
+    );
+    const qrEl = containerEl.querySelector('.fl-referral-controls-qr-button');
+
+    const clipboardEl = containerEl.querySelector(
+        '[data-js="fl-copy-to-clipboard"]'
+    );
+
+    if (clipboardEl) {
+        clipboardEl.addEventListener('click', () => {
+            if (window.dataLayer) {
+                window.dataLayer.push({
+                    event: 'widget_action',
+                    type: 'referral share',
+                    action: 'copy link',
+                });
+            }
+        });
+    }
+
+    if (emailEl) {
+        emailEl.addEventListener('click', () => {
+            if (window.dataLayer) {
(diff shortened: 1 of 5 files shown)


media/js/cms/components/flare-referral-controls.es6.js
```diff
@@ -0,0 +1,58 @@
...
+        '.fl-referral-controls-share-email'
+    );
+    const qrEl = containerEl.querySelector('.fl-referral-controls-qr-button');
+
+    const clipboardEl = containerEl.querySelector(
+        '[data-js="fl-copy-to-clipboard"]'
+    );
+
+    if (clipboardEl) {
+        clipboardEl.addEventListener('click', () => {
```
wen-2018: Should we track successful copy with writeText().then() instead? Currently, if an icon/a label is missing, the copy action will fail but click will still happen.

media/js/cms/pages/flare-referral-hub.es6.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+import setupReferralControls from '../components/flare-referral-controls.es6';
+
+setupReferralControls();
```
wen-2018: It could fail silently if the parser hasn't reached those elements yet when the events fire
```suggestion
if (document.readyState === 'loading') {
    document.addEventListener('DOMContentLoaded', setupReferralControls);
} else {
    setupReferralControls();
}
```

wen-2018 (changes requested):

media/js/cms/components/flare-referral-controls.es6.js
stephaniehobson: ✅

media/js/cms/pages/flare-referral-hub.es6.js
stephaniehobson: ✅

stephaniehobson:
I edited the clipboard script to emit an event we can tie into that's more dependable than click.

I also edited the script to emit a generic event everywhere we use the script.

That means duplicate events on this page but I think that's worth it to group the `referral share` events together.

wen-2018 (approved):
r+wc

The copy-to-clipboard event is firing twice. Since this is intentional, a comment might be helpful

PR #1747, opened 2026-08-18
Validate Contact Page fields according to Basket API fields

maribedran:
## One-line summary

Only allow the Contact Page to use valid Basket API endpoints and validate the form fields according to what the endpoint accepts.

- Restrict the values allowed for the `basket_api_path` field (currently only one endpoint is available)
- Raise errors when a Contact Page's form fields don't match the fields accepted by the Basket API

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1604

## Testing

Edit any contact page and try to:
- Remove a field
- Add a field with an internal identifier not accepted by the Basket API
- Uncheck the required option in any of the required fields


springfield/cms/models/pages.py
```diff
@@ -2236,6 +2235,39 @@ def __str__(self):
...
+            "opt_in",
+            "lead_source",
+            "cta",
+            "message",
+        },
+    },
+}
+
+BASKET_API_PATH_CHOICES = [(path, path) for path in BASKET_ENDPOINT_FIELDS]
+
```
stevejalim: nit: this file is getting very big - might be better to move these to `springfield.settings` or to a new `cms.constants`  module?

springfield/cms/models/pages.py
```diff
@@ -2303,9 +2335,8 @@ class ContactPage(PageThemeMixin, AbstractSpringfieldCMSPage):
     basket_api_path = models.CharField(
         max_length=255,
         blank=True,
-        help_text=(
-            "Basket API path (e.g. /api/v1/contact/). Concatenated with settings.BASKET_URL on submission. Required if Email Address is not set."
-        ),
+        choices=BASKET_API_PATH_CHOICES,
```
stevejalim: This change will mean the form needs to be re-saved to pick up the new API path, yep?

stevejalim (approved):
r+wc plus a question:

springfield/cms/models/pages.py
maribedran: No, that's already the path that we're using. It will just prevent it from being changed.

springfield/cms/models/pages.py
maribedran: That's been bothering me too. Moving out the blog will help a bit. I wonder if there are other page types that should go into different apps.

PR #1748, opened 2026-08-18
chore: reduce Release GHA noise in #www - point to #www-notify instead

stevejalim:
## One-line summary

This changeset reduces noise in our #www Slack channel by only notifying for a Release start and its finish. The other notifications are sent to #www-notify instead

## Testing

Sense-check on its own is OK for now


janriokrause (approved):
gr8️⃣! 🤖

PR #1749, opened 2026-08-18
[WT-1423] Make Jinja IncludeContentsExtension thread safe

### LINKED ISSUE #1590: Investigate & Resolve Errors in Content Loading in Wagtail Dev
Per @Luciano Ratamero:  
The issue is that some content is either loaded in the wrong places or not loaded at all. That only happens in Dev, and only if we bombard the local server with multiple requests concurrently. That's been affecting the way we're running Playwright tests (mostly visual regression ones), because we can't run the tests in parallel, or they randomly fail. The way to reproduce this is to open any page (the home page will do), and hold the keyboard keys to reload the page for a few seconds, then let it go. Since the browser will do a lot of requests, the issue always happens.

I'm 100% available to have a quick meeting about it with whoever takes this task. I've already dug a bit, trying to fix it, so I have a lot of context about things I tried, and some technical specificities about it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/WT-1423)
┆Epic: (None)

James-Osgood:
## One-line summary

Fix thread-unsafe rendering bug in `django-includecontents` causing content to load incorrectly under concurrent requests.

The `django-includecontents` package's Jinja2 extension (`includecontents.jinja2.IncludeContentsExtension`) has a bug: its plain list rendering stack (`self._render_stack`) is shared between threads because Jinja2 instantiates the extension once per environment. This resulted in slot content in `<include:..>` tags sometimes receiving content from other components or missing altogether.

This PR fixes the bug by subclassing `IncludeContentsExtension` and replacing the render stack with a thread-safe [`ContextVar`](https://docs.python.org/3/library/contextvars.html).

See SmileyChris/django-includecontents#11 for additional details. See SmileyChris/django-includecontents#12 for the fix in that library. (The fix differs from the one presented in this PR due to the ease of fixing the bug via subclassing the extension with an attribute versus rewriting several methods; both changes fix the reported issue.) Once `django-includecontents` releases a new version with this bug fixed, and this repo updates its dependency, then this PR can be reverted.

Note: this should unblock parallel Playwright testing.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1423

Closes #1590.

## Testing

1. Open several tabs (either site pages rendered by the CMS and/or pattern library views). Five should do.
2. Make a change to a CSS file, causing webpack to reload all of the tabs.
3. Check that all content was loaded in as expected: no empty blocks, or blocks with content from other pages.

### DIFF
--- a/springfield/base/tests/test_jinja_extensions.py
+++ b/springfield/base/tests/test_jinja_extensions.py
@@ -1,0 +1,128 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import threading
+
+from jinja2 import DictLoader, Environment
+
+from springfield.jinja2 import ThreadSafeIncludeContentsExtension
+
+COMPONENTS = {
+    "components/box.html": "[{{ name }}|{{ contents.slot }}]",
+    "components/outer.html": "[outer|{{ contents.slot }}]",
+}
+
+
+def build_environment(extension):
+    return Environment(loader=DictLoader(COMPONENTS), extensions=[extension])
+
+
+# FUTURE: remove when django-includecontents releases a new version which fixes
+#         django-includecontents#11.
+class TestThreadSafeIncludeContentsExtension:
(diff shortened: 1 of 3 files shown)


James-Osgood:
Superseded by #1775.

PR #1751, opened 2026-08-19
[WT-1638] Ensure minor version component is included in WNP release notes link

knowler:
## One-line summary

For non-CMS What’s New pages, the release notes link was broken for major versions since the minor component (i.e. `.0`) was missing from the URL. This ensures that’s included.

Full disclosure, I used Claude Code for this change as I wasn’t totally familiar with this part of the codebase, the helpers that were available, and testing with Python. I reviewed the code and tested it works in the browser and with the automated tests.

## Issue / Bugzilla link

- https://mozilla-hub.atlassian.net/browse/WT-1638
- https://bugzilla.mozilla.org/show_bug.cgi?id=2063623

## Testing

Pull down and test that the release notes link in the page content goes to the correct place (e.g. link href is set to `/sv-SE/firefox/153.0/releasenotes/` — will redirect to the en-US version):

- Previously broken: http://localhost:8000/sv-SE/whatsnew/153/
    - Try a few different languages (e.g. `cy`)
- Not previously broken, but could be affected by the change (i.e. it’s in the code path): http://localhost:8000/sv-SE/whatsnew/153.0beta/

### DIFF
--- a/springfield/firefox/templates/firefox/whatsnew/evergreen.html
+++ b/springfield/firefox/templates/firefox/whatsnew/evergreen.html
@@ -90,7 +90,7 @@
     </div>
 
     <aside class="mzp-l-content c-utilities">
-      {% set releasenotes_url = url('firefox.releases.index') if not version else '/firefox/%s/releasenotes/'|format(version) %}
+      {% set releasenotes_url = url('firefox.desktop.releasenotes', releasenotes_version, 'release') if releasenotes_version else url('firefox.releases.index') %}
       {% set releasenotes_link = 'href="%s" target="_blank"'|safe|format(releasenotes_url) %}
       <p>{{ ftl('whatsnew-release-notes', url=releasenotes_link) }}</p>
     </aside>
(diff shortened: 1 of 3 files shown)


springfield/firefox/templates/firefox/whatsnew/evergreen.html
```diff
@@ -90,7 +90,7 @@ <h1 class="wnp-main-title">{{ ftl('whatsnew-download-firefox-for-your') }}</h1>
     </div>
 
     <aside class="mzp-l-content c-utilities">
-      {% set releasenotes_url = url('firefox.releases.index') if not version else '/firefox/%s/releasenotes/'|format(version) %}
+      {% set releasenotes_url = url('firefox.desktop.releasenotes', releasenotes_version, 'release') if releasenotes_version else url('firefox.releases.index') %}
```
knowler: I think the else branch here might never be followed since `releasenotes_version` is always set?

https://github.com/knowler/springfield/blob/52f8597d6d21eae5177cfd3bde06f031a4b9637f/springfield/firefox/views.py#L1067-L1069

stevejalim (approved):
Thanks @knowler! Test driven and LGTM!

PR #1753, opened 2026-08-24
Security dep bumps: FE and BE

stevejalim:
Python and JS dep bumps, following Dependabot pings about them 

Integration tests passing: https://github.com/mozmeao/springfield/actions/runs/32725504348


wen-2018 (approved):
r+

PR #1757, opened 2026-08-24
WT-1543: Blog index styles

lucianoratamero:
## One-line summary

This PR implements all styles needed for the blog index page, and double checks other pages, like the `/all/` and `/topics/` pages.

- **Removed dates and tags from article list items** on the index, all and article pages (not present in Figma). This also removed the localized-tag-name test, since tags are no longer rendered in those lists.
- Blog CSS consolidated/simplified and colors + spacing normalized; shared `fl-blog-body` class added to all blog page templates.
- The topics list carousel was kept as is, since it looks very misaligned on figma, and it works both functionally and visually as-is.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1543

## Testing

Run the fixtures, then manually check the blog index, `/all`, topic filter and article pages in light and dark themes.


springfield/cms/templates/cms/blocks/blog-latest-articles-section.html
```diff
@@ -0,0 +1,71 @@
...
+          alignment_class="text-center"
+        />
+        <div class="fl-blog-latest-grid">
+          {% for article in articles %}
+            {% set listing_image = article.get_listing_image() %}
+            <include:blog-article-list-item with_image="{{ listing_image }}">
+              <content:heading>
+                <hgroup class="fl-heading-group">
+                  {% set topic = article.get_topic() %}
+                  <p class="fl-superheading">{{ topic.name if topic else "" }}</p>
```
maribedran: This will render an empty p if there's no topic.

springfield/cms/templates/cms/blocks/blog-latest-articles-section.html
```diff
@@ -0,0 +1,71 @@
...
+                  <h3 class="fl-heading">
+                    <a href="{{ article.url }}" class="fl-link">{{ article.title }}</a>
```
maribedran: We constantly repeat this pattern when we need to wrap a heading with a link. Do you think it makes sense to pass a link prop to the heading component?

springfield/cms/templates/cms/blocks/blog-latest-articles-section.html
```diff
@@ -0,0 +1,71 @@
...
+                </hgroup>
+              </content:heading>
+              {% if listing_image %}
+                <content:image>{{ list_item_media(article) }}</content:image>
+              {% endif %}
+            </include:blog-article-list-item>
+          {% endfor %}
+        </div>
+        <div class="fl-buttons is-horizontal is-center">
+          <a class="fl-button button-ghost" href="{{ value.get_link_url() }}">
```
maribedran: Why not use the button component here?

springfield/cms/templates/cms/blocks/blog-latest-articles-section.html
```diff
@@ -0,0 +1,71 @@
...
+    <div class="fl-blog-latest-scallop-inner"></div>
+  </div>
+</div>
+{% elif is_preview %}
+  <div class="conditional-display is-preview">
+    <include:notification color="red">
+      <content:heading>This section is empty</content:heading>
+      <p>No articles meet this block's filter criteria. Articles that match the filters might already be listed above.</p>
+      <p>Heading: {{ value.heading_text|richtext|remove_p_tag }}</p>
+      <p>Filter: {{ value.source }}</p>
```
maribedran: There's no filter in this block.

springfield/cms/templates/cms/includes/blog/blog-featured-articles.html
```diff
@@ -57,7 +57,28 @@
...
+                {% set dark_mobile_image = block.value.get_mobile_dark_image() %}
+                {% set classes = image_variant_classes(dark_image, mobile_image, dark_mobile_image) %}
+                {% if image.focal_point_x %}
+                  {% set focal_point = (image.focal_point_x / image.width * 100)|round(2) ~ "% " ~ (image.focal_point_y / image.height * 100)|round(2) ~ "%" %}
+                {% else %}
+                  {% set focal_point = "center" %}
+                {% endif %}
+                <content:image>
+                  <div class="image-variants-display" style="--focal-point: {{ focal_point }}">
+                    {{ srcset_image(image, full_width_filterspec, sizes="(min-width: 768px) 50vw, (min-width: 1440px) 680px, 100vw", width=image.width, loading="lazy", class=classes.primary) }}
```
maribedran: The 768px and 1440px sizes are inverted here. This showed up in a previous PR review. I suspect we have other instances of that.

maribedran: The size could also be moved to a variable so we don't repeat it everywhere.

springfield/cms/templates/cms/blog_article_page.html
```diff
@@ -62,28 +62,14 @@
               <include:blog-article-list-item with_image="{{ listing_image }}">
                 <content:heading>
                   <hgroup class="fl-heading-group">
-                    <p class="fl-superheading">{{ article.topic.name }}</p>
+                    {% set topic = article.get_topic() %}
+                    <p class="fl-superheading">{{ topic.name if topic else "" }}</p>
```
maribedran: This will render an empty p if there's no topic.

springfield/cms/templates/cms/blog_index_page.html
```diff
@@ -43,25 +42,35 @@
...
+        {% for block in article_sections if block.block_type != "latest" %}
           {% include_block block %}
         {% endfor %}
-
       </div>
     </section>
+
+    {# The latest section is a full-width dark band, so it sits outside the container. #}
+    {% for block in article_sections if block.block_type == "latest" %}
+      {% if loop.index == 1 %}
```
maribedran: Won't this only render the block if it's the first one?

maribedran: Wagtail has `.blocks_by_name()`

springfield/cms/tests/test_blog_pages.py
```diff
@@ -947,8 +967,8 @@ def test_blog_all_renders_list_articles(blog_setup, rf):
         heading = item.find("h2", class_="fl-heading")
         assert heading and heading.find("a")
         assert item.find("div", class_="fl-body")
-        assert item.find("p", class_="fl-blog-article-date")
-        assert item.find("span", class_="fl-tag")
+        assert not item.find("p", class_="fl-blog-article-date")
```
maribedran: A test asserting that something doesn't exist doesn't make sense. The correct change is to just remove the assertion.

springfield/cms/tests/test_blog_pages.py
```diff
@@ -1556,8 +1576,8 @@ def test_blog_article_renders_related_articles(privacy_articles, rf):
         assert link and link["href"] == related.url
         body = item.find("div", class_="fl-body")
         assert body and body.get_text(strip=True)
-        assert item.find("p", class_="fl-blog-article-date")
-        assert item.find("span", class_="fl-tag")
+        assert not item.find("p", class_="fl-blog-article-date")
```
maribedran: Same here.

maribedran:
Images are right-aligned on the topic page and are too small. This looks like a side effect of some other change here.

image

lucianoratamero:
> Images are right-aligned on the topic page and are too small. This looks like a side effect of some other change here.
> image

this was on purpose. we don't have Figma designs for the topic pages (or I couldn't find any), so I just mirrored what we have for the designs of the "All topics" page.
it does clash with the left-aligned ones in the main page (the "latest posts" section, but there are a lot of these small inconsistencies that I tried to iron out. I'm totally open to discuss and modify them, though!

maribedran:
@lucianoratamero there aren't specific designs for it, but I'm assuming it should look the same as the index page, since it's the same layout.

maribedran:
The topics list isn't scrolling all the way to the beginning.

image

maribedran (approved):

PR #1744, opened 2026-08-17
Blog Index page sections and feed exclusions

maribedran:
## One-line summary

Allow filtering by tags; add feed exclusions to the Blog Index Page and create new blog article sections for the page.

## Significant changes and points to review

- Add a new `?tag=<slug>` filter to `/blog/all/` and `/blog/topic/<slug>/` listings
- Add a new Feed Exclusions field to the Blog Index page to exclude tags and topics from listings
  - The exceptions for the exclusions are listings filtering explicitly for the tag or topic
- Change the `BlogCardsListBlock` to list N articles from a tag or topic
- Create the `BlogLatestArticlesBlock` to list N most recent articles 
- Implement the Blog Index page logic for displaying articles in the sections:
  - Articles already shown in previous sections are excluded from each section filtering
  - Tags or topics from Feed Exclusions are excluded from sections, unless the section explicitly filters by them
- Refactor the Blog Index page views so each route only builds the context it needs for its own rendering

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1494
https://mozilla-hub.atlassian.net/browse/WT-1495

## Testing

Run `./manage.py load_page_fixtures` and check http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/

PR #1760, opened 2026-08-25
Make 0144 bottom_banner migration idempotent

stevejalim:
PR #1744 renumbered the migration in this changeset from `0141` to `0144`, but it's already been applied in Dev, Stage and Prod.

Databases that already applied the old 0141 have the column but not the new migration record, so Django tries to `AddField` again and hits `DuplicateColumn` whicih causes an error running migrations and kills the rollout of the new deployment.

Hopefully, we can fix this by using `SeparateDatabaseAndState` with `ADD COLUMN IF NOT EXISTS` – that will make it safe for the deployed envs with the old migration number and also for any fresh runs of recent migrations (eg demo servers that haven't been used for a while, or the sqlite db export tool)

### DIFF
--- a/springfield/cms/migrations/0144_blogarticlepage_bottom_banner.py
+++ b/springfield/cms/migrations/0144_blogarticlepage_bottom_banner.py
@@ -9,27 +9,42 @@ from django.db import migrations
 import springfield.cms.fields
 
 
+class AddFieldIfColumnExists(migrations.AddField):
+    """
+    AddField that skips schema application when the column is already there.
+
+    This migration was applied to some databases while it was numbered 0141,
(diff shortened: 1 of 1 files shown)


James-Osgood:
[Tests failed](https://github.com/mozmeao/springfield/actions/runs/32897756281/job/97964216294?pr=1760) due to SQLite not having `IF EXISTS` syntax.

I refactored the `SeparateDatabaseAndState` into an `AddField` subclass which checks whether the column exists before applying the schema change. This handles the different edge cases while being SQL-dialect agnostic.

James-Osgood (approved):

PR #1762, opened 2026-08-26
[WT-1652] Draft sharing link for translated pages

James-Osgood:
## One-line summary

Add a draft sharing link (publicly accessible) to the page action menu of translated pages.

The [`wagtail-localize` library](https://wagtail-localize.org/) renders translated pages' action menu with a React app independent of the [`register_page_action_menu_item` Wagtail hook](https://docs.wagtail.org/en/7.1/reference/hooks.html#register-page-action-menu-item), so the "Create draft sharing link" button implemented by the [`wagtaildraftsharing` library](https://github.com/torchbox/wagtaildraftsharing/) are not automatically added to the menu of translated pages. Furthermore, translated pages do not have `Revision` objects to build draft sharing links off of.

This PR adapts `wagtaildraftsharing` to `wagtail-localize`'s translated pages:
- `edit_translation.html` template extended to add a "Create draft sharing link" button to translation pages via a template tag.
- JS to move the button to the page's action menu and implement its async request and response handling.
- A new `cms_translation_draftsharing_create` view to create a revision for the translated page and return a `WagtaildraftsharingLink` for it (re-using `wagtaildraftsharing` functionality).
  - The view cleans up expired/inactive links (and their revisions) for the page. Revision deletions are restricted to those created solely for draft sharing.

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1652

## Testing

image

- Go to a translated page that has unpublished edits.
- Click on the "Create draft sharing link" button in the action menu.
- Open the copied URL in a new browsing context.
- Verify the page is a preview of the translated page with the unpublished edits.

Clean up inactive links:
- Go to ["Sharing Links"](http://localhost:8000/cms-admin/wagtaildraftsharing/).
- Mark the translated page's sharing link as inactive.
- Click on the "Create draft sharing link" button in the translated page's action menu again.
- Verify a new link was generated for the page and the old link was deleted.

Clean up expired links:
- Go to ["Sharing Links"](http://localhost:8000/cms-admin/wagtaildraftsharing/).
- Mark the translated page's sharing link as expired ("Active until" in the past).
- Click on the "Create draft sharing link" button in the translated page's action menu again.
- Verify a new link was generated for the page and the old link was deleted.

### DIFF
--- a/media/js/cms/wagtailadmin-translation-draftsharing.es6.js
+++ b/media/js/cms/wagtailadmin-translation-draftsharing.es6.js
@@ -43,6 +43,15 @@ export async function createSharingLink(button) {
         if (!response.ok) {
             labelNode.nodeValue = originalLabel;
             button.disabled = false;
+            document.dispatchEvent(
+                new CustomEvent('w-messages:add', {
+                    detail: {
+                        clear: true,
+                        text: 'Error generating sharing URL',
+                        type: 'error'
+                    }
+                })
+            );
             return;
         }
 
@@ -70,6 +79,15 @@ export async function createSharingLink(button) {
     } catch (error) {
         labelNode.nodeValue = originalLabel;
         button.disabled = false;
+        document.dispatchEvent(
+            new CustomEvent('w-messages:add', {
+                detail: {
+                    clear: true,
+                    text: 'Error generating sharing URL',
+                    type: 'error'
+                }
+            })
+        );
     }
 }
 
--- a/media/js/cms/wagtailadmin-translation-draftsharing.es6.js
+++ b/media/js/cms/wagtailadmin-translation-draftsharing.es6.js
@@ -131,7 +131,7 @@ export function initTranslationDraftsharing(options) {
     return observer;
 }
 
-// Auto-run only when there is a button (so there is something to share). Skips otherwise.
+// Auto-run only when there is a button. Skips otherwise.
 if (typeof document !== 'undefined' && document.getElementById(TEMPLATE_ID)) {
     if (document.readyState === 'loading') {
         document.addEventListener('DOMContentLoaded', function () {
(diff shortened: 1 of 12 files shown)


stevejalim (approved):
Thanks for taking this on. 

I haven't locally test driven, but I know the code as I brought it up to match modern Wagtail before we adopted using it, and LGTM.

I'm happy with us merging this as-is, but I wondered what you think about actually folding it into `wagtaildraftsharing` as a contribution... 

(I can re-fork the repo to `mozmeao/wagtaildraftsharing` and we can use that fork until it gets accepted back upstream. But I think the key conversation around that idea is whether it'd be straightforward to only enable this new behaviour if `wagtail-localize` is in `INSTALLED_APPS` -- which seems viable at first blush)

James-Osgood:
> I'm happy with us merging this as-is, but I wondered what you think about actually folding it into `wagtaildraftsharing` as a contribution...
> 
> (I can re-fork the repo to `mozmeao/wagtaildraftsharing` and we can use that fork until it gets accepted back upstream. But I think the key conversation around that idea is whether it'd be straightforward to only enable this new behaviour if `wagtail-localize` is in `INSTALLED_APPS` -- which seems viable at first blush)

It is definitely possible, probably even straightforward. I was actually considering this as a standalone adapter package, rather than part of `wagtaildraftsharing`. But it would be more convenient if it was included there. (Despite being a lot of conditional functionality to support for an optional sibling dependency.)

Integrating the JS would be an interesting challenge, as I think we'd want a single script to handle the link behaviour, rather than have parallel implementations. The backend code would be fairly simple to conditionally include. The new view doesn't match `wagtaildraftsharing`'s pattern for its other views though - that's less straightforward to adapt, so I'd prefer to leave as-is. However, I'd consider dropping the automatic deletion of stale links/revisions, since there's nothing in `wagtaildraftsharing` with similar behaviour; a management command would be a better place to do that cleanup.

If you think `wagtaildraftsharing` has a chance of accepting the support for adapting `wagtail-localize`, then I'm willing to give it a shot!

maribedran (approved):

stevejalim:
> If you think wagtaildraftsharing has a chance of accepting the support for adapting wagtail-localize, then I'm willing to give it a shot!

Definitely. I've given you write access to our fork: github.com/mozmeao/wagtaildraftsharing

However, in the meantime, let's get the benefit into Springfield, see if we need to adjust anything, then consider the refactor/change to upstream

PR #1769, opened 2026-08-27
feat: add 'ESR' suffix to links on release notes when it's for an ESR version

stevejalim:
This changeset shows "ESR" in the link labels when the release notes being linked to are for ESR.

For example, see FF 140

Screenshot 2026-08-27 at 17 15 21

This assumes that all ESR versions have a non-zero value for Y in the X.Y.Z versioning pattern, which does also match the data we get from releasse-notes.

There is also support for exceptions like 50.1.0 which was not ESR.

The list comprehension is pretty intense, but breaks down fine when you walk through it slowly.

BZ ticket https://bugzilla.mozilla.org/show_bug.cgi?id=2066059

## Testing

0. Start on `main`, not this branch
1. Look the 140.x release at http://localhost:8000/en-US/releases/ 
2. Confirm that the 140.0.X pills link to FF Release relnotes
3. Confirm that 140.1.X pills link to ESR notes but the pill has no visual indication
4. Switch to this branch, restart Django, reload
3. Confirm that 140.1.X pills now mention "ESR" in their label and still correctly link to ESR notes

### DIFF
--- a/springfield/firefox/templates/firefox/releases/index.html
+++ b/springfield/firefox/templates/firefox/releases/index.html
@@ -38,7 +38,7 @@
         {% if versions.minor -%}
         <ol class="fl-c-releases-list-minor">
           {% for version in versions.minor -%}
-          <li>{{ get_flare_link(int_version, version, 'secondary') }}</li>
+          <li>{{ get_flare_link(int_version, version.version_string, 'secondary', version.is_esr) }}</li>
           {% endfor -%}
         </ol>
         {% endif -%}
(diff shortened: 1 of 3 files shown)


stephaniehobson (approved):
R+ I think this is a good change and we can roll forward with refinements.

The only thing concerning me is 91, 112, 128 (and maybe others) are listing some releases twice. This smells more like a problem with the data, but I didn't dig into it. It doesn't seem to be a problem on 140 or 153 so I'm not inclined to block on it.

Image

stephaniehobson:
Assigning this back to Steve for him to evaluate the co-pilot feedback.

stevejalim:
> R+ I think this is a good change and we can roll forward with refinements.
> 
> The only thing concerning me is 91, 112, 128 (and maybe others) are listing some releases twice. This smells more like a problem with the data, but I didn't dig into it. It doesn't seem to be a problem on 140 or 153 so I'm not inclined to block on it.
> Image

yeah i think we will have to live with this for now and tune it - i suspect it's data, yeah

stevejalim:
> R+ I think this is a good change and we can roll forward with refinements.
> 
> The only thing concerning me is 91, 112, 128 (and maybe others) are listing some releases twice. This smells more like a problem with the data, but I didn't dig into it. It doesn't seem to be a problem on 140 or 153 so I'm not inclined to block on it.

Not the data - was a bug in how ESR versions were normalized. Fixed

PR #1771, opened 2026-08-28
Fix routing blowups with empty db

stevejalim:
This changeset properly addresses issues with being able to bootstrap an empty DB without increasingly gruubby workarounds. The need for workarounds came because the a few things were explicitly or implicitly touching the real DB before it had been fullly refilled with default content - as such the contnettypes table was empty and Django would error out.

The areas fixed in this PR are:

1. `get_routing_trigger` would inadvertently need an instance of a Page model during Wagtail's bootup, which caused a contenttypes table hit and/or automatic row addition (depending on the situation), which would blow up. We had got around this for Sqlite exports by using the SQLITE_EXPORT_MODE to branch its code, but then the need to reset demo databases raised a _similar_ gotcha, and adding a second special condition was getting grim. Because the routing triggers don't actually _need_ a live instance at startup, the fix was to make it a a classmethod, rather than an instance method.

2. There were two dead data migrations that broke on a fresh migrate (springfield/cms/migrations/0108_*.py, 0109_*.py)
Both called management commands that query django.apps.apps (the live, current-code model registry) instead of the historical registry a migration should use. I've commented these out now.

3. SIDE EFFECT WIN! For ages we've had warnings in logs about touching the DB too early:
```/venv/lib/python3.13/site-packages/django/db/backends/utils.py:98: RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.``` This was because of work done to `lazy_langs` a year or two ago. The warning became distracting (and a false clue) for humans and LLMs alike, so it was time to fix it: now we return lang codes as display names during startup, then the the `lazy()` proxy re-evaluates settings.LANGUAGES on every access, so real native names are returned as soon as apps are ready. Phew. Glad to see the back of that noise

## Testing

Test drive locally; confirm runserver works; confirm the language picker at the bottomo of the page still has native names


janriokrause (approved):
Great to see that `RuntimeWarning` going away 🌻

PR #1773, opened 2026-08-28
[WT-1550] "Recommended Articles" on blog posts

James-Osgood:
## One-line summary

Replaces the automatic "Related Articles" section at the bottom of blog post pages with a "Recommended Articles" section that is populated by the editor through a new "Recommended Articles" edit tab.

- A new "Recommended Articles" tab on Blog Article edit forms with two fields:
  - "Hide recommended" checkbox
  - "Recommended articles" streamfield with up to 4 articles
- The "Related Articles" section at the bottom of blog post pages is replaced by a "Recommended Articles" section that displays up to 4 articles selected in this order:
  - From the "Recommended articles" field
  - Matching both topic and at least one tag
  - Matching topic only 
  - Matching tag only
- Recommended articles are not repeated
- An article cannot recommend itself
- "Recommended Articles" section is hidden if "Hide recommended" is checked

## Issue / Bugzilla link

https://mozilla-hub.atlassian.net/browse/WT-1550

## Testing

Run `./manage.py load_page_fixtures`.

Check http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/test-regular-blog-article-4/. It will have 3 automatically recommended articles:
- `test-regular-blog-article-9` matching by topic+tag
- `test-featured-blog-article-4` matching by topic only
- `test-privacy-extra-regular-4` matching by tag only

Check http://localhost:8000/en-US/flare-docs/sample-pages/test-blog-index/test-bottom-banner-article/. It will have 2 custom recommended articles and 2 automatic ones:
- `test-regular-blog-article-3` custom     
- `test-regular-blog-article-4` custom
- `test-listing-image-article` matching by topic+tag 
- `test-hero-video-article` matching by topic+tag

Edit the `/test-bottom-banner-article/` page:
- go to its "Recommended Articles" tag.
- confirm the 2 custom recommended articles
- check "Hide recommended"
- publish
- reload the page and confirm the "Recommended Articles" section is hidden


springfield/cms/tests/test_blog_article_recommendations.py
```diff
@@ -0,0 +1,331 @@
...
+
+    assert get_article_recommendations() == [
+        "Topic and tag",
+        "Older topic and tag",
+        "Topic only",
+        "Tag only",
+    ]
+
+
+def test_default_recommendations_with_no_topic(article, get_article_recommendations, recommendation_pool):
```
maribedran: This test is confusing. It says there's no topic, but the assertion still mentions topics. I usually find tests that create things explicitly more readable, even when they're more verbose. What do you think?

l10n/en/cms/blog.ftl
```diff
@@ -8,7 +8,7 @@ blog-all-articles = All Articles
 blog-read-more = Read more
 blog-view-all = View All
 blog-back = Back
-blog-related-articles = Related Articles
+blog-recommended-articles = Recommended Articles
```
maribedran: The heading will still show "Related Articles"

springfield/cms/tests/test_blog_article_recommendations.py
```diff
@@ -0,0 +1,331 @@
...
+    set_recommended_articles(article, recommendation_pool.unmatched)
+
+    soup = get_article_soup()
+
+    assert soup.select_one(".fl-blog-recommended-articles") is None
+
+
+@pytest.fixture
+def minimal_recommendation(make_recommended_article, topic):
+    recommended_article = make_recommended_article("Matches topic", topic=topic, description="<p>Matches topic description</p>")
```
maribedran: This one is also confusing. The title says "Matches topic", but it doesn't have a topic.

springfield/cms/tests/test_blog_article_recommendations.py
```diff
@@ -0,0 +1,331 @@
...
+
+    recommended_article_elements = soup.select(".fl-blog-recommended-articles .fl-blog-article-list-item")
+    assert len(recommended_article_elements) == 1
+    article_element = recommended_article_elements[0]
+    date = article_element.select_one("p.fl-blog-article-date")
+    topic_heading = article_element.select_one("p.fl-superheading")
+    assert topic_heading is not None
+    assert topic_heading.get_text(strip=True) == "Topic"
+    assert date is not None
+    assert date.get_text(strip=True) == "Jan. 1, 2026"
```
maribedran: The related articles list shouldn't display either the date or the tags.

springfield/cms/tests/test_blog_article_recommendations.py
```diff
@@ -0,0 +1,331 @@
...
+    assert image is not None
+    assert "matches-topic-and-tag" in image.get("src", "")
+
+
+# ---------------------------------------------------------------------------
+# Editing blog article recommendations
+# ---------------------------------------------------------------------------
+
+
+def test_max_four_custom_recommendations_presented():
```
maribedran: This test seems unnecessary to me. We're basically testing that the framework respects the parameter we pass to the field. I'd rather avoid adding too many tests to the repo that don't really test our code.

springfield/cms/models/pages.py
```diff
@@ -2452,22 +2476,14 @@ def clean(self):
...
-            )
-            context["related_articles"] = list(related)
+        if not self.hide_recommended:
+            recommended = self.get_recommended_articles()
+            context["recommended_articles"] = list(recommended)
             blog_index = self.get_parent().specific
-            # Hidden tags are not displayed on the related articles listing
-            cache_localized_tags(context["related_articles"], blog_index.get_hidden_tag_keys())
+            # Hidden tags are not displayed on the recommended articles listing
+            cache_localized_tags(context["recommended_articles"], blog_index.get_hidden_tag_keys())
```
maribedran: Since the designs don't include tags on the related articles, we can remove this cache.

springfield/cms/models/pages.py
```diff
@@ -2509,6 +2525,57 @@ def get_listing_image_variants(self):
...
+            if article is None or not article.live or article.pk in recommended_ids:
+                continue
+            recommended.append(article)
+            recommended_ids.add(article.pk)
+        if recommended:
+            # Apply potential page view restrictions
+            public_article_ids = set(
+                BlogArticlePage.objects.public().filter(pk__in=[article.pk for article in recommended]).values_list("pk", flat=True)
+            )
+            recommended = [article for article in recommended if article.pk in public_article_ids]
```
stevejalim: What a pain that Wagtail doesn't give us a `.public` property on the `BlogArticlePage` that we could use alongside the `.live` check on line 2543

springfield/cms/models/pages.py
```diff
@@ -2509,6 +2525,57 @@ def get_listing_image_variants(self):
...
+        if len(recommended) == MAX_RECOMMENDED_ARTICLES:
+            return recommended
+
+        tag_ids = [tag.pk for tag in self.tags.all()]
+        shares_topic = Q(topic_id=self.topic_id) if self.topic_id else Q(topic_id__in=[])
+        shares_tag = Q(carries_a_matching_tag=True)
+        matching_siblings = (
```
stevejalim: Neat!

stevejalim (approved):
Nice one, thanks. R+

springfield/cms/models/pages.py
James-Osgood: Yes, 100%! I'm not entirely happy with this solution, but it solves the visibility rules without re-implementing `public()`.

springfield/cms/tests/test_blog_article_recommendations.py
maribedran: I do realize that I added similar tests in other PRs. It's been bothering me for a while where to draw the line. My current feeling is that if we let LLMs run free, we'll accumulate an enormous pile of not-so-important tests over time.

springfield/cms/tests/test_blog_article_recommendations.py
James-Osgood: It's a balance. I think working with a common fixture shared among several tests can be useful; you have a working model in your head for a suite of tests, rather than rebuilding it every time for each test. In this case, it helps show that ordering is still applied correctly when matching only by tag. Explicit is good, but I also like simple/short tests that isolate setup differences from the common case.

This test name isn't great though - the docstring says it better. I'll rename the test itself and improve the comments.

springfield/cms/tests/test_blog_article_recommendations.py
James-Osgood: Yes, I need to update that. I missed it during a test refactor.

springfield/cms/tests/test_blog_article_recommendations.py
James-Osgood: (I'm the one who came up with the test case, not the LLM. Although the LLM did implement it. I was doing TDD with the edit form in mind.)

I agree with excising unnecessary tests. Between the two edit tests, the validation one is more useful, so I'll drop this one.

springfield/cms/tests/test_blog_article_recommendations.py
maribedran: Agreed. My first reaction was, "Wait, it's still matching by topic, isn't it?" Then I looked at the setup, and it made sense. The name change will do :)

springfield/cms/tests/test_blog_article_recommendations.py
maribedran: I'd like to pick your brain on how you're using TDD. I'm still not happy with my workflow with Superpowers.

maribedran (approved):

